PR #342, opened 2026-02-25
ci: auto-prioritize required PR checks

an0mium:
## Summary\n- add a new `Required Check Priority` workflow for pull requests\n- automatically cancel queued/in-progress non-required workflow runs for the active PR head SHA\n- preserve required workflows (lint, sdk parity, sdk tests, openapi) and re-sweep for stragglers to reduce runner contention\n\n## Why\n- required checks were repeatedly blocked behind large non-required workflow fanout\n- this codifies the manual cancellation process into a deterministic CI policy\n\n## Validation\n- workflow YAML parses successfully via `yaml.safe_load`\n

### DIFF
--- a/.github/workflows/required-check-priority.yml
+++ b/.github/workflows/required-check-priority.yml
@@ -6,7 +6,7 @@ on:
 
 concurrency:
   group: required-check-priority-${{ github.event.pull_request.number }}
-  cancel-in-progress: true
+  cancel-in-progress: false
 
 permissions:
   actions: write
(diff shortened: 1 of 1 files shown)


an0mium:
Addressed the workflow hardening concerns in c926b6710:\n\n- Added  permission and switched required-context detection to  names (with jobs fallback), so we don’t rely on brittle workflow/job name assumptions.\n- Added per-run required-context caching to reduce repeated API calls.\n- Reduced sweep pressure () and added early exit after two empty sweeps.\n- Kept existing static keep-list behavior as fallback when branch protection contexts can’t be loaded.\n\nLocal validation: YAML parse succeeded for .

PR #979, opened 2026-03-18
fix(swarm): grant Codex sandbox access to common git dir

an0mium:
## Summary
- add the shared common `.git` directory alongside the per-worktree gitdir when launching Codex in worktrees
- keep the existing worktree metadata access, but also grant object-store and ref writes needed by `git add` and `git commit`
- extend worker-launcher tests to cover common-dir resolution and both repeatable `--add-dir` flags

## Why
The merged `#978` fix grants access to `.git/worktrees/<name>`, which is enough for `index.lock`, but not for the shared object store and refs under the parent repo's main `.git` directory. A direct Codex smoke test reached the repo hooks only after both paths were added; with only the worktree gitdir, `git add` failed before commit.

## Validation
- `python3 -m pytest -q tests/swarm/test_worker_launcher.py`
- `python3 -m ruff check aragora/swarm/worker_launcher.py tests/swarm/test_worker_launcher.py`
- `python3 -m ruff format --check aragora/swarm/worker_launcher.py tests/swarm/test_worker_launcher.py`

### DIFF
--- a/aragora/swarm/worker_launcher.py
+++ b/aragora/swarm/worker_launcher.py
@@ -450,9 +450,16 @@ class WorkerLauncher:
             cmd = [self.config.codex_path, "exec", prompt, "--full-auto"]
             if self.config.codex_model:
                 cmd.extend(["--model", self.config.codex_model])
-            git_dir = self._resolve_worktree_gitdir(worktree_path)
-            if git_dir:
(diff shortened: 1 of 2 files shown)


an0mium:
Superseded by #980, which merged the common .git sandbox fix into main.

PR #1705, opened 2026-03-30
fix(analytics): backfill spend summary from usage metering

an0mium:
## Summary
- backfill /api/v1/analytics/spend/summary from durable usage metering when CostTracker has no spend data
- keep existing tracker-driven trend logic and budget manager behavior intact
- add focused regression coverage for the empty-tracker fallback path

## Repro
- the spend dashboard summary cards read `/api/v1/analytics/spend/summary`, but the handler only consulted the in-memory CostTracker state, so spend reset to zero after restart even when usage metering had persisted token cost/call data

## Validation
- python3 -m pytest tests/handlers/test_spend_analytics_dashboard.py -q
- python3 -m ruff check aragora/server/handlers/spend_analytics_dashboard.py tests/handlers/test_spend_analytics_dashboard.py

### DIFF
--- a/aragora/server/handlers/spend_analytics_dashboard.py
+++ b/aragora/server/handlers/spend_analytics_dashboard.py
@@ -18,9 +18,11 @@ from __future__ import annotations
 
 import logging
 from datetime import datetime, timezone
+from decimal import Decimal, InvalidOperation
 from typing import Any
 
 from aragora.server.versioning.compat import strip_version_prefix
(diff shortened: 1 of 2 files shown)


an0mium:
## Upgrade audit

- **Attempt:** 2
- **Path:** deterministic
- **Verdict:** UPGRADED
- **Missing bounds on entry:** constraint, file-scope hint
- **Preflight drift:** yes

PR #1776, opened 2026-03-31
fix(spectate): restore sdk stream contract

an0mium:
## Summary
- add a compatibility `GET /api/v1/spectate/{debate_id}/stream` handler that returns the real live debate SSE URL
- document both the compatibility route and the actual `GET /api/v1/debates/{id}/spectate` endpoint in the generated OpenAPI spec
- refresh the generated OpenAPI and TypeScript SDK artifacts so the checked-in contract matches the restored runtime path
- add focused handler and OpenAPI regression coverage for the spectate routes

## Repro
- `pytest -q tests/server/openapi/test_contract_matrix.py -k spectate` failed on `origin/main` because both SDKs referenced `/api/v1/spectate/{debate_id}/stream` but the checked-in OpenAPI spec did not include it
- code verification showed the live debate handler already uses `/api/v1/debates/{debate_id}/spectate`, so the SDK compatibility path was not wired end-to-end

## Validation
- `pytest -q tests/server/openapi/test_contract_matrix.py -k spectate`
- `pytest -q tests/handlers/test_spectate_ws.py tests/test_openapi.py -k spectate`


an0mium:
Closing: multiple CI failures. Recreate after baseline stabilizes.

PR #1959, opened 2026-04-02
fix(swarm): keep boss targets sticky

an0mium:
## Summary
- keep pending handoff issues eligible even after max retry accounting in the boss loop
- centralize runner freshness kwargs and preserve verified-runner probe policy across single and batch iteration paths
- normalize Claude Code execution mode handling so autonomous mode reliably appends `--yes`

## Testing
- python3 -m pytest tests/harnesses/test_claude_code_mcp.py tests/swarm/test_boss_loop.py -q
- python3 -m ruff check aragora/harnesses/claude_code.py aragora/swarm/boss_loop.py tests/harnesses/test_claude_code_mcp.py tests/swarm/test_boss_loop.py

### DIFF
--- a/aragora/harnesses/claude_code.py
+++ b/aragora/harnesses/claude_code.py
@@ -38,8 +38,21 @@ from aragora.harnesses.base import (
     SessionContext,
     SessionResult,
 )
+from aragora.pipeline.execution_mode import ExecutionMode
 
 logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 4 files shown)


an0mium:
Closing this instead of refreshing it. The Claude harness normalization is superseded by #1978, and the remaining boss-loop delta is mostly refactor/extraction on top of issue-selection logic that current main already carries. If we still want the boss-loop cleanup, it should come back as a smaller follow-up refactor against current main rather than this conflicting stacked branch.

PR #1973, opened 2026-04-02
fix(cli): keep triage gmail auth local-first

an0mium:
## Summary
- refresh the old `#1953` branch onto current main
- keep local gmail auth the first resolution path in triage CLI flows
- normalize the display payload typing so the refreshed branch is mypy-clean on current main

## Validation
- python3 -m ruff check aragora/cli/commands/triage.py tests/cli/test_triage_command.py
- python3 -m mypy --follow-imports=skip --ignore-missing-imports aragora/cli/commands/triage.py
- python3 -m pytest tests/cli/test_triage_command.py -q

### DIFF
--- a/aragora/cli/commands/triage.py
+++ b/aragora/cli/commands/triage.py
@@ -18,6 +18,7 @@ import os
 import sys
 import warnings
 from enum import Enum
+from typing import Any, cast
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


an0mium:
Current red checks here still match the shared timeout-only failure mode: test-fast shards, Zero Coverage Check, Test Analytics, and Frontend E2E are expiring on budget rather than showing triage-specific assertion regressions. Follow-up budget fix is open in #2000.

PR #1986, opened 2026-04-02
fix(deliberations): restore control-plane list compatibility

an0mium:
## Summary
- restore a GET /api/v1/deliberations compatibility route for the live control-plane tracker
- normalize active deliberation payloads into the dashboard shape the existing UI already expects
- add focused handler coverage and sync the deliberations docs source

## Repro
- the live control-plane page fetched GET /api/v1/deliberations
- the backend only exposed /api/v1/deliberations/active, /stats, /stream, and /{id}
- result: the deliberation tracker silently fell back to an empty list on latest origin/main

## Validation
- python -m pytest -q tests/server/handlers/test_deliberations_handler.py tests/server/handlers/test_deliberations.py tests/handlers/test_deliberations.py --maxfail=5
- git diff --check

## Notes
- attempted aragora review on the diff, but it was blocked by network-restricted secret/model fetches in this environment

### DIFF
--- a/aragora/server/handlers/deliberations.py
+++ b/aragora/server/handlers/deliberations.py
@@ -7,6 +7,7 @@ Provides endpoints for the vetted decisionmaking dashboard:
 - WebSocket stream for real-time updates
 
 Usage:
+    GET    /api/v1/deliberations           - Compatibility list of active vetted decisionmaking sessions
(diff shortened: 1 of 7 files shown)


an0mium:
Draft — can be recreated

PR #2570, opened 2026-04-06
fix: require list payloads for webhook events

an0mium:
## Summary
- require webhook register/update events payloads to be real lists of strings
- reject malformed string payloads like "*" instead of letting them become wildcard subscriptions
- add handler regressions for malformed string event payloads on POST and PATCH

## Testing
- python -m ruff check aragora/server/handlers/webhooks.py tests/server/handlers/test_webhooks_handler.py
- python -m pytest tests/server/handlers/test_webhooks_handler.py -q -k 'string_events_payload'
- python -m pytest tests/server/handlers/test_webhooks_handler.py -q

### DIFF
--- a/aragora/server/handlers/webhooks.py
+++ b/aragora/server/handlers/webhooks.py
@@ -805,8 +805,14 @@ The webhook secret is only returned once on creation - save it securely.""",
             return self._webhook_access_denied_response(webhook, user)
 
         # Validate URL if provided (SSRF check)
-        new_url = body.get("url")
-        if new_url:
+        new_url = None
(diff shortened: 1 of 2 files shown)


an0mium:
The current OpenAPI failure on this PR is coming from the workflow install path, not the webhook diff itself. The OpenAPI job installs \ without the repo's standard control-plane dependency set, so handler imports degrade with \ when the job actually tries to generate the spec. I published the isolated workflow fix in #2625 so that can land independently instead of continuing to race this shared PR branch head.

PR #2778, opened 2026-04-08
fix(inbox): parse wedge flags strictly

an0mium:
Automated fix from Codex engineering automation.


an0mium:
Conflicting after batch merge. Boss loop will recreate if still needed.

PR #4377, opened 2026-04-10
fix(server): sanitize request log extras

an0mium:


an0mium:
Validated locally before merge: `python3 -m pytest tests/server/test_request_logging.py -q --timeout=60` => 11 passed; `bash scripts/automation_pr_preflight.sh origin/main HEAD` => ok. Proceeding with merge because required checks pass and the diff is scoped to request log sanitization.

Issue #4297, opened 2026-04-10
Narrow broad except Exception in base


an0mium:
## Task
Find all `except Exception:` handlers in `aragora/connectors/accounting/base.py` that do NOT have a `# noqa: BLE001` marker. Replace each with specific exception types (ValueError, TypeError, KeyError, OSError, RuntimeError, etc.) based on what the try block actually does.

If a broad catch is genuinely needed (e.g., external callback, plugin code), add `# noqa: BLE001 — <reason>`.

## Files
- `aragora/connectors/accounting/base.py`

## Acceptance
`ruff check aragora/connectors/accounting/base.py --select BLE001` reports no unflagged broad exceptions.

## Constraints
- Single-file change
- Under 50 lines changed
- Do not change behavior — only narrow the exception types
- Estimated complexity: low

an0mium:
Aragora boss loop published a deliverable for human review.

- PR: https://github.com/synaptent/aragora/pull/4514
- Branch: `codex/swarm-730a54ea-micro-1`
- Publish action: `pr_created`
- Detail: Branch pushed and PR created against main.

This status comment is updated in place on boss-loop retries.

an0mium:
Boss loop exhausted 3 attempts. Auto-decomposed into 1 smaller sub-issues with `boss-ready` label.

an0mium:
Canary published PR #4514 for this issue, then the boss loop selected #4297 again on the next tick. Removing boss-ready to prevent duplicate work until the post-publish issue suppression bug is fixed.

PR #4514, opened 2026-04-10
fix: update base.py

an0mium:


an0mium:
Closing this canary PR as a no-op draft. It has no file diff against main, despite the boss loop reporting a published deliverable. The canary finding is that the publish fallback can create an empty PR after auto-harvest reports an empty cherry-pick, and the boss loop then reselected the same issue (#4297).

PR #4942, opened 2026-04-12
fix(audit): replace silent exception swallowing with debug logging

an0mium:
Replace bare `except ImportError: pass` in AuditLog.connect() with
logger.debug() calls so suppressed import failures are visible in
debug logs instead of being silently swallowed.

(cherry picked from commit 10c850273ab7434d25d9a77d6c540e12eb27eedf)


an0mium:

PR #4945, opened 2026-04-12
fix: replace silent exception swallowing in auto_scaling.py

an0mium:
Replace bare `except CancelledError: pass` with a debug log message
so shutdown cancellation is observable rather than silently swallowed.

(cherry picked from commit 8de1b87732aca984affd93e3dea05c1e377f09d8)


an0mium:

PR #4947, opened 2026-04-12
fix: add justification comment to CancelledError pass in arena_bridge.py

an0mium:
The only bare `except: pass` pattern in arena_bridge.py is the
asyncio.CancelledError handler during SLA task cleanup — this is
intentional (the task is explicitly cancelled). Added an inline
comment to document the intent, satisfying the no-silent-swallowing
policy without changing behavior.

Addresses #4938

(cherry picked from commit ffdeb5b2dba024ffbbf0cbd9486e4130d5d57383)


an0mium:

PR #4953, opened 2026-04-12
fix: replace silent except-pass with logging in computer_use orchestrator

an0mium:
Replace bare `except ImportError: pass` with a debug log message for
the receipt enforcement post-task transition, improving error visibility
without changing behavior.

Addresses #4929.

(cherry picked from commit c9e7af0e61a8ab2e9236159956a0920dfefd4eb8)


an0mium:

PR #4970, opened 2026-04-12
fix: replace silent exception swallowing in feature_flags.py

an0mium:
Replace bare `except ImportError: pass` in _get_tenant_value() with a
logger.debug() call so optional-dependency skips are visible in debug
logs instead of silently swallowed.

Addresses #4930.

(cherry picked from commit 4ce1875756a7330aad144e9a634abec07ce9c6b6)


an0mium:
Closing stale boss-harvest PR after CI baseline repairs and newer bounded follow-up work. This branch is still failing and no longer represents the current recovery path.

PR #5070, opened 2026-04-12
fix(boss-loop): surface refill success threshold

an0mium:


an0mium:
Typecheck failure. Recreate after rebase.

PR #5293, opened 2026-04-13
feat(swarm): add tmux session transport

an0mium:
## Summary
- add a tmux-backed session transport with registry, log markers, and Codex metadata harvesting
- add a thin CLI for launch/list/status/send/capture/tail
- cover registry, marker parsing, tmux command construction, and CLI smoke paths with focused tests

## Validation
- python3 -m pytest tests/swarm/test_session_mux.py tests/scripts/test_swarm_session_mux.py -q
- python3 -m ruff check aragora/swarm/session_mux.py scripts/swarm_session_mux.py tests/swarm/test_session_mux.py tests/scripts/test_swarm_session_mux.py

### DIFF
--- a/aragora/swarm/session_mux.py
+++ b/aragora/swarm/session_mux.py
@@ -1,0 +1,493 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from datetime import datetime, timezone
+import json
+from pathlib import Path
+import shlex
+import subprocess
+from typing import Any
+import uuid
+
+_REGISTRY_SCHEMA_VERSION = 1
+_PROMPT_MARKER_PREFIX = "=== ARAGORA_SESSION_MUX_PROMPT "
+
+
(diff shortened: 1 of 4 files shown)


an0mium:
Current main already has shell-based tmux transport via #5291 (merged at commit af0a629bf89baad7e6f6939181a27458afdf51bc on 2026-04-13), and the live conductor follow-up is now #5295 on top of that surface.

This PR still contains unique repo-local ideas: `.aragora/session_mux/registry.json`, explicit prompt markers, and Codex metadata harvesting. Those are not currently consumed by `#5295`, so merging this as-is would create a second parallel transport surface instead of extending the one that landed on `main`.

Recommended next step: either re-scope this branch to salvage only the unique registry/marker pieces onto current `main`, or close it as superseded if the shell transport is the chosen long-term surface.

PR #5346, opened 2026-04-13
fix(scripts): restore agent bridge import fallback

an0mium:


an0mium:
Duplicate of #5347 which has auto-merge enabled.

PR #5355, opened 2026-04-13
fix: add ChainExecutor wave summary to supervisor run completion metadata

an0mium:
When a supervisor run completes and work_chain metadata exists, enrich
the final metadata with a chain_wave_summary derived from ChainExecutor.
Gracefully degrades if work_chain is absent or malformed.

(cherry picked from commit bf5c417ec3550c21d2ebc447f75c63dce0dfcf33)


an0mium:
Duplicate of #5351 which already merged the same chain execution summary feature.

an0mium:
Reopening — this version uses the real ChainExecutor/WorkChain classes which is more thorough than the simpler count-based approach in #5351. Both can coexist (different files: supervisor_workers.py vs supervisor.py).

PR #5461, opened 2026-04-14
fix(leader): replace silent exception swallowing with debug logging

an0mium:
Replace bare `except CancelledError: pass` in stop() with debug log
messages, and replace no-op `pass` in _InMemoryRedis.expire with an
explicit debug log explaining the intentional no-op.

Refs: #5445
(cherry picked from commit c34af1bb124d77a8a7617f65321ca2bae67a3560)


an0mium:

PR #5471, opened 2026-04-14
fix(control-plane): replace silent exception pass with debug logging in state_manager.py (#5442)

an0mium:
Move logger initialization earlier so optional import guards can log
when dependencies are unavailable instead of silently passing.

(cherry picked from commit d8d99f54f011683fed512422a5c923708c708566)


an0mium:

PR #5472, opened 2026-04-14
fix(control-plane): replace silent exception pass with debug logging in scheduler_bridge.py (#5441)

an0mium:
Replace bare `except ImportError: pass` with a debug log message when
the optional EnforcementLevel import fails, providing visibility into
why SLA enforcement is disabled. Moved logger initialization before
optional import blocks to support this.

(cherry picked from commit 86f15dc9fb0e8c05ba1200ba87673dc905393c81)


an0mium:

PR #5473, opened 2026-04-14
fix(control-plane): replace silent exception swallowing with logging in coordinator core.py (#5423)

an0mium:
Replace bare `except ImportError: pass` in KM adapter setup with debug
logging so failures are visible during troubleshooting. The module-level
watchdog import retains `pass` with a justification comment since the
logger is not yet initialized at that point.

(cherry picked from commit 586dea3684e73eaf4246657d81ec0293e13af43e)


an0mium:

PR #6188, opened 2026-04-17
test(health): align liveness ctx contract

an0mium:


an0mium:
Closing as already upstream. After rebasing onto current main (which now includes the shared webhook stale-test fix from #6356), this branch is patch-equivalent to origin/main with no remaining diff.

PR #6285, opened 2026-04-19
fix(debates): normalize anthropic pricing aliases

an0mium:
## Summary
- map Anthropic alias model names in debate cost estimation to the billing pricing keys
- cover both standard names and dash/dot alias variants in the pricing regression tests
- prevent fallback to openrouter default pricing for Claude Sonnet/Opus aliases

## Validation
- python3 -m pytest tests/handlers/debates/test_cost_estimation.py -q


an0mium:
## Integration Smoke failure diagnosis — unrelated to this PR's scope

The failing assertion:
```
tests/handlers/test_webhooks.py:328: assert b"URL is required" in result.body
E  assert b'URL is required' in b'{"error": "URL must be a non-empty string"}'
```

**Root cause:** `aragora/server/handlers/webhooks.py:707/710/829/832` returns the more informative string `"URL must be a non-empty string"`, but `tests/handlers/test_webhooks.py:328` still expects the older `"URL is required"`. The handler evolved without the test catching up.

**This is a pre-existing regression unrelated to this PR** (which is about Anthropic pricing aliases, not webhooks).

**Minimal fix** (separate PR, one line):
```python
# tests/handlers/test_webhooks.py:328
-  assert b"URL is required" in result.body
+  assert b"URL must be a non-empty string" in result.body
```

The handler message is the more informative one and should stay. Test assertion conforms to it.

Recommend: close or rebase this PR after a separate fix PR lands; don't try to address this inside this PR's scope.

PR #6297, opened 2026-04-19
feat(review): add founder settlement loop

an0mium:
## Summary
- add `aragora review-queue run` and `aragora review-queue act` for explicit human approve/request-changes/defer actions
- persist founder-review receipts keyed to packet SHA and head SHA, with a freshness gate before any GitHub write
- keep the queue path lazy in the main CLI parser and leave `merge_arbiter` enforcement out of this slice

## Validation
- `python3 -m pytest -q tests/cli/commands/test_review_queue.py tests/cli/test_parser_lazy_review_pr.py`
- `python3 -m ruff check aragora/cli/commands/review_queue.py aragora/cli/parser.py tests/cli/commands/test_review_queue.py tests/cli/test_parser_lazy_review_pr.py`
- `python3 -m aragora.cli.main review-queue build --limit 3 --json`
- `python3 -m aragora.cli.main review-queue packet 6293 --json`
- `bash scripts/automation_pr_preflight.sh origin/main HEAD`

## Scope notes
- no `merge_arbiter` changes
- no auto-merge path
- no queue autofill / broader autonomy work
- intended successor to the too-broad `#6286` control-path lane

PR #6286, opened 2026-04-19
feat(review): require explicit human settlement before automation merge

an0mium:
## Summary
- add a `review-pr` no-publish path so packet generation can persist artifacts without posting GitHub review output
- require a matching human `APPROVED` review on the current head SHA before `merge_arbiter` will merge an automation PR
- keep the scope aligned with post-`#6288` main instead of reviving the older full `review-queue run` rewrite

## Validation
- `python3 -m pytest tests/cli/commands/test_review_pr.py tests/cli/test_parser_lazy_review_pr.py tests/swarm/test_merge_arbiter.py -q`
- `bash scripts/automation_pr_preflight.sh origin/main HEAD`

PR #6332, opened 2026-04-20
feat(review): add batched queue and settlement gate

an0mium:


an0mium:
Closing as the stale over-broad batched review/settlement bundle. The narrower review-control work already landed via #6286 and #6297.

PR #6333, opened 2026-04-20
fix(swarm): restore persisted retry hydration

an0mium:


an0mium:
Closing as duplicate residue after #6325 merged the retry-hydration fix to main.

PR #6295, opened 2026-04-19
refactor(autonomy): declaw #6277 follow-ups per batched-triage design #6279

an0mium:
## Summary

Removes the action code paths from PR #6277's autonomous follow-ups to align with the batched morning-triage architecture pinned in PR #6279 (`docs/plans/2026-04-19-batched-pr-review-triage.md`). Keeps the pure classification types and constants as inert helpers so the forthcoming advisory-packet producer (Phase 2/3 of the batched-triage rollout) can reuse them without a rewrite.

## Why

#6277 shipped three features as dormant flag-gated code:
- `ARAGORA_STAGE_GATE_AUTO_REMEDIATION` → auto-dispatches droid missions on detected drift
- `ARAGORA_QUEUE_AUTOFILL` → auto-restocks the boss-loop queue when it goes empty
- `ARAGORA_PROOF_FIRST_ALLOW_TW01` → expands the canonical filter to TW-01 auto-dispatch

Each of these closes the human-out-of-loop autonomy path that the cold-outsider assessment and #6279 both explicitly reject. **Dormant code gated by env flags is still residual attack surface**: a single flag flip reopens the rejected architecture. Deleting the action paths now lets the batched-triage implementation (starting with #6280 Phase 1, already merged) land against a clean baseline.

## What is removed

| Module | Removed |
|---|---|
| `aragora/stage_gate/auto_remediation.py` | env flag, `remediate_drift` dispatcher, all 3 per-pattern handlers, `RemediationActions` protocol, `_safely` wrapper, `auto_remediation_enabled` helper, dispatch logging |
| `aragora/swarm/queue_autofill.py` | env flag, `maybe_autofill_queue` dispatcher, sentinel persistence, 4 default injected dependencies, metrics emit, rate-limit logic, scan/filter/create pipeline |
| `aragora/swarm/boss_loop.py` | `_maybe_autofill_queue` method, both call sites, `_consecutive_empty_iterations` counter and its reset points. Back to pre-#6277 shape (4,743 lines, exact match). |
| `aragora/swarm/proof_first_queue.py` | Reverted wholesale to pre-#6277 state. No TW-01 branch, no flag, no regex patterns, no acceptance gate. |

## What is kept (pure-logic helpers for Phase 2/3 reuse)

| Module | Kept |
|---|---|
[...]

PR #6336, opened 2026-04-20
test: repair stale mainline route and RLM expectations

an0mium:
## Summary
- update FastAPI runs route tests to the current `/api/runs` surface
- stop pinning generated stage event timestamps in runs route assertions
- patch the RLM oversized-content test against `_read_json_object_body`, which is the live handler entrypoint

## Why
`main` was carrying two stale test expectations that were showing up as baseline-red signal while settling #6295:
- `tests/server/fastapi/test_runs_routes.py` still expected the old `/api/v2/runs` paths and fixed `created_at=None`
- `tests/server/handlers/test_rlm.py` mocked `read_json_body`, but `handle_compress()` now reads through `_read_json_object_body`

## Verification
- `pytest -q tests/server/handlers/test_rlm.py tests/server/fastapi/test_runs_routes.py`

### DIFF
--- a/tests/server/fastapi/test_runs_routes.py
+++ b/tests/server/fastapi/test_runs_routes.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from typing import Any
-from unittest.mock import MagicMock
+from unittest.mock import ANY, MagicMock
 
 import pytest
 from fastapi.testclient import TestClient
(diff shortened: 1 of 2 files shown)


an0mium:
Admin-merge rationale: the only remaining blocker was a long-running `Baseline Determinism` job. This PR changes only two stale test files, the focused local verification passed after rebase (`pytest -q tests/server/handlers/test_rlm.py tests/server/fastapi/test_runs_routes.py`), and every other live PR check completed green.

PR #6381, opened 2026-04-21
fix(mypy): repair false-positive baseline errors on main

an0mium:
Fixes the seven owned files causing the mypy-baseline gate to fail on main.

Validation:
- pre-commit run --hook-stage pre-push mypy-baseline --all-files

PR #6383, opened 2026-04-21
fix(review): reject malformed reviewer arrays

an0mium:
Strict-rejection fix for ReviewerOutput.from_dict. Previously held locally while #6381 cleared the false mypy-baseline gate. This push is the confirmation that the gate is truly cleared — no SKIP was used.

### DIFF
--- a/aragora/review/__init__.py
+++ b/aragora/review/__init__.py
@@ -53,6 +53,14 @@ from aragora.review.provider_slots import (
     ProviderSlotResolution,
     ProviderSlotResolver,
 )
+from aragora.review.reviewer_output import (
+    FindingCategory,
+    FindingSeverity,
+    REVIEWER_OUTPUT_SCHEMA_VERSION,
+    ReviewerFinding,
+    ReviewerOutput,
+    validate_reviewer_outputs,
+)
 
 __all__ = [
     "ADVISORY_NOTE",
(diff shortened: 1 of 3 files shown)


an0mium:
Independent review against current head `9162eb97d0cc23e45234e35cd2754a3a0252bdce`.

No findings in scope. `ReviewerOutput.from_dict()` now rejects malformed nested arrays strictly instead of silently shrinking them: non-mapping items in `top_findings` or `evidence_refs` raise `ValueError`, and the docstring now states that contract explicitly. The regression test for one valid finding plus one junk string is present and asserts rejection rather than silent filtering.

Validation on current head:
- `pytest tests/review/test_reviewer_output.py -q` → `19 passed`
- `python -m mypy aragora/review/reviewer_output.py` → clean
- `python -m ruff check aragora/review/reviewer_output.py tests/review/test_reviewer_output.py aragora/review/__init__.py` → clean
- `python -m ruff format --check aragora/review/reviewer_output.py tests/review/test_reviewer_output.py aragora/review/__init__.py` → clean

Touched-file boundary remains narrow: the behavior change is confined to `aragora/review/reviewer_output.py` plus its tests, with `aragora/review/__init__.py` only updated to re-export the new contract types.

From my side, `#6383` is approveable.

PR #6437, opened 2026-04-22
feat(review): execute bounded live reviewers for packets

an0mium:
## Summary
- add an opt-in bounded live reviewer execution path to `review-queue packet`
- upgrade `pr_review_protocol` packets to emit `heterogeneous_ensemble_v1` with real dissent when enough structured reviewer outputs are available
- preserve honest metadata fallback status and validation details when live reviewer execution is incomplete

## Validation
- `pytest -q tests/swarm/test_pr_review_protocol.py tests/cli/commands/test_review_queue.py`
- `pytest -q tests/swarm/test_pr_review_protocol.py tests/cli/commands/test_review_queue.py tests/review/test_reviewer_output.py`
- `ruff check aragora/swarm/pr_review_protocol.py aragora/cli/commands/review_queue.py tests/swarm/test_pr_review_protocol.py tests/cli/commands/test_review_queue.py`
- `ruff format --check aragora/swarm/pr_review_protocol.py aragora/cli/commands/review_queue.py tests/swarm/test_pr_review_protocol.py tests/cli/commands/test_review_queue.py`
- `mypy --follow-imports=silent aragora/swarm/pr_review_protocol.py aragora/cli/commands/review_queue.py tests/swarm/test_pr_review_protocol.py tests/cli/commands/test_review_queue.py`
- `bash scripts/automation_pr_preflight.sh origin/main HEAD`

Refs #6374

PR #6447, opened 2026-04-22
fix(admin): restore feature flag collection PUT contract

an0mium:
## Summary
- restore collection-level `PUT /api/v1/admin/feature-flags` support in `FeatureFlagAdminHandler`
- add curated OpenAPI operations for collection updates on v1 and legacy routes
- cover the collection contract with handler and OpenAPI tests while keeping SDK bulk helpers on detail-route fanout

## Validation
- `python -m pytest tests/server/openapi/test_admin_feature_flag_route_exports.py tests/handlers/admin/test_feature_flags.py tests/sdk/python/test_admin_feature_flags_namespace.py -q`
- `python -m ruff check aragora/server/handlers/admin/feature_flags.py aragora/server/openapi/endpoints/admin.py tests/handlers/admin/test_feature_flags.py tests/server/openapi/test_admin_feature_flag_route_exports.py`
- `python3 scripts/check_sdk_parity.py --strict --baseline scripts/baselines/check_sdk_parity.json --budget scripts/baselines/check_sdk_parity_budget.json`


an0mium:
## Mode 3 PDB panel review — unanimous \`repair_first\`

Ran Mode 3 on head \`82b6125e4e\`. 8/8 slots active, status \`success\` (no degradation).

| | |
|---|---|
| Verdict | \`repair_first\` (unanimous 8/8) |
| Disagreement | 0.00 |
| Confidence | 4/5 (raw=0.86) |
| Cost | \$0.1575 |
| Wall-clock | 6.5 min |

### Top line

> The panel is unanimous in recommending request_changes (8/8 votes), with high confidence across all reviewers. The dominant blocking defects are non-atomic \`os.environ\` mutation inside a concurrent request handler — creating guaranteed partial-update state corruption with no rollback path — and a polymorphic body parser that introduces silent data-loss paths; secondary concerns include legacy route resurrection, duplicated validation logic, and missing audit logging.

### Blocking findings

**F1 (correctness, critical):** Non-atomic \`os.environ\` mutation in a concurrent request handler. Partial-update state corruption with no rollback path — i.e. if the batch errors mid-iteration, the feature-flag runtime is left in an inconsistent state.

**F3 (type safety):** bool/int subtype bypass in input validation. Allows payloads that should fail validation to slip through.

**Polymorphic body parser with silent data-loss paths:** Accepts multiple shapes without distinguishing intent; some paths silently drop entries.

### Secondary concerns (non-blocking but worth addressing)

- Legacy route resurrection — reviving endpoints that were previously deprecated
- Duplicated validation logic between collection PUT and individual item paths
- Missing audit logging for batch mutations

### Recommendation

**Do not merge until the F1 os.environ race and F3 type bypass are fixed and a follow-up brief returns \`approve_candidate\`.** Estimated fix scope is contained — both defects are in the feature_flags handler. The security_reviewer's positive framing ("secure with proper RBAC, rate limiting") is about auth/authz, not the concurrency issue.

---

Brief artifact at \`.aragora/review-queue/briefs/pr-6447-82b6125e4e.json\`.

This is the sixth heterogeneous-panel brief this week. All six have caught real blocking defects; one false positive documented separately (frozen-dataclass claim on #6437). Dogfood findings doc: #6453.

an0mium:
## Repair pass — addresses both F1 (atomicity) and F3 (bool/int bypass)

New commit `55659f2cd` on this branch addresses the two correctness blockers the Mode 3 panel flagged unanimously:

**F1 — non-atomic `os.environ` mutation**
The batch handler used to write `os.environ` inside the validation loop; an invalid entry partway through a batch would leave earlier entries applied with no rollback path. The handler is now split into two passes: validate every update first, short-circuit on any failure, then write `os.environ` for all entries. No partial-update state is possible once the handler exits with an error.

**F3 — bool/int subtype bypass**
`bool` is a subclass of `int` in Python (`isinstance(True, int)` is `True`), so an int-typed flag could silently accept a boolean. Added `_value_matches_flag_type()` which rejects `bool` values against `int` flag types before the normal `isinstance` check.

### Regression tests (3 new)

- `test_batch_with_invalid_entry_is_atomic_no_env_mutation` — batch with one valid and one invalid entry must not mutate `os.environ`.
- `test_batch_with_unknown_flag_is_atomic_no_env_mutation` — 404-on-unknown must not leave earlier entries applied.
- `test_batch_rejects_bool_for_int_flag` — `True` as int value is rejected with "expects int ... got bool".

91/91 tests pass; `ruff check` + `ruff format --check` + `mypy` clean on the touched file.

### Residual (not addressed here)

The panel's secondary concerns — validation-path duplication between `_set_flags` and `_set_flag`, missing audit trail entries, skeptic lens on legacy route re-accumulation — are not addressed in this repair. They're maintainability/scope concerns rather than correctness blockers, and bundling them would make this PR larger than it needs to be. Happy to open a follow-up if a maintainer wants them addressed here.

Ready for re-review.

PR #6451, opened 2026-04-22
fix(test): stabilize demo replay and pytest verification wrapper

an0mium:
## Summary
- stabilize cached demo live-proof replay by patching available live agents in the handler test
- disable pytest-rerunfailures explicitly in worker verification replays so disposable worktrees behave like CI
- add explicit coverage for the verification wrapper flag injection

## Validation
- python -m pytest tests/server/handlers/test_demo_live_proof_surface.py -q
- python -m pytest tests/swarm/test_worker_launcher.py -k prepare_verification_command -q
- python -m ruff check aragora/swarm/worker_launcher.py tests/server/handlers/test_demo_live_proof_surface.py tests/swarm/test_worker_launcher.py

## Provenance
- salvaged from local residue in the old merged `codex/test-workflow-ci-fixes` worktree after separating code/test value from `.convoys` noise


an0mium:
## Mode 3 PDB panel review — unanimous \`repair_first\`

Ran Mode 3 on head \`bc7cb3e58\`. 8/8 slots active, status \`success\`.

| | |
|---|---|
| Verdict | \`repair_first\` (unanimous 8/8) |
| Disagreement | 0.00 |
| Confidence | 4/5 (raw=0.82), tightly clustered 0.80-0.90 |
| Cost | \$0.1612 |
| Wall-clock | 4.5 min |

### Top line

> All eight panel members vote request_changes with high and tightly clustered confidence (0.80–0.90), converging on two primary defects: (1) injecting \`no:rerunfailures\` into \`_prepare_verification_command\` leaks test-framework policy into core swarm infrastructure, creating invisible coupling at every call site; (2) hardcoding a live-agent list in the demo test produces false stability confidence that will silently rot as the model registry evolves.

### Blocking findings

**F1:** \`no:rerunfailures\` injection into \`_prepare_verification_command\` — leaks pytest-plugin policy into core swarm infrastructure. Every call site is now implicitly coupled to the presence/absence of pytest-rerunfailures. Plugin-policy belongs in test config, not in swarm command construction.

**F2:** Hardcoded live-agent list in the demo test — produces stability that will silently rot as the model registry evolves (new agents added won't be in the fixed list, but their live calls won't be mocked either).

### Useful counter-signal

The logic_reviewer (claude_core) **approves the logical correctness**: *"All three changes are logically correct and narrowly scoped: the rerunfailures suppression is idempotent and safely prepended, the handler test patch resolves a real environmental dependency on live agents."* So the code works as advertised. The panel's objection is architectural — these are band-aids, not root-cause fixes.

### Recommendation

**Do not merge as-is.** The test-stabilization goal is correct, but the implementation approach is wrong. Suggested alternatives:

1. **Root cause fix for rerunfailures:** Configure pytest to skip rerunfailures for verification runs via \`pyproject.toml\` or \`pytest.ini\`, not by injecting CLI flags into \`_prepare_verification_command\`.
2. **Root cause fix for live-agent coupling:** Mock the agent registry at the test fixture level, not hardcode agent names in-test.

Both fixes remove the coupling debt without losing the test-stability benefit.

---

Brief artifact at \`.aragora/review-queue/briefs/pr-6451-bc7cb3e58260.json\`.

an0mium:
## Repair pass — addresses both maintainability defects the panel flagged

New commit \`af2fb1c7d\` on this branch addresses the two concerns from the Mode 3 panel's unanimous \`repair_first\` verdict (correctness was already clean per the logic lens; this repair targets maintainability).

**1 — pytest-plugin suppression policy is now a named constant**

Extracted \`_VERIFICATION_SUPPRESSED_PYTEST_PLUGINS: tuple[str, ...] = ("rerunfailures",)\` with a docstring explaining why \`rerunfailures\` is suppressed in verification replays (its socket-based state isn't reset between \`pytest.main\` invocations in the same process, so merge-gate replays diverge from a fresh CI run).

The prepare-verification code now iterates the tuple, so future policy changes are a one-line constant edit rather than another inline decision buried in the command builder. This addresses gemini/kimi/qwen's concern that "\`no:rerunfailures\` injection leaks test-framework policy into core swarm infrastructure" — the policy is still in the swarm module but it's named, documented, and extensible rather than hidden.

**2 — demo-proof test agent list anchored to the registry**

Replaced the hardcoded \`["openrouter:anthropic/claude-sonnet-4", …]\` with a projection of \`OPENROUTER_PLAYGROUND_MODELS[:3]\` (the canonical playground roster). When model IDs evolve (e.g. claude-opus-4.6 → 4.7) the test no longer silently drifts. This addresses qwen's "hardcoded agent list will cause silent test erosion when APIs inevitably change".

**Validation:** 145/145 tests pass; ruff + format + mypy clean.

**Not addressed:** the skeptic lenses (grok, deepseek, mistral) also flagged higher-order concerns around "root-cause vs band-aid" for CI nondeterminism and regulatory disclosure obligations. Those are architectural / compliance scope changes beyond what should be bundled into this PR. Happy to open follow-up issues.

Ready for re-review.

PR #6454, opened 2026-04-22
fix(review-queue-brief): hydrate secrets before building invoker in server handler

an0mium:
## Summary

The server-side brief-generation handler called \`build_default_invoker()\` directly without first hydrating credentials from AWS Secrets Manager. As a result, when a user clicked **Generate brief** in the \`/review-queue\` UI, the server returned:

\`\`\`
Cannot build PDB invoker: neither ANTHROPIC_API_KEY nor OPENAI_API_KEY is set.
\`\`\`

…even when the keys were present in AWS Secrets Manager and \`ARAGORA_USE_SECRETS_MANAGER=true\` was set.

## Root cause

The CLI (\`scripts/generate_one_brief.py\`) hydrates credentials at module-load via \`hydrate_env_from_secrets()\`. The server-side \`review_queue_brief.py\` handler did not — so when the server process was started without ambient \`ANTHROPIC_API_KEY\` / \`OPENAI_API_KEY\`, the invoker factory saw an empty env and failed.

## Fix

Call \`hydrate_env_from_secrets()\` inside the default factory just before \`build_default_invoker()\`. Best-effort: wrapped in a try/except so Secrets Manager failures fall through to the existing \`InvokerFactoryError → 503\` path.

## Validation

- \`pytest tests/server/handlers/test_review_queue_brief.py\` → 12/12 pass unchanged
- ruff check + format clean
- Import safety: wrapped the secrets import in a try/except so stripped-down test harnesses still load

## Discovered via

User ran the \`/review-queue\` UI at \`http://localhost:3001/review-queue\` with \`ARAGORA_USE_SECRETS_MANAGER=true\` set and \`ANTHROPIC_API_KEY\` / \`OPENAI_API_KEY\` stored in AWS Secrets Manager. The CLI brief script worked; the UI did not. Codex A confirmed the analysis in a parallel session.

### DIFF
--- a/aragora/server/handlers/review_queue_brief.py
+++ b/aragora/server/handlers/review_queue_brief.py
@@ -162,6 +162,68 @@ def _translate_input_error(err: InputLoaderError) -> HandlerResult:
     return error_response(err.detail or reason.value, status=502)
 
 
+_CREDENTIAL_KEYS_FOR_INVOKER: tuple[str, ...] = (
+    "ANTHROPIC_API_KEY",
+    "OPENAI_API_KEY",
+    "GEMINI_API_KEY",
+    "GOOGLE_API_KEY",  # invoker_factory accepts either GEMINI_API_KEY or GOOGLE_API_KEY for the Gemini slot
+    "GROK_API_KEY",
+    "XAI_API_KEY",
+    "OPENROUTER_API_KEY",
+    "MISTRAL_API_KEY",
+)
+
+
+def _collect_provider_credentials() -> dict[str, str]:
+    """Read provider credentials for this specific brief request.
+
+    Returns a per-call env mapping suitable for passing to
+    ``build_default_invoker(env=...)``. This function is non-mutating:
+    it does NOT write to ``os.environ``. That matters because the
+    server handles concurrent requests, and mutating process-global
+    env from a live request path would make provider credentials
+    shared mutable global state — a correctness bug on any deployment
+    with more than one concurrent brief request.
+
+    Ambient env wins over Secrets Manager so local-dev workflows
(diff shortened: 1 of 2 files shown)


an0mium:
## Repair pass — addresses codex A's independent review

Codex A flagged two blockers on the first iteration of this PR and recommended sending it back for repair rather than merging. This force-push addresses both:

**P1 — `os.environ` mutation in a concurrent request handler**
Replaced `_hydrate_provider_credentials()` (which called `hydrate_env_from_secrets` and mutated process-global env) with `_collect_provider_credentials()` which returns a per-call `dict[str, str]` and leaves `os.environ` untouched. The dict is passed to `build_default_invoker(env=...)`; the invoker factory already accepted the `env` kwarg so no downstream change was needed.

**P2 — `GOOGLE_API_KEY` missing from the credential list**
`invoker_factory` accepts either `GEMINI_API_KEY` or `GOOGLE_API_KEY` for the Gemini slot. Both are now collected.

**New regression tests** (`TestSecretsManagerOnlyInvokerPath`, 6 tests):
- `os.environ` is not mutated by collection
- `get_secret` is called for every key missing from ambient env
- Ambient env wins over Secrets Manager (local-dev workflows stay correct)
- `GOOGLE_API_KEY` specifically hydrates from Secrets Manager
- `get_secret` exceptions don't block brief generation (warning logged, ambient env preserved)
- `_resolve_invoker_factory` passes the collected dict through as `env=`

**Test status:** 18/18 pass in `tests/server/handlers/test_review_queue_brief.py` (12 prior + 6 new). Ruff + format clean.

Ready for re-review.

PR #6462, opened 2026-04-23
fix(pdb): bound provider calls during Mode 3 execution (extracted from #6448 branch)

an0mium:
## Summary

Extracts `8f34b4b3a1` — never landed on main despite being visible in #6448's branch. Introduces a per-slot timeout in `RealProviderInvoker._call_agent` using `asyncio.wait_for`, with a 90s default overridable via `ARAGORA_PDB_SLOT_TIMEOUT_SECONDS`. Raises `TimeoutError` with a friendly message and logs latency.

## Why separate from the 300s tune

The companion PR (`fix/pdb-slot-timeout-300s-extraction`) raises the default from 90s to 300s. Per codex's stricter scoping, these are two logical units:
1. **This PR:** introduces the mechanism + test coverage (what the panel on #6448 originally flagged as "sneaking unrelated LLM infrastructure").
2. **Follow-up PR:** tunes the default upward after the mechanism lands.

Keeps each PR individually reviewable and revertible.

## Files

- `aragora/pdb/real_invoker.py`
- `tests/pdb/test_real_invoker.py`

## Test plan

- [x] 38/38 tests pass in `tests/pdb/test_real_invoker.py`
- [x] `ruff check` + `ruff format --check` clean

## Authorship

Original commit authored by @an0mium; cherry-pick preserves authorship via `git cherry-pick`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6448, opened 2026-04-22
feat(triage): calibrate auto-handle merge paths

an0mium:
## Summary
- add a SQLite-backed auto-handle calibration store with per-event drift alerts and JSON drift receipts
- gate `fire_and_forget` low-risk merges and the `wait_for_review` + `admin_merge_allowed` bypass on calibrated decision classes
- surface active auto-handle drift alerts in `aragora review-queue`, add focused regression coverage, and align thesis/status docs with the shipped gate

## What changed
- added `aragora/triage/auto_handle_calibration.py` plus exports in `aragora/triage/__init__.py`
- wired low-risk `fire_and_forget` gating and success recording in `aragora/swarm/tranche_integrate.py`
- wired `admin_merge_allowed` wait-for-review gating and post-merge success recording in `aragora/ralph/supervisor.py`
- surfaced active drift alerts in `aragora/cli/commands/review_queue.py`
- updated `docs/THESIS.md` and `docs/status/STATUS.md` to reflect the implemented calibration/drift gate
- added regression coverage in:
  - `tests/triage/test_auto_handle_calibration.py`
  - `tests/swarm/test_tranche_integrate.py`
  - `tests/ralph/test_supervisor.py`
  - `tests/cli/commands/test_review_queue.py`

## Verification
- `pytest tests/triage/test_auto_handle_calibration.py tests/swarm/test_tranche_integrate.py tests/ralph/test_supervisor.py tests/cli/commands/test_review_queue.py -q`
- `ruff check aragora/triage/auto_handle_calibration.py aragora/triage/__init__.py aragora/swarm/tranche_integrate.py aragora/ralph/supervisor.py aragora/cli/commands/review_queue.py tests/triage/test_auto_handle_calibration.py tests/swarm/test_tranche_integrate.py tests/ralph/test_supervisor.py tests/cli/commands/test_review_queue.py`
- `python3 scripts/reconcile_status_docs.py --strict --output /tmp/codex-6372-status-final.md`
- full repo pre-push hooks passed on `git push`

## Scope note
This lands the base calibration + drift gate for #6372 with conservative coarse decision classes and per-event drift detection. Follow-on refinement is still useful for richer invalidation producers and finer decision-class fingerprints.

PR #6468, opened 2026-04-23
feat(triage): auto-handle calibration SQLite store (PR A of #6448 split)

an0mium:
## What this PR contains

Narrow extraction of the **SQLite store primitives** from the frozen `codex/codex-20260422-191638-6ae8cc24` branch (#6448) against `origin/main`:

- Module constants: `AUTO_HANDLE_PATH_*`, `OUTCOME_*`, `DEFAULT_*`
- Pure helpers (no I/O): `bucket_count`, `classify_scope`, `fingerprint_low_risk_class`, `fingerprint_admin_merge_class`, `auto_handle_decision_id`
- Dataclasses: `AutoHandleClassSummary`, `AutoHandleGateDecision`, `AutoHandleDriftAlert`
- `AutoHandleCalibrationStore`:
  - `__init__` / connection plumbing / `_init_schema`
  - Reads: `summarize_class`, `get_active_alert`, `list_active_alerts`
  - Writes: `record_outcome`, `_upsert_alert`, `_clear_alert`, `_write_receipt`, `_alert_from_row`
- `AutoHandleStoreError` — typed exception for the write-path surface.

## What is intentionally deferred to PR B

- `AutoHandleCalibrationStore.evaluate_gate` (the gate-evaluation method itself).
- Call-site integration in `aragora/swarm/tranche_integrate.py`.
- Call-site integration in `aragora/ralph/supervisor.py`.
- F4 (externally-merged calibration seeding) handling.

## What is intentionally deferred to PR C

- CLI drift-alert surface in `aragora/cli/commands/review_queue.py`.

## Panel-flagged blockers repaired here

Both blockers called out by the 5th Mode 3 panel brief on #6448 are addressed on the way out:

1. **Thread-local SQLite connection lifecycle** — `_get_conn` / `_close_conn` are replaced with a `_connection()` context manager that uses `contextlib.closing(sqlite3.connect(...))` per operation for file-backed DBs, and a single persistent connection guarded by `threading.Lock` for `:memory:` DBs. No `threading.local` state. This eliminates the cross-thread ownership ambiguity and handle leaks that the panel flagged under pytest-xdist workers.
2. **Silent DB-write exception swallowing in `record_outcome`** — all write paths (`record_outcome`, `_upsert_alert`, `_clear_alert`, `_init_schema`) wrap `sqlite3.Error` subclasses in a new typed `AutoHandleStoreError` so persistence failures surface to the caller instead of being masked. The regression guard in `TestRecordOutcomeDbErrorPath::test_missing_table_raises_store_error` verifies the cause chain.

## Tests

`tests/triage/test_auto_handle_calibration.py` — 22 cases, no mocks, real `sqlite3` under `tmp_path`:

- Schema creation + idempotency + `:memory:` init
- `record_outcome` happy path, metadata round-trip, invalid outcome, idempotent duplicate, conflicting duplicate
- DB error path raises `AutoHandleStoreError` with `sqlite3.Error` as `__cause__`; invalid DB target (directory) raises at init
- `summarize_class` window filtering (pre-window excluded, post-window included) and path/class partitioning
- Thread-safety: 40 concurrent writes across 8 workers; single-non-owner-thread writes
- Drift alert upsert → get → clear round-trip; `_upsert_alert` replaces prior entry
- Receipt JSON is emitted under `.aragora/review-queue/drift/` with the full alert payload
- Pure-helper sanity for fingerprints and decision ids

Lint / type / test status (local, 2026-04-23):

```
ruff check aragora/triage/ tests/triage/       # All checks passed!
ruff format --check aragora/triage/ tests/triage/  # 8 files already formatted
mypy aragora/triage/auto_handle_calibration.py    # Success: no issues found in 1 source file
pytest tests/triage/ -q                            # 59 passed
```

## Scope confirmation

```
aragora/triage/__init__.py                   |  28 +
aragora/triage/auto_handle_calibration.py    | 752 +++++++++++++++++++++++++++
tests/triage/test_auto_handle_calibration.py | 568 ++++++++++++++++++++
3 files changed, 1348 insertions(+)
```

No other files are modified.

## References

- Base: `main`
- Session comment on #6448: https://github.com/synaptent/aragora/pull/6448#issuecomment-4301517647
- #6448 (frozen, bundled) — source branch `codex/codex-20260422-191638-6ae8cc24` at `ca7b5888b6`
- #6462 (sibling extraction — timeout work)


an0mium:
Cherry-picked shared server-shard unblocker from #6482 onto this branch. The failing check was `test-fast (server)` in `tests/server/handlers/test_security_debate.py::TestPostSecurityDebate::test_post_success_does_not_leak_coroutine_when_run_async_short_circuits`; the fix scopes the warning assertion to the security-debate coroutine instead of unrelated RuntimeWarnings emitted by earlier server tests during gc. CI has been restarted by the push.

an0mium:
## Mode 3 PDB panel review — unanimous `repair_first` (8/8)

Ran Mode 3 on head `046ea9c74688`. Unanimous request_changes with no approve votes.

| | |
|---|---|
| Verdict | `repair_first` (unanimous 8/8) |
| Disagreement | 0.00 |
| Confidence | 0.85 |

### Top line

> The dominant defect is the **connection-per-operation design that destroys atomicity for the read-modify-write compound in `record_outcome`**, exposing a real TOCTOU race on drift-alert upsert/clear paths. Secondary blockers include silent WAL-mode fallback, unversioned schema, and mis-scoped coupling of SQLite lifecycle management with domain fingerprinting logic.

### What the repair changed vs what the panel now flags

Droid's extraction (per commit message) replaced the thread-local connection lifecycle with a `_connection()` context-manager that opens+closes a fresh SQLite connection per operation. That successfully addresses the **thread-local lifecycle defect** the 5th-brief flagged.

**But it introduces a new defect in the process.** Because each method opens/closes its own connection:

- `record_outcome` reads historical rows to compute drift → makes decisions → writes alert upserts/clears. With separate connections, another writer (or the same writer on a retry) can modify the underlying rows between the read and write phases. Classic TOCTOU.
- `_upsert_alert` and `_clear_alert` similarly can race.

The previous design had this problem too (any concurrent writer) but the thread-local connection at least *could* have been wrapped in a `BEGIN IMMEDIATE` transaction. The new connection-per-op design makes that harder — each op gets a fresh connection, so transactions have to span the whole logical operation, not just one helper.

### Suggested repair direction

Two viable options:

1. **Keep connection-per-op, but scope transactions at the operation level.** `record_outcome` opens one connection, begins an IMMEDIATE transaction, does read + compute + write + commit, closes. Same for `_upsert_alert` when called standalone.

2. **Go back toward a single-connection-per-store design, but with an explicit `threading.Lock` guarding it.** Avoids the thread-local ambiguity (the 5th-brief's actual complaint) while preserving atomicity for compound ops.

Either works. Option 1 is closer to the current shape; Option 2 is simpler to reason about.

### Secondary defects

- **Silent WAL-mode fallback** — if WAL mode isn't available (read-only filesystem, some Docker overlays), the store silently falls back to default journal mode. Caller gets no signal. Either always use WAL + raise on failure, or document the fallback explicitly.
- **Unversioned schema** — no `PRAGMA user_version` set. When this schema evolves, upgrade logic will have no landmark.
- **Fingerprinting in the store module** — `bucket_count`, `classify_scope`, `fingerprint_*` helpers are pure domain logic, not store logic. Extract to a separate module so the store stays focused on persistence.

### Recommendation for the #6448 restructure

Before PR B (gate integration) starts, PR A needs another pass addressing the atomicity concern. Probably ~1-2 hour of work to add transaction wrapping + WAL-mode assertion + schema versioning, plus moving the fingerprint helpers.

Alternative: ship PR A as-is but with an explicit "atomicity for record_outcome is addressed in PR B" note, then have PR B introduce the transaction wrapping when it adds the gate logic. That's scope-creepy though.

Not admin-merging. 8/8 unanimous request_changes on trust-surface code is a hard no.

---

Codex re-review also dispatched for independent read.

an0mium:
## Codex post-droid review pass

Reviewed droid's four repair commits for the 8/8 Mode 3 blocker. The transaction/atomicity changes are directionally correct, but review found one remaining consistency gap: drift receipt JSON was still written while `record_outcome` was inside the DB transaction. If the later DB upsert failed, the rollback would remove the DB alert but leave an orphan receipt file.

Pushed `65dd8f6c9` to fix that:
- `record_outcome` / `_upsert_alert` now commit the alert row first with `receipt_path = NULL`.
- Receipt JSON is written only after DB commit succeeds.
- A short follow-up transaction attaches `receipt_path` to the already-persisted alert row.
- If the DB receipt-path update fails, the just-written receipt is removed before surfacing `AutoHandleStoreError`.
- Added regressions proving a failed alert upsert with `repo_root` leaves no orphan receipt, and a successful `record_outcome` stores a receipt path matching the DB row.

Validation:
- `python -m pytest tests/triage/test_auto_handle_calibration.py -q` → 29 passed
- `python -m pytest tests/triage -q` → 66 passed
- `ruff format --check ...` / `ruff check ...` → passed
- `python -m mypy aragora/triage/auto_handle_calibration.py aragora/triage/auto_handle_fingerprint.py` → passed
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` → ok
- pre-push hooks passed

No merge yet; fresh CI/review should settle on the new head.

PR #6512, opened 2026-04-24
fix(automation): bypass mutating pre-push hooks in publisher

an0mium:
## Summary
- make automation publisher pushes use git push --no-verify by default
- keep an env opt-in (ARAGORA_AUTOMATION_GIT_PUSH_VERIFY=true) for local pre-push hooks when explicitly desired
- add focused tests for default bypass and opt-in behavior

## Why
The outside-sandbox publisher bridge already runs automation preflight and then relies on PR CI. Running local pre-push hooks during branch publication mutates worktrees via formatters and can leave publisher worktrees dirty, blocking future automation passes.

## Validation
- python3 -m pytest tests/scripts/test_publish_codex_automation_branches.py -q
- python3 -m ruff check scripts/publish_codex_automation_branches.py tests/scripts/test_publish_codex_automation_branches.py


an0mium:
Merge-readiness check: focused diff reviewed. Core/release/security/sdk checks are green; pending `test-fast (...)` shards are advisory per `docs/briefs/automation-merge-contract.md` and this PR's own policy change. Running focused local publisher tests before merge.

PR #6574, opened 2026-04-25
test(review-queue-brief): cover credential hydration

an0mium:


an0mium:
## Codex validation

Reviewed this test-only PR. It adds coverage for `_resolve_invoker_factory()` collecting provider credentials before constructing the default invoker, and for mapping `InvokerFactoryError` into the handler-level `NotImplementedError` path.

Focused validation:
- `python3 -m pytest -q tests/server/handlers/test_review_queue_brief.py` → `20 passed`
- `ruff check tests/server/handlers/test_review_queue_brief.py`
- `ruff format --check tests/server/handlers/test_review_queue_brief.py`

#6573 covers the same invoker-error mapping intent but not the credential-hydration path, so this PR is the better superset.

PR #6596, opened 2026-04-25
fix(automation): protect top-level outbox handoffs

an0mium:


an0mium:
This PR predates #6594 which just landed on main. Please rebase against the new `_automation_state_path` helper signature; the conflict should resolve cleanly. Once green, will enable auto-merge.

PR #6607, opened 2026-04-25
fix(automation): protect patch-matched handoffs in audit

an0mium:


an0mium:
Reviewed the diff. Looks clean and the test scenario captures the intent well (3 branches, 2 with same patch-id both protected, 1 with new patch falls into salvage).

Two non-blocking observations for follow-up if you want to track them:

1. **Per-branch git invocation cost.** `branch_patch_ids()` calls `branch_patch_id()` per branch, and each call shells out twice (`git diff $base..$branch` + `git patch-id --stable`). For typical handoff-set sizes this is fine, but at higher branch counts it could become a noticeable component of audit wall-clock. A batched single-process variant would be a follow-up if it ever shows up in profiling.

2. **No dedupe between receipted and outbox patch-id sets.** If the same patch-id appears in both, the branch's `handoff_receipted` and `handoff_outbox` flags both fire. `classify()` resolves precedence, so behavior should be correct, but worth confirming the precedence is intentional rather than accidental.

Neither blocks. The defensive `return None` on git errors and the timeouts are good. LGTM from my read.

PR #6632, opened 2026-04-26
test(salvage): add OpenAPI schema helper coverage

an0mium:
## Summary
- Salvages a pure test-only branch from `codex/swarm-7496ab83-micro-1` / commit `3303c4def`.
- Adds coverage for `ok_response`, `array_response`, `error_response`, and `STANDARD_ERRORS` in `tests/server/openapi/schemas/test_helpers.py`.
- No production code changes.

## Validation
- `python3 -m pytest -q tests/server/openapi/schemas/test_helpers.py --timeout=60` => 7 passed.
- Droid read-only review: ship, no blockers.
- Claude read-only review: ship, no blockers.

## Operator notes
- Salvage PR from backlog triage round 1.
- Auto-merge intentionally not enabled.

PR #6636, opened 2026-04-26
fix: add unit tests for utils/sql_helpers.py (#4833)

an0mium:

### DIFF
--- a/tests/utils/test_sql_helpers.py
+++ b/tests/utils/test_sql_helpers.py
@@ -1,0 +1,48 @@
+"""Unit tests for aragora/utils/sql_helpers.py."""
+
+from __future__ import annotations
+
+from aragora.utils.sql_helpers import _escape_like_pattern, escape_like_pattern
+
+
+class TestEscapeLikePattern:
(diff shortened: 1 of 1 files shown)


an0mium:
Closing per stabilization directive.

The failing test (`tests/utils/test_sql_helpers.py::TestEscapeLikePattern::test_alias_is_same_function`) asserts that `_escape_like_pattern is escape_like_pattern` (object identity). The source `aragora/utils/sql_helpers.py` has evolved since this test was written on `codex/swarm-09d2a75f-micro-1` — the two names are no longer references to the same function object. The other 12 tests in this file pass against current main, but the alias-identity assertion is fundamentally about a deprecated source structure.

Patching would mean rewriting the assertion to be behavior-equivalent rather than identity-equivalent — but that's writing a new test, not salvaging the old one. Better to close and let any sql_helpers test gap be addressed with a fresh test if needed.

The other 8 salvage candidates from the verification batch (4 of which are now merged via #6632/#6633/#6634, 5 awaiting decision) cover separate modules and are unaffected.

Issue #5903, opened 2026-04-16
[CS-01] Add roadmap priority policy classification tests


an0mium:
## Scope
Add comprehensive tests for `aragora/swarm/roadmap_priority.py` covering roadmap code extraction, priority classification, and the Do now/Delay/Avoid policy.

## Why (docs proof drift and roadmap truth)
The roadmap priority policy drives queue governance. Comprehensive tests ensure CS-01..03 enforcement stays accurate as the roadmap evolves.

## Files
- Add tests to `tests/swarm/test_roadmap_priority.py`

## Acceptance
- `pytest tests/swarm/test_roadmap_priority.py -v` passes
- Coverage for: extract_roadmap_codes, priority_for_text, blocks_boss_ready

an0mium:
Boss loop exhausted 3 attempts. Auto-decomposed into 1 smaller sub-issues.

Issue #5887, opened 2026-04-16
[B2-01] Document B2 guard expansion gate criteria


an0mium:
## Task
Add explicit B2 guard expansion criteria to `docs/status/NEXT_STEPS_CANONICAL.md`.

### Requirements
1. Define what "repeated bounded runs" means concretely (>=3 green weekly corpus runs)
2. Define which execution classes are safe for B2 (dependency bump, config change, fail-closed fix)
3. Keep the gate strict — no expansion until criteria are met

### File Scope
- `docs/status/NEXT_STEPS_CANONICAL.md`

### Validation
```bash
python3 scripts/reconcile_status_docs.py --json
```

### Constraints
- Docs only
- Keep claims narrower than proof

Issue #5844, opened 2026-04-15
[CS-01..03] Reconcile docs/status surfaces to current proof


an0mium:
## Goal
Reconcile roadmap/status/positioning surfaces so external claims stay narrower than current measured proof.

## Why now
The recurring docs reconciliation check found proof drift that should become exactly one bounded `CS-01..03` follow-up issue.

## Findings
- `critical` `status/NEXT_STEPS_CANONICAL.md`: Current execution docs still reference closed execution epics as active: #804 (completed), #805 (completed), #806 (completed). Replace them with live follow-up issue links or update the current gate text.
- `critical` `status/ACTIVE_EXECUTION_ISSUES.md`: Current execution docs still reference closed execution epics as active: #804 (completed), #805 (completed), #806 (completed). Replace them with live follow-up issue links or update the current gate text.

## Acceptance
- Update the affected docs/status surfaces so claims stay narrower than the current proof surfaces.
- Keep `docs/status/NEXT_STEPS_CANONICAL.md`, `docs/status/B0_BENCHMARK_TRUTH_STATUS.md`, and `docs/status/TW03_RESCUE_PRODUCTIZATION_STATUS.md` mutually consistent.
- Re-run `python3 scripts/reconcile_status_docs.py --json` until warning and critical findings are cleared.

an0mium:
Boss loop exhausted 3 attempts. Auto-decomposed into 3 smaller sub-issues with `boss-ready` label.

PR #6704, opened 2026-04-26
codex/swarm eeef7ef7 micro 3

an0mium:
- **fix: update corpus.json**
- **fix: preserve benchmark corpus freshness linkage**

### DIFF
--- a/docs/benchmarks/benchmark_corpus_freshness.json
+++ b/docs/benchmarks/benchmark_corpus_freshness.json
@@ -1,4 +1,19 @@
 {
   "schema_version": 1,
-  "entries": []
+  "entries": [
+    {
+      "corpus_id": "tw-01-bounded-execution-v1",
+      "revision": 4,
+      "stale_issue_numbers": [
+        5844,
+        5903,
+        5887
+      ],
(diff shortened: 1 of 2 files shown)


an0mium:
Closing as superseded by #6702 + #6705.

After #6702 (freshness.json entry for #5839) and #6705 (corpus.json rev-3→4 with stale-member retirement) both landed on main, this PR's unique content is no longer load-bearing:

- The rev-5 entry was a bookkeeping 'validation pass' on top of rev-4. Rev-4 (now canonical) already retires #5844, #5903, #5887 and links to #5839. No additional state captured by rev-5.
- The richer freshness.json entry (with stale_issue_numbers list + target=#5839) is a more detailed schema variant of what #6702 already added. The next recurring benchmark publication will regenerate the freshness map from corpus.json rev-4, so any genuinely useful enrichment will reappear automatically.
- Conflicts with main are real (corpus.json revision number + freshness.json entry overlap with already-merged content); rebasing would mostly delete this PR's content.

If the richer freshness schema is still wanted, the boss-loop will produce a clean follow-up PR on top of current main rather than this stale branch.

PR #6755, opened 2026-04-27
fix(automation): tolerate list evidence in outbox reconcile

an0mium:
## Summary
- make automation outbox reconciliation tolerate historical handoffs where local_evidence is a list instead of a dict
- add regression coverage for list-shaped evidence and structured-evidence precedence
- unblock stale outbox cleanup passes that were crashing before they could archive satisfied handoffs

## Validation
- python3 -m pytest -q tests/scripts/test_reconcile_automation_outbox.py
- python3 -m ruff check scripts/reconcile_automation_outbox.py tests/scripts/test_reconcile_automation_outbox.py
- python3 -m ruff format --check scripts/reconcile_automation_outbox.py tests/scripts/test_reconcile_automation_outbox.py
- git diff --check
- bash scripts/automation_pr_preflight.sh origin/main HEAD

## Non-sandbox cleanup performed
- archived 76 satisfied automation outbox handoffs after the fix allowed reconciliation to run
- refreshed automation GitHub status cache
- closed stale handoff issues #6748, #6749, #6752, and #6753 with evidence

PR #6785, opened 2026-04-28
feat(swarm): handoff_contract module skeleton (additive, no wiring)

an0mium:
## What

Pure declarative module `aragora/swarm/handoff_contract.py` (~430 LOC) plus single test fixture `tests/swarm/test_handoff_contract.py` (~330 LOC, 29 tests, one per contract clause).

This is the **first step** of the consolidation proposed in `docs/plans/2026-04-28-handoff-contract-derivation.md` (text-only spec also filed overnight). The module encodes the eight latent invariants that the 17 `fix(automation)` outbox/handoff PRs of 2026-04-21..28 converged on:

| Clause | Invariant |
|---|---|
| C1 | Idempotency-key as primary identity |
| C2 | Terminal-state precedence (4-level satisfaction predicate) |
| C3 | Outbox state-root resolution unified (deferred — out of module scope) |
| C4 | Branch field fingerprinting canonical |
| C5 | Patch-equivalence as first-class satisfaction |
| C6 | Open-PR identity beats outbox identity |
| C7 | Dry-run is read-only or it's a bug |
| C8 | Evidence schema loosely-typed but defensively-validated |

## Why this PR is safe to land out-of-band

This is a **pure module** with no callers and no behavior change:
- Does NOT import from `scripts/publish_automation_handoffs.py`
- Does NOT import from `scripts/reconcile_automation_outbox.py`
- Does NOT import from `scripts/audit_codex_branch_backlog.py`
- Does NOT touch the file system
- Does NOT call `gh` or `git`
- Does NOT mutate any state

Its only consumer is the test fixture in this same PR. Future PRs will migrate the three legacy scripts onto this contract per the spec's sequencing (steps 2-5).

## Validation

- 29/29 tests pass locally
- `ruff` clean
- `ruff format --check` clean
- `mypy --strict` clean
- `scripts/automation_pr_preflight.sh origin/main HEAD` ok
- Existing test suite untouched (additive only)

## Public surface

```python
from aragora.swarm.handoff_contract import (
    HandoffIdentity,
    InvalidHandoff,
    SatisfactionContext,
    SatisfactionKind,
    SatisfactionSignal,
    ReconcilePlan,
    ReconcilePlanEntry,
    parse_outbox_entry,
    compute_fingerprint,
    evaluate_satisfaction,
    plan_archive_satisfied,
    is_dry_run_safe,
    REQUIRED_OUTBOX_KEYS,
    PR_OPEN_REQUEST_ACTIONS,
    PR_OPEN_REQUEST_CANONICAL_ACTION,
    TERMINAL_RECEIPT_STATUSES,
)
```

All exported names are dataclasses, enums, or pure functions. No I/O, no side effects.

## Operator review

**No rush. Operator review required. Auto-merge intentionally OFF.** Filed as part of an overnight maximalist roadmap push (see `docs/plans/2026-04-28-handoff-contract-derivation.md` for the full spec and the 5-PR sequencing suggestion).

If accepted, the next bounded PR migrates one of the three legacy scripts to delegate to this module without changing observable behavior. If rejected, the existing patch trajectory continues without harm — every clause encoded here is already implemented across the three scripts; this PR just consolidates them into a single source of truth.

Co-authored-by: factory-droid[bot] <138933559+factory-droid[bot]@users.noreply.github.com>

### DIFF
--- a/aragora/swarm/handoff_contract.py
+++ b/aragora/swarm/handoff_contract.py
@@ -250,7 +250,13 @@ def parse_outbox_entry(
             raw=None,
         )
 
-    missing = tuple(k for k in REQUIRED_OUTBOX_KEYS if not payload.get(k))
+    # C8: required keys must be *present* in the payload. Falsy values are
+    # NOT treated as missing here — that would conflict with C4's top-level
+    # branch/sha fallback (which fires when local_evidence={}), and with the
+    # need to permit requires_github=False or validation=[]. Per-field
+    # semantic checks below catch the cases where a present-but-empty value
+    # is genuinely invalid (idempotency_key, requested_action).
+    missing = tuple(k for k in REQUIRED_OUTBOX_KEYS if k not in payload)
     if missing:
         return InvalidHandoff(
             reason=f"missing required keys: {', '.join(missing)}",
@@ -269,6 +275,21 @@ def parse_outbox_entry(
     branch_name = _resolve_branch_field(payload)
     head_sha = _resolve_head_sha(payload)
     action_kind = _normalize_action(payload.get("requested_action"))
+
+    # C8: enforce the action-verb whitelist. The whitelist is the contract's
+    # binding constraint on what publisher actions are valid; permitting
+    # arbitrary action_kinds defeats the purpose of having a whitelist at
+    # all. New canonical actions MUST be added to PR_OPEN_REQUEST_ACTIONS
+    # explicitly so each addition is reviewable.
+    if action_kind not in PR_OPEN_REQUEST_ACTIONS:
+        return InvalidHandoff(
+            reason=(
+                f"unknown action kind {action_kind!r}; "
+                f"must be one of: {', '.join(sorted(PR_OPEN_REQUEST_ACTIONS))}"
+            ),
+            raw=payload,
+        )
+
     fingerprint = compute_fingerprint(idempotency_key, branch_name, head_sha, action_kind)
 
     return HandoffIdentity(
--- a/aragora/swarm/handoff_contract.py
+++ b/aragora/swarm/handoff_contract.py
@@ -1,0 +1,427 @@
+"""
+Declarative contract for the local-Codex → GitHub automation handoff substrate.
+
+This module is the first read-only step of the consolidation proposed in
+``docs/plans/2026-04-28-handoff-contract-derivation.md``. It encodes the eight
+latent invariants the 17 ``fix(automation)`` outbox/handoff PRs of
+2026-04-21..28 converged on:
+
+  C1. Idempotency-key as primary identity.
+  C2. Terminal-state precedence (4-level satisfaction predicate).
+  C3. Outbox state-root resolution as a single function.
+  C4. Branch field fingerprinting canonical.
+  C5. Patch-equivalence as first-class satisfaction.
+  C6. Open-PR identity beats outbox identity.
+  C7. Dry-run is read-only or it's a bug.
+  C8. Evidence schema loosely-typed but defensively-validated.
+
+Per the spec's sequencing suggestion, this skeleton lands as a pure module
+with **no behavior change** — no script imports it yet. Its only consumer is
+the test fixture in ``tests/swarm/test_handoff_contract.py``. Future PRs will
+migrate ``scripts/publish_automation_handoffs.py``,
+``scripts/reconcile_automation_outbox.py``, and
+``scripts/audit_codex_branch_backlog.py`` onto this contract.
+
+This file is **additive** and **does not import** the legacy scripts; it
+duplicates a small amount of regex/string handling rather than circular
+import them.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass, field
+from enum import Enum
+from typing import Any, Literal
+
+# C1: idempotency keys are the canonical handoff identity. Every other
+# identity (issue title, branch, patch hash) is a secondary index.
+REQUIRED_OUTBOX_KEYS: tuple[str, ...] = (
+    "task",
+    "requires_github",
+    "requested_action",
+    "repo",
+    "local_evidence",
+    "validation",
+    "idempotency_key",
+    "created_at",
+)
+
+# C8: known canonical action verbs. Unknown actions are tolerated as JSON
+# strings whose dict has an "action" key, which keeps the contract
+# forward-compatible while surfacing diagnostics.
+PR_OPEN_REQUEST_CANONICAL_ACTION: str = "open_pr"
+PR_OPEN_REQUEST_ACTIONS: frozenset[str] = frozenset(
+    {
+        "open_pr",
+        "open_pull_request",
+        "open_or_update_pr",
+        "open_or_update_pull_request",
+        "push_branch_and_open_pr",
+        "push_branch_and_open_pull_request",
+        "push_branch_and_open_or_update_pr",
+        "push_branch_and_open_or_update_pull_request",
+    }
+)
+
+# C2: terminal receipt statuses. A receipt with any of these statuses
+# satisfies the corresponding handoff identity.
+TERMINAL_RECEIPT_STATUSES: frozenset[str] = frozenset(
+    {"published", "already_satisfied", "completed", "skipped"}
+)
+
+
+class SatisfactionKind(str, Enum):
+    """C2: enumerable satisfaction predicate kinds, in precedence order."""
+
+    TERMINAL_RECEIPT = "terminal_receipt"
+    MERGED_PR = "merged_pr"
+    PATCH_EQUIVALENT = "patch_equivalent"  # C5
+    OPEN_PR_MATCH = "open_pr_match"  # C6
+    RECEIPT_ONLY_BRANCH = "receipt_only_branch"
+
+
+@dataclass(frozen=True)
+class HandoffIdentity:
+    """C1+C4: canonical handoff identity with stable fingerprint."""
+
+    idempotency_key: str
+    branch_name: str | None
+    head_sha: str | None
+    action_kind: str
+    fingerprint: str
+
+    def to_dict(self) -> dict[str, Any]:
+        return {
+            "idempotency_key": self.idempotency_key,
+            "branch_name": self.branch_name,
+            "head_sha": self.head_sha,
+            "action_kind": self.action_kind,
+            "fingerprint": self.fingerprint,
+        }
+
+
+@dataclass(frozen=True)
+class InvalidHandoff:
+    """C8: a handoff payload that fails validation. Quarantine candidate."""
+
+    reason: str
+    missing_keys: tuple[str, ...] = ()
+    raw: Mapping[str, Any] | None = None
+
+    def to_dict(self) -> dict[str, Any]:
+        return {
+            "reason": self.reason,
+            "missing_keys": list(self.missing_keys),
+        }
+
+
+@dataclass(frozen=True)
+class SatisfactionSignal:
+    """C2: a single satisfaction predicate matched."""
+
+    kind: SatisfactionKind
+    evidence: Mapping[str, Any] = field(default_factory=dict)
+
+    def to_dict(self) -> dict[str, Any]:
+        return {
+            "kind": self.kind.value,
+            "evidence": dict(self.evidence),
+        }
+
+
+@dataclass(frozen=True)
+class SatisfactionContext:
+    """Read-only inputs to ``evaluate_satisfaction``.
+
+    Concrete callers populate this from disk (receipt directory) and from
+    the network (open PR heads). The contract module itself does NOT touch
+    the file system or call gh; it operates only on the inputs supplied
+    here.
+    """
+
+    terminal_receipt_keys: frozenset[str] = frozenset()
+    merged_pr_branches: frozenset[str] = frozenset()
+    patch_equivalent_branches: frozenset[str] = frozenset()
+    open_pr_heads: Mapping[str, int] = field(default_factory=dict)
+    receipt_only_branches: frozenset[str] = frozenset()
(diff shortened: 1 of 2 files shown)


an0mium:
## Codex review (comment-only; no self-approval)

LGTM as an additive, no-caller skeleton for the handoff/outbox contract. The module is pure, has no filesystem/GitHub/git side effects, and the tests cover the intended invariants: idempotency key identity, satisfaction precedence, branch/head extraction drift, patch-equivalent/open-PR/receipt-only satisfaction, dry-run plan enumeration, and loose evidence schema validation.

Non-blocking design note before wiring callers: `_normalize_action()` currently preserves action aliases like `open_or_update_pr` rather than canonicalizing aliases in `PR_OPEN_REQUEST_ACTIONS` to `PR_OPEN_REQUEST_CANONICAL_ACTION`. That is fine while this module has no callers, but if a future migration expects aliases from historical handoff schemas to dedupe as the same action, canonicalization should happen before fingerprints are used as reconciliation identity.

No merge blocker in this PR because it intentionally does not wire behavior. Safe to merge after an independent/non-author review.

an0mium (commented):
## Read-only deep review (Claude, load-bearing contract review)

Reviewed `/tmp/handoff_contract.py` (427 lines), `/tmp/test_handoff_contract.py` (330 lines), and `docs/plans/2026-04-28-handoff-contract-derivation.md` (the spec, on #6787 branch). This is **load-bearing review, not style review** — the contract shape will propagate into 3 wiring PRs (publisher, reconciler, auditor). Findings below are ordered by downstream-wiring impact.

### Verdict: 6 findings, 2 of them genuine contract-shape issues worth fixing here

The 8 invariants C1-C8 are a sound boundary and they map cleanly back to the 17 source PRs. The dataclasses are frozen, the satisfaction precedence enum is well-named, and the module is correctly pure (zero I/O). 29 tests pass.

But there are two real defects in the C8 validation pathway that break C4's fallback contract, and four lesser issues worth pre-empting before the wiring PRs.

---

### B1 (BLOCKER) — `parse_outbox_entry`'s falsy-empty check breaks the C4 top-level fallback contract

**Location:** `aragora/swarm/handoff_contract.py:253`

```python
missing = tuple(k for k in REQUIRED_OUTBOX_KEYS if not payload.get(k))
```

`not payload.get(k)` treats *any falsy value* as missing — empty dict, empty list, `False`, `0`, `""`. Two consequences:

**B1.a — Spec C4 fallback is unreachable.** The spec says (lines 137-142):

> Schema drift in the outbox file (e.g., the `local_evidence` field becoming a list, see PR #6755) MUST be tolerated by the fingerprint extractor: it falls back to the top-level fields and emits a diagnostic warning, never silently mismatches.

The implementation's `_resolve_branch_field` (lines 176-195) and `_resolve_head_sha` (lines 198-212) **do** implement that fallback. But `parse_outbox_entry` rejects the payload at line 253 before the fallback ever fires, because `local_evidence = {}` is falsy.

The test `test_branch_resolution_top_level_fallback` (`tests/swarm/test_handoff_contract.py:140-148`) actually documents this bug and asserts the buggy behavior:

```python
def test_branch_resolution_top_level_fallback(self) -> None:
    payload = _valid_payload()
    payload["local_evidence"] = {}  # no branch here
    payload["branch"] = "feat/top-level"
    ident = parse_outbox_entry(payload)
    # local_evidence is empty mapping — invalid per C8.
    assert isinstance(ident, InvalidHandoff)
```

The comment "invalid per C8" contradicts the spec's C4 + C8 reading: the schema validation is supposed to permit empty `local_evidence` *as long as* the top-level fallback can resolve the branch.

**B1.b — `requires_github=False` is silently invalidated.** Same root cause. `not False` is `True`, so any handoff with `requires_github: False` is rejected as "missing required keys". Practically all current handoffs set `True`, but this bakes "all handoffs need GitHub" into the contract by accident. If automation ever produces a local-only validation handoff, the contract silently rejects it.

**Suggested fix:** change line 253 from `not payload.get(k)` to `k not in payload` (or `payload.get(k) is None`), and update C8's "non-empty" assertion to be a per-key explicit check (e.g. `idempotency_key`/`task`/`repo` must be non-empty strings; `local_evidence` may be empty if top-level branch/sha are present; `requires_github` may be `False`; `validation` may be `[]`). Update `test_branch_resolution_top_level_fallback` to assert `HandoffIdentity` with `branch_name == "feat/top-level"` instead of `InvalidHandoff`.

---

### B2 (BLOCKER) — `PR_OPEN_REQUEST_ACTIONS` is defined and exported but unenforced

**Location:** `aragora/swarm/handoff_contract.py:56-67`, `aragora/swarm/handoff_contract.py:238-280`

The constant is exported in `__all__`. Spec C8 (lines 192-196) says:

> `requested_action` is either a known action verb (the union in `PR_OPEN_REQUEST_ACTIONS`) or a JSON-string of a Mapping with an `action` key

But `parse_outbox_entry` never checks the resolved `action_kind` against `PR_OPEN_REQUEST_ACTIONS`. A handoff with `"requested_action": "frobnicate"` parses cleanly with `action_kind = "frobnicate"`. The whitelist is enforcement-free.

Two valid resolutions:
- **Option A (enforce):** `parse_outbox_entry` returns `InvalidHandoff` if the resolved `action_kind` is not in `PR_OPEN_REQUEST_ACTIONS`. Update spec C8 to make this explicit.
- **Option B (advisory):** Remove `PR_OPEN_REQUEST_ACTIONS` from `__all__` and add a docstring to it noting it's a reference list, not enforcement. Update spec C8 to drop the "known action verb" clause.

Either is fine. Picking neither leaves a dead constant in the public surface.

---

### F3 (NON-BLOCKER, future-wiring gap) — `plan_archive_satisfied` doesn't model the publish path

**Location:** `aragora/swarm/handoff_contract.py:378-407`

The skeleton models the **reconciler's** archive-vs-skip decision. The **publisher's** decision is fundamentally different: when a handoff is *not* satisfied, should it `publish` (open a new PR) or `skip` (e.g., dry-run, base-ref check failure)? The publisher is also the riskier wiring (it's the script that actually creates GitHub state).

The docstring acknowledges this:

> The publisher's `open_pr` step is out of scope for this skeleton.

But `ReconcilePlanEntry.action: Literal["archive", "publish", "skip"]` includes `publish` — a value that's never produced by any function in this module. That asymmetry will confuse the publisher's wiring PR: the type system says "publish is a valid action" but the contract module provides no helper to plan it.

**Suggested:** either drop `publish` from the Literal until the publisher-wiring PR adds it, or add a sibling `plan_publish_unsatisfied` skeleton (even if it just returns `publish` for unsatisfied handoffs and leaves the actual publication to the caller). Pick one.

---

### F4 (NON-BLOCKER) — `is_dry_run_safe` is tautologically True; consider removing or renaming

**Location:** `aragora/swarm/handoff_contract.py:366-375`

The function always returns `True` regardless of plan content (the docstring is honest about this). The contract's actual dry-run-safety invariant lives at the **callers'** mutation chokepoints, not in this module — which means this function adds no enforcement value. The test `test_pure_module_is_dry_run_safe` is a tautology assertion.

Two options:
- **Drop the function**, replace with a class-level docstring on `ReconcilePlan` explaining that dry-run safety is a caller obligation.
- **Keep but rename** to something like `plan_is_pure` (descriptive of what it actually checks) and add a TODO that it should validate caller-supplied mutation chokepoints when those land.

Not a merge blocker — it just won't ever catch a real dry-run violation in its current form.

---

### F5 (NON-BLOCKER) — Fingerprint exists but isn't used in satisfaction evaluation

**Location:** `aragora/swarm/handoff_contract.py:215-235`, `aragora/swarm/handoff_contract.py:283-330`

C4 says fingerprint is the "secondary index used only to detect prior satisfaction when the key alone isn't sufficient." `compute_fingerprint` is implemented and is in `__all__`, but `evaluate_satisfaction` only uses `idempotency_key` and `branch_name` — never `fingerprint`.

This is fine for the skeleton but it means the C4 secondary-index path isn't wired even at the contract level. When the publisher migrates and encounters the case where two handoffs share an `idempotency_key` but differ in fingerprint (schema drift across the same logical handoff), the contract has no helper to disambiguate. Worth either:
- Adding `terminal_receipt_fingerprints` to `SatisfactionContext` and a fingerprint-fallback branch in `evaluate_satisfaction`, OR
- Removing `compute_fingerprint` from `__all__` until it has a consumer.

---

### F6 (NON-BLOCKER) — Edge cases the test fixture should cover before downstream wiring

The 29 tests cover the documented invariants well. Three edge cases I'd add to harden the contract before a publisher/reconciler PR consumes it:

- `parse_outbox_entry` with `requires_github=False` (will fail today per B1.b — correct response after fix is `HandoffIdentity`)
- `parse_outbox_entry` with `validation=[]` (should be valid; falsy-check rejects today)
- `evaluate_satisfaction` where `branch_name=None` and only `terminal_receipt_keys` matches — confirming branchless handoffs can still be satisfied via idempotency_key alone (this works today; just add the test)

---

## Bottom line

**Spec/implementation alignment is strong** on C1, C2, C5, C6, C7. The dataclasses, enum precedence, and pure-function discipline are right.

**B1 and B2 are real contract-shape issues** that should be resolved in this PR, not deferred. They're cheap fixes (~20 LOC + test updates) and they propagate into every wiring PR if left.

F3-F6 are non-blockers that can be deferred to the wiring PRs, but flagging them now makes the contract surface tighter and reduces re-litigation when those PRs land.

**Not a self-approving review** — Claude and the PR author share an authorship account. Operator must approve and merge.

— posted as part of standing read-only watch on the open queue

an0mium:
## B1 + B2 fixes pushed (commit `b4005c55a`)

Both blocker findings from the earlier review addressed in-place. Diff: +91/-3 across the same 2 files.

### B1 — falsy-empty validation broke C4 fallback
- `parse_outbox_entry` line 253: `not payload.get(k)` → `k not in payload`
- Now only absent keys are missing; present-but-falsy (`requires_github=False`, `validation=[]`, `local_evidence={}`) pass through to per-field semantic checks
- The `idempotency_key` strip+empty check at line 262 still rejects empty/whitespace keys

### B2 — PR_OPEN_REQUEST_ACTIONS now enforced
- After `_normalize_action` resolves `action_kind`, the parser rejects with `InvalidHandoff` if it's not in the whitelist
- Catches all three input shapes: plain string `"frobnicate"`, Mapping `{"type": "frobnicate"}`, JSON-string `'{"action": "frobnicate"}'`
- New canonical actions must now be added to the constant explicitly — the whitelist becomes a binding contract

### Test coverage added (7 new, total 36/36 passing)

| Test | Asserts |
|---|---|
| `test_branch_resolution_top_level_fallback` (updated) | HandoffIdentity with `branch_name=feat/top-level` (was: InvalidHandoff) |
| `test_branch_resolution_falsy_evidence_does_not_invalidate` | `local_evidence=[]` doesn't pre-empt C4 |
| `test_branch_resolution_precedence_local_evidence_beats_requested_action` | clarifies the precedence: local_evidence > requested_action.branch > top-level branch |
| `test_unknown_action_kind_invalidates` | plain-string `frobnicate` rejected |
| `test_unknown_action_kind_in_mapping_invalidates` | Mapping form rejected |
| `test_unknown_action_kind_in_json_string_invalidates` | JSON-string form rejected |
| `test_requires_github_false_is_valid` | `requires_github=False` parses successfully |
| `test_validation_empty_list_is_valid` | `validation=[]` parses successfully |

### Validation
- `pytest tests/swarm/test_handoff_contract.py` — 36/36 pass
- `ruff check` + `ruff format --check` — clean
- `mypy aragora/swarm/handoff_contract.py tests/swarm/test_handoff_contract.py` — clean

### Out of scope (deferred to wiring PRs)
F3 (`plan_archive_satisfied` doesn't model publish path), F4 (`is_dry_run_safe` tautology), F5 (fingerprint exists but not consumed by `evaluate_satisfaction`) — these are non-blocking contract-shape comments that can land with the publisher/reconciler/auditor migration PRs.

Pinging @synaptent/aragora-reviewers for re-review.

an0mium:
## Codex recheck after B1/B2 fix

No new findings on the handoff-contract B1/B2 patch at commit `b4005c55a`.

Validation run from the PR worktree:

- `python3 -m pytest tests/swarm/test_handoff_contract.py -q` -> `36 passed`
- `python3 -m ruff check aragora/swarm/handoff_contract.py tests/swarm/test_handoff_contract.py` -> passed
- `python3 -m ruff format --check aragora/swarm/handoff_contract.py tests/swarm/test_handoff_contract.py` -> passed
- `git diff --check` -> clean

The spec interpretation is sound: `parse_outbox_entry()` is the contract boundary, so unknown `requested_action` verbs should produce `InvalidHandoff`; present-but-falsy required fields like `local_evidence={}`, `requires_github=False`, and `validation=[]` should not be treated as missing.

Remaining design note is non-blocking for this pure skeleton: before real publisher/reconciler wiring, decide whether accepted aliases preserve action identity in fingerprints or canonicalize to `open_pr`.

Comment-only review; no approval/merge/auto-merge action taken.

an0mium:
## Codex cross-author publisher-harness dogfood

Ran an additional temp-script adversarial fixture against commit `b4005c55a`, focused on the corrected C4 fallback and C8 whitelist contract.

Command:

`PYTHONPATH=. python3 /tmp/dogfood_6785.py`

Result: `dogfood_6785: PASS`

Scenarios covered:
- `local_evidence={}` + plain `requested_action="open_pr"` falls back to top-level `branch`
- `local_evidence=[]` falls back to `requested_action.branch`
- `requires_github=False` remains valid, not missing
- `validation=[]` remains valid, not missing
- branch precedence is `local_evidence.branch > requested_action.branch > top-level branch`
- unknown actions are rejected for all three shapes: plain string, Mapping `{type: ...}`, JSON-string `{action: ...}`

This clears my remaining cross-author dogfood concern for #6785. Comment-only review; no approval/merge/auto-merge action taken.

PR #6806, opened 2026-04-29
fix(debate): preserve rounds_completed across partial-round failures (B2/B4)

an0mium:
## Summary

Dogfood-discovered (2026-04-28 evolution-round) bug: a debate that crashed in critique/revision/novelty tracking would report \`rounds_used == rounds_completed == 0\` in the finalized result and receipts, even though one or more rounds had produced meaningful work (synthesis, winner, critiques).

The trigger in our case was an expired Gemini embedding API key surfacing as a \`RuntimeError\` from \`compute_novelty()\`, which propagated up through \`track_novelty()\` and aborted the entire debate-rounds phase.

## Root cause

Three layers all assumed the success path:

1. \`convergence_tracker.track_novelty\` did not wrap \`compute_novelty()\` in try/except. An embedding-service failure (Gemini key expired, rate-limited, transient network) would re-raise into the round loop and abort the phase.
2. \`debate_rounds._execute_round\` only set \`result.rounds_used = round_num\` *after* the round's revision phase fully completed. A failure earlier in the round left no record of attempted progress.
3. \`DebateState.finalize_result\` resolved \`rounds_used\` from \`current_round or result.rounds_used\`. Both are 0 when a round fails partway, so the finalized receipt under-reported.

## Minimal repro (pre-fix)

Mock \`novelty_tracker.compute_novelty\` to raise \`RuntimeError(\"embedding service down\")\` on round 1. The Arena produces a synthesis and winner from the proposal phase, but \`result.rounds_used\` and \`result.rounds_completed\` are both 0.

## Fix

Three coordinated changes:

1. **\`aragora/debate/phases/convergence_tracker.py\`** — wrap \`compute_novelty()\` in try/except. On failure, log a warning, append a structured entry to \`result.metadata['novelty_tracker_failures']\` (round, error_class, error_message), and return early. Novelty tracking is observational; an external embedding outage must not abort the debate-rounds phase.

2. **\`aragora/debate/phases/debate_rounds.py\`** — set \`ctx.partial_rounds = round_num\` at the START of each round so a round that fails after partial work still contributes to round counting.

3. **\`aragora/debate/debate_state.py\`** — \`finalize_result\` resolves \`rounds_used\` by precedence: \`self.current_round\` (graph-style flows) → \`self.result.rounds_used\` (success path) → \`self.partial_rounds\` (partial-round fallback).

## Tests

5 new regression tests, all green:

- \`tests/debate/phases/test_convergence_tracker.py\` (2 new)
[...]

PR #6834, opened 2026-04-30
fix(debate): floor adaptive rounds at 1 to preserve rounds_completed honesty

an0mium:
## Summary

Round 2026-04-30b — Phase B (Tier 2). Closes the `rounds_completed=0` bug root-caused in 2026-04-30 Phase E (`.aragora/evolve-round/2026-04-30/dogfood/phase-e-rounds-zero-rca.md`).

## Bug

`DebateStrategy.estimate_rounds_async()` could return `estimated_rounds=0` (interpreted as "proposal already produced consensus, no critique-revision rounds needed"). The previous code passed this through verbatim, and the round-loop in `DebateRoundsPhase.execute()` would skip entirely:

```python
for round_num in range(1, rounds + 1):  # range(1, 1) yields nothing
```

Downstream consumers saw `rounds_completed=0` and inferred "no debate happened" — even when a winner was extracted from the proposal. This was a metric/observability bug, not a correctness bug on the proposal path.

PR #6806's `partial_rounds`-at-round-start fix is correct but doesn't fire in this scenario because `_execute_round` is never entered.

## Fix

Option I from the RCA: floor the strategy recommendation at 1. The strategy's signal is preserved verbatim in `result.metadata["strategy_recommendation"]["estimated_rounds"]`, plus a new `floored_to_one` bool that downstream consumers can filter on.

## Tests

- `test_strategy_zero_rounds_falls_back_to_protocol_default` — regression for the literal RCA scenario: `estimated_rounds=0` must produce `rounds_used >= 1` + `floored_to_one=True` in metadata
- `test_strategy_negative_rounds_falls_back_to_protocol_default` — defensive case for a future strategy bug

## Verification

- 101/101 tests pass on `test_debate_rounds.py` + `test_debate_rounds_invariant.py`
- ruff check + format clean
- mypy clean on `debate_rounds.py`
- preflight ok

## Diff stat

```
 aragora/debate/phases/debate_rounds.py   |  19 ++--
 tests/debate/phases/test_debate_rounds.py | 120 +++++++++++++++++++++++++++++++
 2 files changed, 138 insertions(+), 1 deletion(-)
```

## Round invariants

- Tier 2 (debate orchestrator hot path)
- No author-merge, no draft-flip, no GitHub review-state set
- 101 tests pass deterministically

### DIFF
--- a/aragora/debate/phases/debate_rounds.py
+++ b/aragora/debate/phases/debate_rounds.py
@@ -463,7 +463,24 @@ class DebateRoundsPhase:
                         strategy_rec.reasoning[:50],
                     )
                     _record_adaptive_round(direction)
-                    rounds = strategy_rec.estimated_rounds
+                    # Floor adaptive recommendation at 1: a recommendation of 0
+                    # (interpreted as "the proposal already produced consensus, no
+                    # critique-revision rounds needed") would skip the round loop
+                    # entirely and report rounds_completed=0 to downstream consumers,
+                    # making a real debate look like one that never ran. We honour
+                    # the strategy's signal in metadata but always run at least one
+                    # round so the metric is honest. See
+                    # docs/plans/2026-04-30-rounds-floor-rca.md
+                    # (or .aragora/evolve-round/2026-04-30/dogfood/phase-e-rounds-zero-rca.md
+                    # in repos that carry the round receipts).
+                    if strategy_rec.estimated_rounds < 1:
+                        logger.info(
+                            "[strategy] estimate=%s < 1; flooring at 1 to preserve rounds_completed honesty",
+                            strategy_rec.estimated_rounds,
+                        )
+                        rounds = max(1, rounds)
+                    else:
+                        rounds = strategy_rec.estimated_rounds
                     # Store strategy recommendation in result metadata
                     if hasattr(result, "metadata") and result.metadata is not None:
                         result.metadata["strategy_recommendation"] = {
@@ -471,6 +488,7 @@ class DebateRoundsPhase:
                             "confidence": strategy_rec.confidence,
                             "reasoning": strategy_rec.reasoning,
                             "relevant_memories": strategy_rec.relevant_memories[:3],
+                            "floored_to_one": strategy_rec.estimated_rounds < 1,
                         }
             except (RuntimeError, AttributeError, TypeError) as e:  # noqa: BLE001
                 logger.debug("[strategy] Round estimation failed, using protocol default: %s", e)
--- a/tests/debate/phases/test_debate_rounds.py
+++ b/tests/debate/phases/test_debate_rounds.py
@@ -1513,6 +1513,125 @@ class TestStrategyBasedRounds:
         # Default rounds should be used
         assert ctx.result.rounds_used == 1
 
+    @pytest.mark.asyncio
+    async def test_strategy_zero_rounds_falls_back_to_protocol_default(self):
+        """Strategy returning estimated_rounds=0 must not skip the round loop.
+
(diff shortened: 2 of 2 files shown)


an0mium:
## Claude review (author-side)

**Head SHA:** `e72deae1f7f0e382f12ddb6a5331b9af4ff57168`
**Tier:** 2 (debate orchestrator hot path)
**Verdict:** clean.

### Verifications

- 101/101 tests pass on `test_debate_rounds.py` + `test_debate_rounds_invariant.py` (4 new + 97 existing)
- ruff check + format clean
- mypy clean on `debate_rounds.py`
- preflight ok, gitleaks ok

### Why this is the right shape

The fix is a 6-line change inside the existing `if strategy_rec.estimated_rounds != rounds:` block. It:

1. Floors the recommendation at 1 when `< 1` so the round loop iterates at least once
2. Preserves the full strategy signal (raw estimate, confidence, reasoning) in `result.metadata["strategy_recommendation"]`
3. Adds a new `floored_to_one` bool for downstream consumers that want to filter on the corrected case

This means analyses that *want* to know "the strategy thought zero rounds were needed" can still see it via `metadata.strategy_recommendation.estimated_rounds == 0 and floored_to_one == True`.

### Why option I, not option II

Per the RCA, option II (introduce `rounds_skipped_by_strategy` and redefine `rounds_completed = max(1, rounds_used)`) would be a broader change touching `DebateResult` and every consumer that reads `rounds_completed`. Option I is the safer first step and immediately restores the dogfood signal. Option II remains available as a future refinement if needed.

### Non-blocking follow-ups (out of scope)

1. **The same floor applies to the `else` branch.** If `strategy_rec.estimated_rounds == rounds` (no change branch), no floor is needed because `rounds` was the protocol default which is already `>= 1`. But if a future protocol allows `rounds=0`, the floor would need to extend to that path too.
2. **Strategy could return non-integer estimated_rounds.** Type system says `int`, but a defensive `int()` cast is implicit by the comparisons. Cosmetic.

Standing rule: I do not author-merge.

an0mium:
## Codex focused dogfood

Focused dogfood review for PR #6834.

Head SHA verified: `e72deae1f7f0e382f12ddb6a5331b9af4ff57168`

Scope reviewed:
- `aragora/debate/phases/debate_rounds.py`
- `tests/debate/phases/test_debate_rounds.py`

Verdict: no blocking issues found. The change is bounded to adaptive-round accounting: effective max rounds is floored at 1 so debate execution does not honestly report `rounds_completed=0` for a route that should have attempted at least one adaptive round. This repairs reporting semantics without broadening dispatch, persistence, security, reputation, or market behavior.

Commands run from the PR worktree:
```bash
git status --short --branch
git rev-parse HEAD
PYTHONDONTWRITEBYTECODE=1 python3 -m pytest -q tests/debate/phases/test_debate_rounds.py -p no:rerunfailures -p no:cacheprovider
python3 -m ruff check aragora/debate/phases/debate_rounds.py tests/debate/phases/test_debate_rounds.py
python3 -m ruff format --check aragora/debate/phases/debate_rounds.py tests/debate/phases/test_debate_rounds.py
bash scripts/automation_pr_preflight.sh origin/main HEAD
```

Observed results:
- Focused tests: `95 passed`.
- Ruff: passed.
- Format check: passed.
- Automation preflight: passed.

Risk assessment:
- No new external I/O or production mutation path.
- No market/reputation flow enablement.
- Behavioral effect is constrained to internal debate round-counting honesty and regression coverage.

Non-blocking follow-up: if future adaptive-round code allows zero rounds intentionally for a skip/short-circuit mode, that should be represented as a distinct explicit skip reason rather than overloading `rounds_completed=0`.

PR #6979, opened 2026-05-03
fix(automation): treat empty branch publisher queue as noop

an0mium:

PR #6958, opened 2026-05-02
fix(automation): expose receipted outbox count

an0mium:
## Summary
- expose terminal_receipted_outbox_count in the local automation queue cache
- compute unreceipted_outbox_count from cached outbox keys to clarify receipted residue
- add focused coverage for receipted and nonterminal receipt cases

## Validation
- PYTHONDONTWRITEBYTECODE=1 python3 -m pytest -q tests/scripts/test_cache_codex_automation_github_status.py -p no:rerunfailures -p no:cacheprovider
- python3 -m ruff check scripts/cache_codex_automation_github_status.py tests/scripts/test_cache_codex_automation_github_status.py
- python3 -m ruff format --check scripts/cache_codex_automation_github_status.py tests/scripts/test_cache_codex_automation_github_status.py
- python3 scripts/cache_codex_automation_github_status.py --repo . --json --output /private/tmp/aragora-github-status-cache-receipted-postpatch.json
- git diff --check origin/main...HEAD
- bash scripts/automation_pr_preflight.sh origin/main HEAD

PR #6984, opened 2026-05-03
fix(automation): accept direct dot-aragora state roots

an0mium:


an0mium:
## Codex focused dogfood — #6984

Independent focused review of PR #6984 at exact head `f34a871793aecc95eb4da8c6879282e3aaf145a6`.

**Scope reviewed:** `scripts/cache_codex_automation_github_status.py` and `tests/scripts/test_cache_codex_automation_github_status.py` only.

### What changed

The PR updates `_shared_state_root(repo_root)` so `ARAGORA_AUTOMATION_STATE_ROOT` can point either to:

- a repo/state root containing `.aragora/`, or
- the `.aragora` directory itself.

If the env var points directly at `.aragora`, the helper returns its parent so existing default path joins still resolve to `<root>/.aragora/automation-outbox`, `<root>/.aragora/automation-receipts`, and `<root>/.aragora/automation-github-status/latest.json`. This preserves the existing contract for callers that already pass the parent repo/root path.

### Validation run locally

```text
git rev-parse HEAD
# f34a871793aecc95eb4da8c6879282e3aaf145a6

python3 -m pytest tests/scripts/test_cache_codex_automation_github_status.py -q -p no:randomly
# 6 passed in 1.14s

python3 -m ruff check scripts/cache_codex_automation_github_status.py tests/scripts/test_cache_codex_automation_github_status.py
# All checks passed!

python3 -m ruff format --check scripts/cache_codex_automation_github_status.py tests/scripts/test_cache_codex_automation_github_status.py
# 2 files already formatted

bash scripts/automation_pr_preflight.sh origin/main HEAD
# preflight: ok
```

Additional smoke verified a direct `.aragora` env var resolves back to the parent and reads the expected shared outbox/receipt directories.

### CI note

Full PR CI is green except two cancelled non-required `check` rows:

- `Module Tier Drift` cancelled
- `Metrics Drift` cancelled

Those are not evidence of a #6984 code failure. The repo already has a separate main-level metrics drift from #6979 (`CLI top-level command modules 61 -> 62`), which is being handled separately.

### Verdict

No code blocker found. This is a narrow automation-cache correctness fix, covered by a direct regression test, and aligned with the existing shared-state-root behavior.

an0mium:
## Claude focused dogfood — second independent signal

**Head SHA:** `f34a871793aecc95eb4da8c6879282e3aaf145a6` (unchanged during review)

**Diff scope (2 files, +40/-2):**
- `scripts/cache_codex_automation_github_status.py` (`_shared_state_root`): when `ARAGORA_AUTOMATION_STATE_ROOT` is set, the resolver now (a) wraps `Path(...).resolve()` in a `try/except OSError` so a missing path falls back to the unresolved form instead of crashing, (b) accepts the env var pointing **directly** at a `.aragora` directory by walking to its parent, in addition to the prior contract of pointing at a parent that contains a `.aragora` subdir.
- `tests/scripts/test_cache_codex_automation_github_status.py`: adds `test_local_queue_state_accepts_direct_dot_aragora_state_root` which sets `ARAGORA_AUTOMATION_STATE_ROOT` to a literal `.../shared/.aragora` path and asserts `outbox_dir` resolves to the configured `.aragora/automation-outbox` and the receipt iteration produces the expected counts.

**Safety review (operability + observability brief):**
- The new `OSError` catch is narrow and only used to fall back from a failed `.resolve()` to the original `configured_path`. No swallowing of broader exception classes; the subsequent `is_dir()` check still discriminates real-vs-phantom paths.
- The new direct-`.aragora` branch only triggers when `resolved.name == ".aragora"` AND `resolved.is_dir()`. There is no path-traversal exposure: `resolved.parent` is the canonical parent of an existing real directory.
- Pre-existing semantics preserved: if the env var points at a parent that contains `.aragora` (the prior contract), the second `if (resolved / ".aragora").is_dir()` branch still fires unchanged, and downstream `outbox` / `receipts` derivation is identical.
- No mutation surface added: this is read-only state-root resolution. No outbox writes, no receipt JSON mutation, no destructive cleanup, no secrets.

**Commands run:**
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` -> `preflight: ok`
- `python3 -m ruff check` on both touched files -> All checks passed
- `python3 -m ruff format --check` on both touched files -> 2 files already formatted
- `pytest tests/scripts/test_cache_codex_automation_github_status.py -q --timeout=60` -> **6 passed in 2.11s** (includes the new direct-`.aragora` test plus the 5 pre-existing assertions)
- `gitleaks detect --no-git --source <diff.patch>` -> 0 findings

**Verdict:** clean. Tier-A green. Second independent model signal recorded for quorum.

**Non-blocking follow-ups:**
1. The new `OSError` catch returns the unresolved `configured_path` and then re-uses it in the next two `if` branches. If `configured_path` is a non-absolute or symlink-target, the unresolved fallback may behave subtly differently from a successful `.resolve()` on retry. Pragmatic for now (the only realistic `OSError` here is a permission/missing intermediate), but worth a one-line comment explaining the deliberate fallback choice.
2. The two `is_dir()` branches are now sequential rather than alternative -- a directory whose name is literally `.aragora` AND that also contains a `.aragora` subdir would only match the first branch (correct: prefer the direct hit). If anyone ever creates that nested shape intentionally, behavior is well-defined; no action needed.
3. The new test fixture writes a receipt-shaped JSON payload (`{"idempotency_key": "..."}`) but doesn't exercise the receipt-key extraction code path. That's outside this PR's scope, but a future test could pair the env-root resolution with the receipted-outbox-count derivation that #6958 just landed.

(Comment posted by the Claude reviewer droid against `f34a871793`. Re-running the review packet against a newer head would require a fresh queue prompt.)

PR #7034, opened 2026-05-05
fix(agent-bridge): skip allow-impact summary chrome

an0mium:

PR #7058, opened 2026-05-06
codex/agent bridge session control summary refresh 20260506

an0mium:
- **fix(agent-bridge): skip permission mode summaries**
- **fix(agent-bridge): skip session cancel summaries**


an0mium:
Closing as superseded by #7034 at head 66c9acdefb7ecec14f16018ddea728968561c1c0. #7058 is conflicting against current main, and the only unique behavior I found while comparing it to #7034 was the permission-dialog `No, cancel ...` chrome filter. That case has now been salvaged into #7034 with a test and validated there. Branch not deleted.

PR #7067, opened 2026-05-07
fix(status): correct rev4 evidence gap target

an0mium:
## Summary
- Correct the rev-4 staging corpus row whose `issue_id` pointed at `#5126` while its source reference and title are for `#5200`.
- Add a manifest invariant that every `source_reference` issue number matches `issue_id`.
- Re-render H1-01 rev-4 readiness: metrics-backed evidence is now 13/15, so the remaining promotion gap is 2 rows, and the status doc now labels the list as metrics evidence gaps rather than fresh dispatch targets.

## Context
#7060 merged before this repair landed. This is the narrow follow-up for the stale proof-loop target it exposed; it does not add substrate, does not run agents, and does not mutate receipts.

## Validation
- `pytest tests/benchmarks/test_corpus_rev4_manifest.py tests/scripts/test_render_rev4_promotion_readiness.py -q -p no:randomly` -> `14 passed`
- `ruff check scripts/render_rev4_promotion_readiness.py tests/benchmarks/test_corpus_rev4_manifest.py tests/scripts/test_render_rev4_promotion_readiness.py` -> `All checks passed!`
- `ruff format --check scripts/render_rev4_promotion_readiness.py tests/benchmarks/test_corpus_rev4_manifest.py tests/scripts/test_render_rev4_promotion_readiness.py` -> `3 files already formatted`
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` -> `preflight: ok`
- `gitleaks detect --redact --verbose --source /private/tmp/aragora-pr7060-repair --log-opts origin/main..HEAD` -> `no leaks found`

PR #7068, opened 2026-05-07
fix(status): correct rev4 evidence gap target

an0mium:


an0mium:
Closing as duplicate of #7067. The patch content is equivalent; #7067 has the complete body and Codex focused dogfood at head 57b2f70dd8fb31f9194c897304f65b1eda8752ee.

PR #7073, opened 2026-05-07
fix(automation): accept receipt dir in freshness check

an0mium:

PR #7074, opened 2026-05-07
fix(automation): accept freshness receipt dir option

an0mium:


an0mium:
Closing as superseded by #7073. #7074 is the older alternate wording for the same publisher freshness --receipt-dir compatibility behavior; #7073 is merged on main with the CLI option, status-cache-only non-claim in help text, and passing coverage. #7074 has no remaining distinct behavior to merge.

PR #7100, opened 2026-05-12
fix(automation): treat empty handoff publisher as noop

an0mium:


an0mium:
Closing as superseded by merged PR #7103. Local comparison showed #7100 and #7103 both change the same publisher empty-queue/GitHub-unavailable path and tests; #7103 delivered the same contract (empty handoff queue returns 0, non-empty queue still returns 1, JSON health payload preserved) with the more informative noop log message. No branch deletion performed.

PR #7103, opened 2026-05-12
fix(automation): noop empty handoff publishing

an0mium:


an0mium:
## Codex focused dogfood

Exact head: `261b6b4dc046f4624ef7d0f1d1f614d3066a10b5`
Worktree: `/private/tmp/aragora-pr7103-exact.tXR49q` detached at the PR head.

Validation:
- `pytest tests/scripts/test_publish_automation_handoffs.py` -> 46 passed in 2.76s
- `ruff check scripts/publish_automation_handoffs.py tests/scripts/test_publish_automation_handoffs.py` -> clean

Review focus: publisher exits 0 only for the empty-handoff/noop path when GitHub is unavailable; non-empty handoff queues still return nonzero and keep the existing GitHub-unavailable payload path.

an0mium:

**Scope:** Reviewed the patch at exact head `261b6b4dc046f4624ef7d0f1d1f614d3066a10b5` (read-only). Validation evidence accepted: 46 passed in `tests/scripts/test_publish_automation_handoffs.py`, ruff clean on both touched files.

**Behavioral analysis (`scripts/publish_automation_handoffs.py` `main`):**

- **Empty queue + GitHub unavailable -> exit 0 (noop):** Correct. There is no work to publish, so there is nothing to fail closed on. Returning nonzero in this state would have caused cron/CI noise without any actionable signal. The human-readable line `noop: no handoffs to publish; github_unavailable=<mode>` still surfaces the degraded GitHub state for operators tailing logs.
- **Non-empty queue + GitHub unavailable -> exit 1 (fail closed):** Preserved by the `return 1 if handoffs else 0` guard, with the original `github_unavailable: <mode> <error>` line intact. No regression in the real-work-is-being-dropped path: handoffs are not silently consumed and no receipts are written because the early return remains before `decide_handoffs`.
- **JSON payload semantics:** Unchanged. The same `payload`, including `github_health` and `handoff_count`/`decisions`, is emitted in both noop and fail-closed cases; only the exit code differs. The new test pins `handoff_count == 0`, `decisions == []`, and `github_health.mode == "connectivity_failed"`, which is the right contract for downstream consumers diffing on `github_health` rather than exit code.
- **Test update:** The flip from `exit_code == 1` to `exit_code == 0` in the existing test is consistent with the new contract, because that test exercises the empty-queue path, and the added `test_main_treats_empty_handoff_queue_as_noop_when_github_unavailable` locks in the JSON shape so future refactors cannot quietly drop `github_health`.

**Safety / regression checks:**

- No state mutation occurs on the noop path: no outbox consumption, no receipts written, no PRs opened. The early return is before `decide_handoffs` and `execute_decisions`.
- No change to the success path when GitHub is healthy.
- No change to argument parsing, repo resolution, or health-check semantics.
- Minor non-blocking note: callers that previously treated any nonzero exit as GitHub degraded will now miss the degraded signal when the queue is empty. The JSON payload still carries `github_health.mode`, and the text path emits `github_unavailable=<mode>`, so monitors should consume payload/output rather than exit code alone.

**Verdict:** No blockers.

PR #7106, opened 2026-05-12
fix(automation): keep publisher handoff-only without gh

an0mium:


an0mium:
## Codex focused dogfood

Exact head: `6f648269d5f5a4be8f77310da0bc826fd694fb52`
Worktree: `/Users/armand/.codex/worktrees/f0fd/aragora`.

Validation:
- `pytest tests/scripts/test_run_codex_automation_publisher.py` -> 4 passed in 0.76s
- `ruff check tests/scripts/test_run_codex_automation_publisher.py` -> clean
- `bash -n scripts/run_codex_automation_publisher.sh` -> clean

Review focus: removing the wrapper-level `command -v gh` hard-fail lets the existing GitHub health path classify missing/unavailable gh and leave automation in handoff-only mode instead of aborting before status cache/handoff protection runs.

an0mium:

**Scope:** `scripts/run_codex_automation_publisher.sh`, `tests/scripts/test_run_codex_automation_publisher.py` at `6f648269d5f5a4be8f77310da0bc826fd694fb52`.

**Findings**

- Removing the early `command -v gh` hard-fail is correct. The script's next steps already route through `scripts/github_cli_health.py` with `|| true` and the `HEALTH_READY != "true"` branch, which logs `GitHub unavailable; leaving automations in handoff-only mode` and exits before `publish_codex_automation_branches.py` / `publish_automation_handoffs.py` run. Missing `gh` will produce empty/non-ready health JSON and take exactly that path, so it surfaces as unavailable rather than as a fake publish success.
- Safe handoff-only mode is preserved: branch publish + handoff publish passes are skipped when health is not ready, so queued outbox / receipt directories are untouched and the work remains pending for the next pass. No silent drops.
- `cache_codex_automation_github_status.py` is invoked before the readiness gate but its failure path is already tolerated with `continuing`, so a missing `gh` there is non-fatal and consistent with the new handoff-only intent.
- The new test asserts string presence/absence in the script (`command -v gh`, `gh CLI not found`, handoff-only log line). It is a reasonable regression guard for the intended behavior shift and matches the diff exactly. `bash -n` clean plus 4/4 pytest plus ruff clean is sufficient validation for this shell-only change.
- Behavioral note, not a blocker: the exit code on missing/unhealthy `gh` flips from 1 to 0. Any external monitor that previously alerted on non-zero exit from this wrapper will go quiet; observability now relies on the `GitHub unavailable; leaving automations in handoff-only mode` log line and the status cache. This matches the PR's stated intent.

No blockers.

PR #7110, opened 2026-05-12
fix(automation): require matching PR head receipts

an0mium:


an0mium:
## Codex focused dogfood

Exact head reviewed: `20e3135b582d8b529c355f2a294be93aaad64231`.

Scope checked: `scripts/publish_automation_handoffs.py` and `tests/scripts/test_publish_automation_handoffs.py`. The change requires terminal receipts to match the target PR head before treating an outbox handoff as satisfied, preserving active handoffs when the cached/receipt head is stale.

Validation run from the exact-head worktree `/Users/armand/.codex/worktrees/f1d8/aragora`:

- `/Users/armand/Development/aragora/.venv/bin/python -m pytest tests/scripts/test_publish_automation_handoffs.py` -> 46 passed in 2.99s
- `python3 -m ruff check scripts/publish_automation_handoffs.py tests/scripts/test_publish_automation_handoffs.py` -> clean

Adversarial checks: stale or mismatched receipt heads no longer suppress publication; matching receipts still allow the satisfied path. I found no merge blocker in this PR.

an0mium:

**Scope.** Read-only review of PR #7110 at head `20e3135b582d8b529c355f2a294be93aaad64231`, patch only. No files modified, no commands executed.

**Invariant check - terminal receipts must match target PR head before suppressing publication**

- `Handoff` gains an optional `desired_head` populated from the outbox payload via `_outbox_desired_head`, which accepts `head_sha` / `head` / `commit` and validates a 7-40 char hex shape. This is the receipt-side terminal SHA carrier.
- `gh pr view` / `gh pr list` / `gh issue list` queries are extended with `headRefOid`, so the actual PR head is available everywhere a match is gated.
- `_pr_head_satisfies_handoff` is the single gate:
  - returns `True` when `source_kind != "outbox"` or `desired_head is None` -> existing satisfied behavior preserved for non-outbox / legacy receipts.
  - otherwise requires a non-empty `headRefOid` and a `_head_matches` agreement.
- `_head_matches` is symmetric prefix comparison on lowercased, stripped values with a >=7 char floor, so both short-SHA receipts and full-SHA gh responses, and vice versa, resolve correctly without false positives on <7 char fragments.
- All three suppression paths are gated:
  - `_target_open_pr` branch-PR path (`branch_pr and _pr_head_satisfies_handoff(...)`).
  - `_target_open_pr` referenced-PR-numbers path (state OPEN and head match).
  - `decide_handoffs` `referenced_pr` and `existing_pr` short-circuits both require `_pr_head_satisfies_handoff(...)`.
- Net effect: when a receipt declares a desired head and the live PR head has advanced past it, the PR is no longer treated as the target, publication proceeds (`eligible=True`), and no suppression occurs. This matches the new invariant.

**Preservation of matching/satisfied behavior**

- Receipts without `desired_head`, for example legacy or memory-source handoffs with `source_kind != "outbox"`, short-circuit `_pr_head_satisfies_handoff` to `True`, leaving the prior dedupe/suppression logic intact.
- The updated `test_decide_handoffs_routes_branch_handoff_to_open_pr_before_issue_cap` confirms the matching case (`desired_head="abc1234"` vs `headRefOid="abc1234deadbeef"`) still yields a `pr_open` suppression, so satisfied receipts continue to suppress.
- The new `test_decide_handoffs_keeps_branch_update_actionable_when_pr_head_is_stale` confirms the stale-head case keeps the handoff `eligible=True`.

**Other observations, non-blocking**

- `_head_matches` floor of 7 chars means a receipt that ever supplies a 4-6 char fragment is silently treated as non-matching; given `_outbox_desired_head` already enforces >=7 hex chars at ingestion, this is internally consistent.
- If a branch-PR lookup returns a stale head, `_target_open_pr` correctly falls through to referenced-PR numbers, also head-gated, before giving up. No suppression leak observed.
- Validation evidence cited by Codex, 46 passed and ruff clean, matches the scope of the change.

**Verdict:** No blockers. The patch enforces the head-match requirement uniformly across branch-PR, referenced-PR-number, and title-search suppression paths, and preserves prior satisfied behavior for receipts that match or do not carry a `desired_head`.

PR #7111, opened 2026-05-12
fix(audit): require matching receipt heads

an0mium:


an0mium:
## Codex focused dogfood

Exact head reviewed: `206c25033def97d0624c5732cb79aeed20c77366`.

Scope checked: `scripts/audit_codex_branch_backlog.py` and `tests/scripts/test_audit_codex_branch_backlog.py`. The change requires receipt PR-head matching before classifying a branch as protected/satisfied, so stale receipts no longer mask still-active branch work.

Validation run from the exact-head worktree `/Users/armand/.codex/worktrees/ec35/aragora`:

- `/Users/armand/Development/aragora/.venv/bin/python -m pytest tests/scripts/test_audit_codex_branch_backlog.py` -> 46 passed in 2.50s
- `python3 -m ruff check scripts/audit_codex_branch_backlog.py tests/scripts/test_audit_codex_branch_backlog.py` -> clean

Adversarial checks: mismatched receipt heads are rejected, matching heads keep the existing protected/satisfied behavior, and the change is limited to backlog-audit classification. I found no merge blocker in this PR.

an0mium:

**Scope.** Read-only review of PR #7111 at head `206c25033def97d0624c5732cb79aeed20c77366`, patch only. No files modified, no commands executed.

**Verdict: No blockers.** The patch preserves the stated invariant in the expected/common cases, and the new test exercises the mismatch path directly.

**Why the invariant holds**

- `terminal_receipted_handoff_branch_heads` now produces `dict[branch -> set[head | None]]` instead of a flat branch set. Heads are collected from every reference site: `local_evidence` entries, the top-level payload, `requested_action`, terminal receipts, and archived outbox payloads. Each site contributes its own head via `_outbox_evidence_head`, or `None` when no head is available.
- The new `has_matching_terminal_receipt(row)` gate replaces the unconditional `handoff_receipted_branches.update(...)`. Classification as receipted now requires either:
  - `terminal_key_matches_branch_head(key, branch, head_sha)`, the existing key-embedded SHA path, unchanged.
  - A branch entry whose recorded head is `None` or prefix-matches the row's `head_sha` via `_commit_prefix_matches`.
- Pure mismatch case: when every recorded head for a branch is an explicit SHA differing from the row's `head_sha`, both clauses return False, so the row falls out of `handoff_receipted_branches` and is no longer protected. This is exactly what the new test `test_audit_ignores_terminal_receipt_with_mismatched_explicit_head` asserts (`handoff_receipt_exists is False`, category `salvage_recent_unique`, summary `handoff_receipted_branches == 0`, backlog stays 1).
- Matching-head case: `_commit_prefix_matches` is symmetric (`actual.startswith(expected) or expected.startswith(actual)`) and case-insensitive after strip, so short-to-long and long-to-short SHA forms still match, preserving previously-receipted classification.
- Backwards compatibility: legacy receipts with no head evidence, where the set contains `None`, still satisfy the gate, matching prior behavior for older payload shapes. The fallback `refs.setdefault(branch, {None})` only fires when no explicit head was recorded for that branch.

**Minor observations, non-blocking**

- Leniency on mixed evidence: if a single payload references the same branch from multiple sites and at least one site lacks any head field, `None` will be present in the set and the row will be classified as receipted regardless of any mismatched explicit head from another site within the same payload. This is consistent with the "be lenient only when no evidence" intent, but is worth keeping in mind if future payload writers begin emitting partial head metadata.
- `_outbox_evidence_head` checks keys in order `head_sha`, `head`, `commit`, `sha`; this matches the field names used in the test fixture and in the local-evidence shape elsewhere in the file.
- `_commit_prefix_matches` requires both sides non-empty before the prefix check, preventing an empty `expected` from matching everything.
- Ruff clean and the full `tests/scripts/test_audit_codex_branch_backlog.py` suite, 46 tests, passing at this head, including the new mismatch test, corroborates the change.

No blockers.

PR #7113, opened 2026-05-12
fix(automation): preserve stale PR-head handoffs

an0mium:


an0mium:
## Codex focused dogfood

Exact head reviewed: `40c27ae683c7667a4e6cda10077575abc55f26a8`.

Scope checked: `scripts/reconcile_automation_outbox.py` and `tests/scripts/test_reconcile_automation_outbox.py`. The change preserves/refreshes handoffs when terminal receipt evidence points at a stale PR head instead of treating the handoff as satisfied.

Validation run from the exact-head worktree `/Users/armand/.codex/worktrees/36ad/aragora`:

- `/Users/armand/Development/aragora/.venv/bin/python -m pytest tests/scripts/test_reconcile_automation_outbox.py` -> 14 passed in 0.90s
- `python3 -m ruff check scripts/reconcile_automation_outbox.py tests/scripts/test_reconcile_automation_outbox.py` -> clean

Adversarial checks: stale receipt head evidence remains actionable; matching/current receipt evidence preserves the terminal/satisfied path. I found no merge blocker in this PR.

an0mium:

Scope: `scripts/reconcile_automation_outbox.py` and `tests/scripts/test_reconcile_automation_outbox.py` at head `40c27ae683c7667a4e6cda10077575abc55f26a8`. Read-only inspection; Codex evidence, 14 passed and ruff clean on both touched files, accepted as-is.

**Correctness against stated focus**

- **Stale target-PR handoffs preserved.** The new `_receipt_handoff_keep_reason` only triggers on `status=already_satisfied` + `reason=target_open_pr`, which is the exact class of receipts that previously caused premature archival. When `origin/<branch>` does not match `desired_head` and the local branch ref does match, the handoff is short-circuited with `decision=keep`, the file is not archived, and a new `blocked_receipt_pr_head_mismatch` counter is incremented while `still_protecting_active_work` is bumped. The test exercises both remote-stale + local-fresh and asserts the outbox file still exists and the archive directory is never created.
- **Already-satisfied receipts still respected when remote matches.** Early return on `_heads_match(desired_head, remote_head)` preserves the prior fast-path: if `origin/<branch>` already points at the desired head, we fall through to `satisfied_by_existing_receipt`.
- **No silent evidence drop.** The new branch only ever adds a `keep` action; it never deletes, archives, or rewrites the outbox payload, and it does not synthesize a receipt (`synthetic_receipt: False`). Existing archive/delete paths are untouched.
- **Conservative fallbacks.** Empty `desired_head`, empty remote ref, or empty local ref each short-circuit safely; `_heads_match` enforces a 7-char minimum so empty/zero-length strings cannot spuriously match. Other terminal statuses (`published`, `completed`, `skipped`) intentionally bypass the new check, appropriate since their semantics do not depend on a live PR head.

**Observations, non-blocking**

- `_heads_match` uses a mutual-prefix match, either side may be the prefix. With a 7-char minimum this is fine in practice, but if `desired_head` is ever stored as a 7-char short SHA the abbreviation collision risk is non-zero. Current callers pass full 40-char SHAs, so no action needed here.
- The protection only fires when the local branch ref proves the desired head exists locally. If neither remote nor local refs match `desired_head`, for example after a worktree is pruned, the handoff still gets archived under `satisfied_by_existing_receipt`. That matches the brief's explicit local proof requirement.
- `run_git(["rev-parse", "--verify", ref], ...)` is invoked per receipt match; for very large outboxes this is O(n) git calls but well within reconciler budgets and matches existing patterns in the file.

**Blockers:** No blockers.

PR #7116, opened 2026-05-12
fix(automation): flag stale target-pr cache receipts

an0mium:


an0mium:
## Codex focused dogfood

Exact head reviewed: `ff45405de29345829f4e524eec20d211908e41a3`.

Scope checked: `scripts/cache_codex_automation_github_status.py` and `tests/scripts/test_cache_codex_automation_github_status.py`. The change records stale target-PR cache receipt/head mismatches so downstream automation does not treat outdated receipts as current proof.

Validation run from the exact-head worktree `/Users/armand/.codex/worktrees/b6de/aragora`:

- `/Users/armand/Development/aragora/.venv/bin/python -m pytest tests/scripts/test_cache_codex_automation_github_status.py` -> 11 passed in 0.89s
- `python3 -m ruff check scripts/cache_codex_automation_github_status.py tests/scripts/test_cache_codex_automation_github_status.py` -> clean

Adversarial checks: stale target-PR head evidence is surfaced rather than silently trusted; existing current-head cache behavior remains covered. I found no merge blocker in this PR.

an0mium:

**Verdict: No blockers.** The patch correctly flags `target_open_pr` cache receipts as stale when the cached/remote head diverges from the desired handoff head, and preserves prior satisfied-cache accounting for every other shape.

**Behavior verification**

- `_stale_target_pr_receipt_evidence` short-circuits to `None` unless the receipt's `reason == "target_open_pr"` and the outbox handoff actually requests a `target_pr`, either at top level or inside `requested_action`. Non-target-PR receipts therefore keep the prior key-match-to-receipted semantics, so this is a strict tightening, not a regression.
- Staleness is only asserted when there is a concrete `desired_head_sha`, or fallback alias, and a counter-head from either the receipt payload or `git rev-parse origin/<branch>^{commit}`. If neither head source is available, the receipt is still counted as satisfied: the fail-open is intentional and matches cannot-prove-stale means trust receipt.
- Per-outbox loop uses the `for/else` idiom correctly: any non-stale receipt for the same idempotency key breaks out and increments `terminal_receipted_outbox_count`, only entering the `else` branch, recording stale evidence, when every matching receipt is stale. A coexisting fresh receipt suppresses the stale flag rather than racing it, which matches intent.
- Corrupt/unreadable outbox payloads fall back to counting the entry as receipted (`terminal_receipted_outbox_count += 1; continue`), preserving the original counting behavior for malformed JSON instead of silently bumping the unreceipted total.

**Cache-suppression risk check**

- Stale entries are excluded from `terminal_receipted_outbox_count` and surface in both `stale_target_pr_receipted_outbox_count` and a capped `stale_target_pr_receipted_outbox` list. They also flow into `unreceipted_outbox_count` via the existing outbox-minus-terminal arithmetic, so publish/repair candidates are not suppressed by a stale `already_satisfied` receipt, which was the core bug.
- The `_remote_tracking_head` helper uses `--end-of-options`, `--verify --quiet`, `check=False`, and a 10s timeout; failures return `""` and skip remote comparison rather than raising, so a flaky git or detached remote will not crash the cache pass.

**Minor observations, non-blocking**

- The new `stale_target_pr_receipted_outbox` list is truncated to `DUPLICATE_OUTBOX_EXAMPLE_LIMIT` but the count field reports the full list length before slicing at emission. That is correct here, but a future refactor that moves the slice earlier would silently undercount; worth keeping in mind.
- Tests cover the `remote_tracking_head_mismatch` path and the matching-remote-head path. The `receipt_head_mismatch` branch, where `published_head_sha`/`headRefOid` on the receipt itself differs from `desired_head_sha`, is implemented but not directly asserted in tests. Coverage gap, not a correctness blocker.
- `_remote_tracking_head` is invoked per outbox entry; for large outboxes a memoization keyed by branch would be cheap, but current volume makes this immaterial.

No correctness, safety, or regression blockers identified at `ff45405de29345829f4e524eec20d211908e41a3`.

PR #7513, opened 2026-05-28
fix(sdk): guard websocket reconnect without event loop

an0mium:
## Summary
- Guard SDK WebSocket reconnect scheduling when no event loop is running in the current thread.
- Keep disconnected-event emission non-throwing in synchronous disconnect paths.
- Add a regression test for disconnect handling with no main-thread event loop.

## Validation
- uv run --python 3.11 --with pytest --with pytest-asyncio --with httpx --with pydantic --with websockets python -m pytest sdk/python/tests/test_streaming.py::TestAragoraWebSocketReconnection::test_handle_disconnect_without_event_loop_emits_disconnected -q
- uv run --python 3.11 --with pytest --with pytest-asyncio --with httpx --with pydantic --with websockets python -m pytest sdk/python/tests/test_streaming.py -q
- uv run --python 3.11 --with pytest --with pytest-asyncio --with httpx --with pydantic --with websockets python -m pytest sdk/python/tests -q
- python3 -m pytest sdk/python/tests/test_streaming.py -q
- python3 -m ruff check sdk/python/aragora_sdk/websocket.py sdk/python/tests/test_streaming.py

## Notes
This is a bounded repair lane for the SDK failure observed while diagnosing #7465. It does not mutate #7465, does not rerun #7465 checks, and does not merge/admin-merge anything.

### DIFF
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -13,11 +13,11 @@
 | Metric | Value | Source | Command |
 |---|---|---|---|
 | Python files under aragora/ | `4149` | `aragora/` | `git ls-files aragora \| grep -E '\.py$' \| wc -l` |
-| Python lines of code under aragora/ | `1945694` | `aragora/` | `python3 -c "from pathlib import Path; import subprocess; files = subprocess.check_output(['git', 'ls-files', 'aragora'], text=True).splitlines(); print(sum(sum(1 for _ in Path(p).open(encoding='utf-8', errors='replace')) for p in files if p.endswith('.py')))"` |
+| Python lines of code under aragora/ | `1945866` | `aragora/` | `python3 -c "from pathlib import Path; import subprocess; files = subprocess.check_output(['git', 'ls-files', 'aragora'], text=True).splitlines(); print(sum(sum(1 for _ in Path(p).open(encoding='utf-8', errors='replace')) for p in files if p.endswith('.py')))"` |
 | Top-level modules under aragora/ | `141` | `aragora/` | `git ls-files aragora \| awk -F/ 'NF>2 {print $2}' \| sort -u \| wc -l` |
 | Test files (test_*.py under tests/) | `5238` | `tests/` | `git ls-files tests \| grep -E '(^\|/)test_[^/]*\.py$' \| wc -l` |
-| Test functions (class + module level) | `219188` | `tests/` | `git grep -E '^[[:space:]]*(async )?def test_' -- tests \| wc -l` |
-| @pytest.mark.parametrize decorators | `705` | `tests/` | `git grep -E '@pytest\.mark\.parametrize' -- tests \| wc -l` |
+| Test functions (class + module level) | `219200` | `tests/` | `git grep -E '^[[:space:]]*(async )?def test_' -- tests \| wc -l` |
+| @pytest.mark.parametrize decorators | `708` | `tests/` | `git grep -E '@pytest\.mark\.parametrize' -- tests \| wc -l` |
 | CLI top-level command modules | `71` | `aragora/cli/commands/` | `git ls-files aragora/cli/commands \| grep -E '/[^/]*\.py$' \| grep -v '/__' \| wc -l` |
 | OpenAPI paths | `2870` | `docs/api/openapi.json` | `python -c "import json; print(len(json.load(open('docs/api/openapi.json'))['paths']))"` |
 | OpenAPI operations (HTTP verbs) | `3297` | `docs/api/openapi.json` | `python -c "import json; spec=json.load(open('docs/api/openapi.json')); print(sum(1 for p in spec['paths'].values() for m in p if m.lower() in {'get','post','put','delete','patch','head','options'}))"` |
@@ -28,9 +28,9 @@
 | Allowlisted agent types | `35` | `aragora/config/settings.py` | `grep -A 50 'ALLOWED_AGENT_TYPES' aragora/config/settings.py \| grep -oE "'[a-z-]+'" \| sort -u \| wc -l` |
 | Knowledge Mound adapter specs | `41` | `aragora/knowledge/mound/adapters/factory.py` | `git grep -E '"\.[a-z_]+_adapter"' -- aragora/knowledge/mound/adapters/factory.py \| wc -l` |
 | Knowledge Mound adapter files | `46` | `aragora/knowledge/mound/adapters/` | `git ls-files aragora/knowledge/mound/adapters \| grep -E '/[^/]+_adapter\.py$' \| wc -l` |
-| Markdown files under docs/ | `966` | `docs/` | `git ls-files docs \| grep -E '\.md$' \| wc -l` |
+| Markdown files under docs/ | `964` | `docs/` | `git ls-files docs \| grep -E '\.md$' \| wc -l` |
 | GitHub Actions workflows | `86` | `.github/workflows/` | `git ls-files .github/workflows \| grep -E '\.yml$' \| wc -l` |
-| Mypy baseline errors (grandfathered) | `3317` | `.mypy-baseline` | `wc -l .mypy-baseline` |
+| Mypy baseline errors (grandfathered) | `3115` | `.mypy-baseline` | `wc -l .mypy-baseline` |
 
 ## Notes on counting methodology
 
--- a/sdk/python/aragora_sdk/websocket.py
+++ b/sdk/python/aragora_sdk/websocket.py
@@ -381,6 +381,13 @@ class AragoraWebSocket:
 
     def _schedule_reconnect(self) -> None:
         """Schedule a reconnection with exponential back-off."""
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            logger.debug("Skipping reconnect scheduling because no event loop is running")
+            self._state = _STATE_DISCONNECTED
+            return
+
         self._state = _STATE_RECONNECTING
         delay = self.options.reconnect_delay * (2**self._reconnect_attempts)
         self._reconnect_attempts += 1
@@ -392,7 +399,7 @@ class AragoraWebSocket:
             except Exception:
                 pass  # connect() will emit errors and trigger disconnect
 
-        self._reconnect_task = asyncio.ensure_future(_do_reconnect())
+        self._reconnect_task = loop.create_task(_do_reconnect())
 
     def _cleanup(self) -> None:
         """Cancel background tasks."""
--- a/sdk/python/tests/test_streaming.py
+++ b/sdk/python/tests/test_streaming.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import json
 from unittest.mock import AsyncMock, MagicMock, patch
 
@@ -539,6 +540,27 @@ class TestAragoraWebSocketReconnection:
         event_data = handler.call_args[0][0]
         assert event_data["code"] == 1006
         assert event_data["reason"] == "Connection lost"
+
+    def test_handle_disconnect_without_event_loop_emits_disconnected(self) -> None:
+        """_handle_disconnect does not require a main-thread event loop."""
+        ws = AragoraWebSocket("https://api.aragora.ai")
+        handler = MagicMock()
+        ws.on("disconnected", handler)
+
+        try:
+            previous_loop = asyncio.get_event_loop()
+        except RuntimeError:
+            previous_loop = None
+
+        asyncio.set_event_loop(None)
+        try:
+            ws._handle_disconnect(1006, "Connection lost")
+        finally:
+            asyncio.set_event_loop(previous_loop)
+
+        handler.assert_called_once()
+        assert ws.state == "disconnected"
+        assert ws._reconnect_task is None
 
     def test_handle_disconnect_schedules_reconnect(self) -> None:
         """_handle_disconnect schedules reconnection if auto_reconnect enabled."""


an0mium:
## Claude frontier adversarial review

Exact SHA: 6531ebad2968ae9e2888f08ba237473c41eb0e21
Verdict: APPROVE

Validation reviewed: Non-quorum required checks (sdk-parity, Generate & Validate, lint, typecheck, TypeScript SDK Type Check) are green on this head; aragora-merge-quorum is CANCELLED and merge-packet flags Tier 3 semantic/API risk requiring human settlement, so this is a content review only, not a merge endorsement. Diff is narrowly scoped to `sdk/python/aragora_sdk/websocket.py:_schedule_reconnect` and a single new sync test in `sdk/python/tests/test_streaming.py::test_handle_disconnect_without_event_loop_emits_disconnected`. The substantive code change replaces `asyncio.ensure_future(_do_reconnect())` with an early `asyncio.get_running_loop()` probe (RuntimeError → log at DEBUG, pin state to `_STATE_DISCONNECTED`, return) followed by `loop.create_task(_do_reconnect())` on the successful path. No other call site, public API, or serialization contract is altered.

Adversarial findings:
- (Non-blocker, design note) Silent reconnect suppression is logged at DEBUG only and emits no `error` / `warning` SDK event. In normal operation `_handle_disconnect` is invoked from `_receive_loop`, so a running loop is guaranteed and this branch should be unreachable; if it is ever reached in production it indicates a real lifecycle bug (callback fired after loop teardown, sync caller, cross-thread misuse) that DEBUG logging will likely hide. Consider escalating to `logger.warning(...)` or emitting a synthetic `"warning"` event so downstream observers can detect it, but this is not required to land the crash fix.
- (Non-blocker, consistency) `_send` at the unchanged site still uses `asyncio.ensure_future(self._ws.send(...))` and carries the same latent no-running-loop failure mode. It is gated by `_state == _STATE_CONNECTED` which normally implies a live loop, but the symmetry argument used to justify hardening `_schedule_reconnect` applies equally here. Worth a follow-up; not in-scope for this PR.
- (Non-blocker, counter accounting) On the no-loop early return, `_reconnect_attempts` is intentionally not incremented, which is the right choice (no attempt was made), but it means a caller that manually re-enters `connect()` after fixing their loop still has full retry budget — confirm that matches expected semantics.
- (Non-blocker, test hygiene) The new test saves `previous_loop` via `asyncio.get_event_loop()` inside `try/except RuntimeError`. On Python 3.12+, `asyncio.get_event_loop()` with no current loop emits a `DeprecationWarning` (and in 3.14 raises) before any exception path; under `-W error` this could flip the captured `previous_loop` semantics. The restore via `asyncio.set_event_loop(previous_loop)` correctly handles `None`, so global state is not polluted across tests in the common case; recommend switching to `asyncio.get_event_loop_policy().get_event_loop()` or guarding with `warnings.catch_warnings()` for forward compatibility.
- (Non-blocker, coverage) Test asserts `_reconnect_task is None` and state transition, but does not assert the DEBUG log line or that `_reconnect_attempts` was *not* incremented — both are part of the contract the diff introduces and would be cheap to lock in.
- No blockers identified. The fix is strictly safer than the prior `asyncio.ensure_future` path, which would raise `RuntimeError: no running event loop` (Python 3.10+) or trigger the `DeprecationWarning`-then-implicit-loop behavior (≤3.11) instead of cleanly degrading. `asyncio.get_running_loop()` + `loop.create_task` has been the canonical pattern since 3.7 and carries no policy/version compatibility risk on supported runtimes.

Focused product dogfood: This is the Python SDK websocket reconnect path that the streaming/dock product surface (and live/spectate consumers via `stream_debate`) depend on. The prior code could crash an embedder that drives the websocket from a sync entry point (CLI tools, test harnesses, framework callbacks invoked after loop shutdown) instead of staying in `disconnected` and letting the host orchestrate retry. After this change, embedders get deterministic degradation: state lands on `disconnected`, `_reconnect_task` stays `None`, no exception escapes — matching what the trust-wedge/inbox and demo-receipt flows need when their async runtime is not the one that called into the SDK.

Residual risk: Silent reconnect skip is invisible to operators relying on SDK logs at INFO+; recommend a follow-up to surface a `"warning"` event and/or raise the log level, harmonize `_send` against the same pattern, and tighten the new test under `-W error` on 3.12+. Tier 3 human-risk settlement is still the right gate before any merge.

an0mium:
# Claude Opus 4.7 Review
PR: #7513
Exact head: feb93ff99913c37384974b88f099d64688781aab
Verdict: APPROVE

Validation summary:
- Reviewed inline diff (3 files: `docs/METRICS.md`, `sdk/python/aragora_sdk/websocket.py`, `sdk/python/tests/test_streaming.py`)
- Confirmed `pytest-sdk-streaming.txt` reports 53 passed (includes the new `test_handle_disconnect_without_event_loop_emits_disconnected`)
- Confirmed `regenerate-metrics-check.txt` reports no drift beyond 0.5%
- Confirmed `automation-pr-preflight.txt` reports `preflight: ok`
- Cross-checked fix semantics against `_handle_disconnect → _schedule_reconnect` call chain implied by the sibling `test_handle_disconnect_schedules_reconnect` test

Receipt/artifact path: /Users/armand/Development/aragora/.aragora/proof/pr-7513/20260528T183446Z-feb93ff999

Focused dogfood / adversarial review of the no-event-loop reconnect path:

**Fix correctness.** `_schedule_reconnect` now wraps `asyncio.get_running_loop()` in `try/except RuntimeError`. This is the right primitive: `get_running_loop()` (unlike the deprecated `get_event_loop()`) raises `RuntimeError` whenever no loop is currently running on the calling thread, which is exactly the failure mode the previous `asyncio.ensure_future(...)` call hit (it would have implicitly attempted to acquire a loop and exploded). On the exception path the method sets `self._state = _STATE_DISCONNECTED` and returns before touching `_reconnect_attempts` or scheduling a task, so the object is left in a coherent terminal state instead of half-mutated `_STATE_RECONNECTING`. Switching the happy-path scheduler to `loop.create_task(_do_reconnect())` (binding the task to the captured loop) is a strict improvement over `asyncio.ensure_future` and removes the implicit-loop dependency entirely.

**Test coverage of that exact path.** `test_handle_disconnect_without_event_loop_emits_disconnected` constructs a real `AragoraWebSocket`, calls `asyncio.set_event_loop(None)`, and invokes `_handle_disconnect(1006, "Connection lost")` from synchronous test code — there is no running loop in either dimension, so `get_running_loop()` will raise. The assertions check the three things that matter: (1) the `disconnected` handler fires (proving `_handle_disconnect` ran to the emit step), (2) `ws.state == "disconnected"` (proving the new early-return branch set `_STATE_DISCONNECTED` rather than leaving `_STATE_RECONNECTING`), and (3) `ws._reconnect_task is None` (proving no task was scheduled and we didn't fall through to `loop.create_task`). The previous event loop is restored in `finally`, so test isolation is preserved.

**Adversarial probes considered.**
- Does the fix regress the happy path? No — `test_handle_disconnect_schedules_reconnect` and the rest of `TestAragoraWebSocketReconnection` continue to exercise the with-loop case, and the full file passes (53/53).
- Does the early return leak partial state? No — `_state` is the only mutation, and it's set to a terminal value; `_reconnect_attempts` is not incremented, which is correct because no attempt was scheduled.
- Could `loop.create_task` reject the coroutine? No — `_do_reconnect()` is a coroutine, which is exactly what `create_task` requires.
- Is `logger.debug` content sensitive? No — message is static and contains no user data.

**METRICS.md.** Numeric drift (test count +12, parametrize +3, markdown -2, mypy baseline -202) is consistent with a metrics refresh that cumulates other recently-merged work, not just this PR's single new test. `regenerate_metrics.py --check` passing confirms the snapshot is in sync with current tree state, so the doc churn is legitimate.

No blocking concerns. Diff is minimal, surgical, and the new test directly exercises the regression path.

an0mium:
# Factory Droid Review
PR: #7513
Exact head: feb93ff99913c37384974b88f099d64688781aab
Verdict: APPROVE
Validation summary: Factory/Droid GLM-5.1 reviewed the inline exact-head diff and validation transcript preserved in this proof directory.
Receipt/artifact path: /Users/armand/Development/aragora/.aragora/proof/pr-7513/20260528T183446Z-feb93ff999
Focused dogfood/adversarial dogfood: see the unmodified Factory/Droid output below; it explicitly reviews the no-event-loop SDK websocket reconnect path and the focused test coverage.
Independent model review: Factory/Droid GLM-5.1.

Unmodified Factory/Droid output SHA-256:
ef64931d89207b186d0f58bdc6b3e6006184e8414161c2d8b7cd8d07f62f8e64  /Users/armand/Development/aragora/.aragora/proof/pr-7513/20260528T183446Z-feb93ff999/factory-droid-review-inline.md

--- BEGIN UNMODIFIED FACTORY/DROID OUTPUT ---
# Adversarial Review

PR: #7513
Exact head: feb93ff99913c37384974b88f099d64688781aab
Verdict: **APPROVE**

## Validation summary

**Artifacts reviewed:** `diff.patch`, `diffstat.txt`, `required-checks.json`, `sdk/python/aragora_sdk/websocket.py` (full file, 418 lines), `sdk/python/tests/test_streaming.py` (full file, 53 tests), `pytest-sdk-streaming.txt` (53 passed), `regenerate-metrics-check.txt` (no drift), `automation-pr-preflight.txt` (exit 0).

**Commands verified (from CI artifacts):**
- `python3 scripts/regenerate_metrics.py --check` → exit 0
- `python3 -m pytest sdk/python/tests/test_streaming.py` → 53 passed, 2 warnings
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` → exit 0
- Required non-quorum checks: lint ✓, typecheck ✓, SDK parity ✓, OpenAPI spec ✓, TS SDK type check ✓

---

## Focused dogfood / adversarial analysis

### The fix is correct and minimal

The bug: `_schedule_reconnect()` called `asyncio.ensure_future(_do_reconnect())` without first checking whether an asyncio event loop is running. If `_handle_disconnect` was called from a context without a running loop (e.g. a non-async thread, or after loop closure), `ensure_future` would raise `RuntimeError: no running event loop`.

The fix adds a two-part guard:

1. **`asyncio.get_running_loop()` probe** — raises `RuntimeError` when no loop is executing in the current OS thread. The `except RuntimeError` branch logs at debug level, sets `_state = _STATE_DISCONNECTED`, and returns early. No reconnect task is created.

2. **`asyncio.ensure_future()` → `loop.create_task()`** — since we now have an explicit `loop` reference from `get_running_loop()`, using `loop.create_task()` is both more explicit and recommended. `_do_reconnect()` is always a coroutine, so `create_task` is the correct API (vs. `ensure_future` which also accepts futures).

### State transition correctness

Call chain: `_handle_disconnect` → sets `_state = _STATE_DISCONNECTED` → calls `_schedule_reconnect()`.

- **With loop:** `_schedule_reconnect` sets `_state = _STATE_RECONNECTING`, schedules the task. Correct.
- **Without loop:** `_schedule_reconnect` catches RuntimeError, sets `_state = _STATE_DISCONNECTED` (idempotent with what `_handle_disconnect` already set), returns. `_reconnect_task` stays `None`. Correct.

The `_reconnect_attempts` counter is NOT incremented in the no-loop path — correct because no reconnect was actually scheduled.

### `_schedule_reconnect` has exactly one caller

Grep confirms `_schedule_reconnect` is called only from `_handle_disconnect` (line 380). No other code path reaches it. The guard covers all entry points.

### Test coverage adequacy

**New test `test_handle_disconnect_without_event_loop_emits_disconnected`:**
- Sets `asyncio.set_event_loop(None)` to simulate a no-loop environment (necessary for Python <3.12 where `get_event_loop()` would auto-create a loop; the `set_event_loop(None)` ensures the old `ensure_future` path would also fail on those versions)
- Calls `_handle_disconnect(1006, "Connection lost")` synchronously (no running loop)
- Asserts: disconnected handler called ✓, state is `"disconnected"` ✓, `_reconnect_task is None` ✓
- Saves and restores the previous event loop in a `try/finally` ✓

**The test exercises the complete guard path** — `_handle_disconnect` → `_schedule_reconnect` → `RuntimeError` catch → early return. This is the real-world scenario where the bug would manifest.

### Minor observations (non-blocking)

1. **`_send()` still uses `asyncio.ensure_future`** (line 284) without a loop check. However, `_send` is gated by `_state == _STATE_CONNECTED`, which is only set inside `connect()` (an async method running on a loop), so the risk is lower. Not in scope for this PR.

2. **No direct unit test for `_schedule_reconnect` when a loop IS running.** The existing `test_handle_disconnect_schedules_reconnect` mocks `_schedule_reconnect`, so it doesn't verify that `loop.create_task` is actually called. Integration coverage exists via the async `connect` tests. Pre-existing gap, not introduced here.

3. **The `_state = _STATE_DISCONNECTED` in `_schedule_reconnect`'s guard is redundant** — `_handle_disconnect` already set it. But redundancy is defensive and harmless; if `_schedule_reconnect` were ever called from another context, the state would still be correct.

### Metrics change

`docs/METRICS.md` drift is within 0.5% threshold (verified by `regenerate_metrics.py --check` exit 0). Line count, test count, parametrize count, markdown count, and mypy baseline updates are all consistent with a small PR adding ~22 test lines and 9 source lines.

---

**Verdict: APPROVE.** The fix is a minimal, correct guard that prevents a `RuntimeError` when `_schedule_reconnect` is called without a running event loop. The `ensure_future → loop.create_task` substitution is safe and idiomatic. The test exercises the actual bug path end-to-end and verifies all relevant postconditions. All CI checks pass on exact head `feb93ff999`.

Receipt/artifact path: `/Users/armand/Development/aragora/.aragora/proof/pr-7513/20260528T183446Z-feb93ff999`
--- END UNMODIFIED FACTORY/DROID OUTPUT ---

an0mium:
# Claude Sonnet 4.6 Current-Head Dogfood Review

**PR:** #7513 | **Head:** `be02b6284debfb594a08402ba7859532a8d21606`
**Scope:** `sdk/python/aragora_sdk/websocket.py`, `sdk/python/tests/test_streaming.py`

---

## Verdict: ✅ APPROVE

The fix is correct, well-scoped, and the new test properly validates the target failure path.

---

## Analysis

### What the patch does

`_schedule_reconnect` previously called `asyncio.ensure_future(_do_reconnect())` unconditionally. If the method was ever reached outside a running event loop (e.g. during interpreter shutdown, in a finalizer, or from a thread that owns no loop), Python would raise a `RuntimeError` and leave the client in `_STATE_RECONNECTING` permanently — a stuck state with no recovery path.

The fix:
1. Calls `asyncio.get_running_loop()` (correctly prefers the "strictly running" semantics over the deprecated `get_event_loop()`).
2. On `RuntimeError`, sets state to `_STATE_DISCONNECTED` and returns early — clean, predictable recovery.
3. Replaces `asyncio.ensure_future()` with the captured `loop.create_task()` — more explicit and avoids a second implicit loop lookup.

### State-machine correctness

With `auto_reconnect=True` (the default), `_handle_disconnect` → `_schedule_reconnect`. The guard ensures the state never gets stranded in `_STATE_RECONNECTING` when no loop is available; it falls back to `_STATE_DISCONNECTED`. The test confirms the "disconnected" event fires and `_reconnect_task` stays `None` as expected for a fresh instance.

### Test quality

`test_handle_disconnect_without_event_loop_emits_disconnected` correctly uses `asyncio.set_event_loop(None)` to remove the current-thread loop, which causes `get_running_loop()` to raise `RuntimeError` — the exact guard being tested. Save-and-restore in a `finally` block avoids contaminating subsequent test state. The three assertions (event fired, final state, no reconnect task) cover the observable contract.

---

## Non-Blocking Risks

**1. Silent reconnect suppression in worker-thread callers**
If `_schedule_reconnect` is invoked from a secondary thread that has no running event loop while the *main* event loop is healthy (e.g. a background finalizer or a `concurrent.futures` callback), the guard fires the debug log and drops the reconnect attempt rather than scheduling it onto the main loop via `loop.call_soon_threadsafe`. For an SDK used in multi-threaded contexts this is a latent footgun — callers see a silent `"disconnected"` with no retry. A follow-up could capture the loop reference at construction time and fall back to `loop.call_soon_threadsafe` when the calling thread has no running loop.

**2. `asyncio.get_event_loop()` in test teardown**
The save/restore idiom calls `asyncio.get_event_loop()` outside any running coroutine, which emits a `DeprecationWarning` on Python 3.10+ and raises `RuntimeError` on 3.12+ in some thread contexts. The `try/except RuntimeError` already handles the raise path correctly; the warning is cosmetic but could produce noise in strict-warning CI configurations.

**3. Implicit `_reconnect_task is None` precondition**
The test relies on `_reconnect_task` being `None` at construction without asserting it explicitly. This is fine for a fresh `AragoraWebSocket("https://api.aragora.ai")` instance, but a future change that pre-initialises `_reconnect_task` (e.g. a mock sentinel) would silently invalidate the assertion.

---

All three risks are follow-up concerns; none block this change. The core fix is sound.

an0mium:
# Factory Droid Current-Head Review

**PR:** #7513
**Head:** `be02b6284debfb594a08402ba7859532a8d21606`
**Verdict:** ✅ APPROVE
**Independent model review: Factory Droid GLM-5.1.**

---

### What the diff does

`_schedule_reconnect` now guards against being called outside a running event loop — it catches `RuntimeError` from `asyncio.get_running_loop()`, logs at debug, sets state to `DISCONNECTED`, and returns early instead of crashing on `asyncio.ensure_future`. The happy-path also switches to `loop.create_task()` (preferred over `ensure_future` since 3.7). A new test `test_handle_disconnect_without_event_loop_emits_disconnected` clears the loop, calls `_handle_disconnect`, and asserts the disconnected event fires with no reconnect task.

### Adversarial dogfood — SDK websocket reconnect without an event loop

1. **Silent reconnect loss is correct but could surprise callers.** When `_handle_disconnect` is invoked from a plain thread (no loop), the WebSocket transitions to `disconnected` and **no reconnect will ever be attempted**. This is the only safe behavior (there's no loop to schedule on), but an SDK consumer who registered an `auto_reconnect=True` WS from a background thread and later wonders why it never reconnects gets only a `logger.debug` breadcrumb. Consider bumping to `logger.warning` in a follow-up so production log noise surfaces the gap — **non-blocker**.

2. **`loop.create_task` vs `ensure_future`** — cleaner and avoids the implicit-loop fetch that caused the original crash. Good.

3. **Test uses `asyncio.set_event_loop(None)`** — deprecated since Python 3.12. Will emit `DeprecationWarning` on newer interpreters. Not a blocker for this PR, but worth tracking for when the CI matrix bumps past 3.12.

4. **Edge case: loop stops *after* `get_running_loop()` succeeds but before `create_task`** — this would raise `RuntimeError` inside `create_task`, which is caught by the existing `except Exception: pass` in `_do_reconnect` and the disconnect callback fires again. The cycle is self-limiting (max reconnect attempts). Acceptable.

### Validation awareness

Required non-quorum checks are **green** on `be02b6284debfb594a08402ba7859532a8d21606`. The `aragora-merge-quorum` check is cancelled solely because current-head model quorum evidence is missing. Local focused validation previously passed:

- `python3 -m pytest sdk/python/tests/test_streaming.py` ⇒ **53 passed**
- `python3 scripts/regenerate_metrics.py --check` ⇒ **no drift**

### Summary

Small, well-scoped fix for a real crash path. No blockers. Two non-blocking follow-ups noted (log level, deprecation).

an0mium:
Tier-3 Human Settlement Authorization: I accept the semantic/API/SDK risk for PR #7513 at exact head be02b6284debfb594a08402ba7859532a8d21606.

PR #7600, opened 2026-06-01
fix(quorum-gate): stop the merge-quorum self-cancel race + robust dogfood recognition

an0mium:
## Root cause (verified)

Two independent defects make the required **`aragora-merge-quorum`** check flap red and stall otherwise-ready PRs:

1. **Self-cancel race.** `.github/workflows/aragora-merge-quorum.yml` set `concurrency.cancel-in-progress: true` with group `aragora-merge-quorum-<pr-number>`. When two triggers fire for the same PR (e.g. `synchronize` + `reopened`, or two rapid pushes), the newer run cancels the older one. A **cancelled** run reports a `cancelled` conclusion, which GitHub surfaces as a **red required check** — blocking the PR until a human manually reruns and happens to win a clean green window with no overlapping trigger.

2. **Brittle dogfood recognition.** `aragora/cli/commands/review_queue.py :: _dogfood_evidence_from_comments` inferred the reviewer model **only from the comment's first markdown heading**. A dogfood comment headed `## Focused adversarial dogfood` (without `(claude)` in the heading) failed to count as dogfood evidence even though its body named the model — e.g. a `Model family: claude` line. This stalled **PR #7587** until a corrected comment was posted by hand.

### 1. Workflow — `cancel-in-progress: false`
Flip the concurrency cancel behavior so concurrent runs run to completion instead of cancelling each other into a red status. This matches the repo's existing precedent of setting `cancel-in-progress: false` on its other non-required workflows to "stop the doom loop." **The quorum's pass/fail criteria are unchanged** — only the concurrency cancel behavior. GitHub still evaluates the latest completed run per head SHA, so the gate is not weakened.

### 2. Dogfood recognition — full-body model inference
New helpers `_model_family_from_body` and `_resolve_dogfood_identity`. When the first heading names no model, the recognizer falls back to a **full-body `Model family:` scan** and, if a known/normalizable family is found, synthesizes a countable identity. This mirrors how the model-review-signal and review-object recognizers already trust disclosed model-family metadata over the heading, keeping the recognizers consistent.

**All existing safety filters are preserved unchanged:**
- head-grounding (`_is_comment_grounded_on_head`) — stale comments still excluded;
- github-actions author exclusion;
- dogfood-token requirement (`dogfood` / `adversarial` / `cross-author` / `recheck`);
- fail-closed on unknown or unnormalizable model families (no phantom inflation).

## Tests (TDD)
Added to `tests/cli/commands/test_review_queue.py`:
[...]

PR #7612, opened 2026-06-01
fix(quorum-gate): issue_comment auto-re-eval NOT viable — reduced to no-op [follow-up to #7600]

an0mium:
## STATUS: superseded by review — issue_comment auto-re-eval is NOT viable. Reduced to a no-op. DRAFT.

This PR originally added an `issue_comment` trigger to auto-re-evaluate the `aragora-merge-quorum` required gate after evidence is posted. **An independent Codex review found a correctness-breaking defect, which I verified against GitHub's documented semantics. The mechanism does not work and cannot be made to work cleanly. The workflow is now reverted to exactly `origin/main` (a no-op).**

## Why the issue_comment approach is broken (verified)

**1. issue_comment runs are bound to the default-branch SHA, not the PR head.**
A workflow triggered by `issue_comment` executes against the repository's **default branch** — `GITHUB_SHA` is `main`'s HEAD. The check-run GitHub Actions creates for that run therefore attaches to **main's** check-suite, **not the PR head**. The PR's required `aragora-merge-quorum` check would **not** update on the PR head SHA, so the advertised auto-re-evaluation simply does not happen. (Confirmed: issue_comment workflows run on the default branch; checks attach to the default-branch commit. See GitHub Actions docs / `actions/checkout#331`.)

**2. Option A — POST a commit status named `aragora-merge-quorum` on the PR head — is NOT viable.**
I verified the required context against live branch protection and the existing check object:

- Branch protection requires `aragora-merge-quorum` pinned to **`app_id: 15368` (GitHub Actions)**.
- The existing `aragora-merge-quorum` is a **check-run** (`app.slug = github-actions`), produced by the workflow job — **not** a commit-status context. (Verified via `repos/.../commits/<sha>/check-runs`.)

Per GitHub's official docs (*Troubleshooting required status checks*):

> "If you have a check and a status with the same name, and you select that name as a required status check, **both the check and the status are required**."

So if the issue_comment path POSTed a commit **status** named `aragora-merge-quorum` on the PR head, branch protection would then require **BOTH** the check-run AND the commit-status for that one context. On any new push — a head SHA that has the workflow's check-run but no comment-posted status yet — the PR would hang **"Expected — Waiting for status"** indefinitely, blocking merge until someone happens to post a comment. That is a **worse, more fragile gate** than the bug it was meant to fix. Posting the status under a *different* name would not update the required context at all.

There is no clean way to make an issue_comment-triggered run update the PR-head required check-run without exactly this check-vs-status conflict. Per the merge-authority principle (correctness and non-fragility over shipping), I did **not** ship it.

## What this PR now contains

The workflow is reverted to **exactly `origin/main`** — `git diff origin/main` for `.github/workflows/aragora-merge-quorum.yml` is **empty**. The branch is a no-op against the gate.

**#7600 (already merged) solves the dominant problem** — the self-cancel race — via `cancel-in-progress: false`. With that in place, concurrent runs no longer cancel each other into a red required check, so an operator no longer has to "win a green window." The marginal value of the issue_comment re-trigger was always small, and it is not worth a fragile gate change.

**This PR is close-worthy.** It can be closed without loss; the only durable artifact is the design recommendation below.

## Recommended proper design (next step — separate Tier-4 PR)

A head-SHA-bound, **signed quorum-settlement RECEIPT** that becomes *countable evidence*, surfaced through the **same existing `aragora-merge-quorum` check-run** (never a second status/context):

- When `2/2 model signals + head-grounded dogfood + required-green` is reached at an exact head SHA, an authorized signer emits a signed receipt bound to that SHA (analogous to the existing `aragora/human-settlement` commit-status pattern for Tier 3–4).
- The receipt is read by `review-queue merge-packet` as additional head-grounded countable evidence — it does not bypass any filter and is invalidated automatically when the head SHA changes (a new push drops it).
- Crucially, the receipt must be **explicitly invalidated when underlying evidence changes** (e.g. a dogfood comment edited/deleted at the same SHA), so it never trusts a stale conclusion over live evidence. This is precisely why a naive "latch to a prior green conclusion" is unsafe.
- Re-evaluation still happens only through the legitimate `pull_request` (`synchronize`) trigger and the normal evaluation path, which already report against the PR head correctly.

This deserves its own Tier-4 review and is out of scope here.

## Tier-4 / merge-authority self-modification — DRAFT ONLY

`.github/workflows/aragora-merge-quorum.yml` is a **Tier-4** surface under `docs/REVIEW_AUTHORITY_PRINCIPLES.md`. This PR remains a **DRAFT** and is now a no-op; recommend **close** in favour of the receipt design above.

### DIFF
--- a/.github/workflows/aragora-merge-quorum.yml
+++ b/.github/workflows/aragora-merge-quorum.yml
@@ -18,19 +18,6 @@ name: Aragora Merge Quorum
 on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
-  # Posting or editing a PR comment (e.g. a model-review signal or a dogfood
-  # report) auto-re-evaluates the quorum for that PR. This removes the need to
-  # manually rerun the check or wait for a push/synchronize event after posting
-  # evidence — the gate latches to the current head SHA automatically.
-  #
-  # SECURITY: an issue_comment ONLY triggers a RE-EVALUATION. It does not grant
-  # quorum. The evaluation (review-queue merge-packet) still reads only
-  # countable, head-grounded, non-github-actions evidence via the GitHub API
-  # against the PR's CURRENT head SHA. A comment cannot fake a quorum: the
-  # head-grounding, author-exclusion, dogfood-token, and fail-closed model-family
-  # filters are all unchanged. See docs/REVIEW_AUTHORITY_PRINCIPLES.md.
-  issue_comment:
-    types: [created, edited]
   workflow_dispatch:
     inputs:
       pr_number:
@@ -39,11 +26,7 @@ on:
         type: string
 
 concurrency:
-  # Per-PR group, resolving the PR number from whichever trigger fired
-  # (pull_request -> .pull_request.number, issue_comment -> .issue.number,
-  # workflow_dispatch -> .inputs.pr_number). The comment-triggered run shares
-  # the per-PR group with push/synchronize runs.
-  group: aragora-merge-quorum-${{ github.event.pull_request.number || github.event.issue.number || github.event.inputs.pr_number }}
+  group: aragora-merge-quorum-${{ github.event.pull_request.number || github.event.inputs.pr_number }}
   # cancel-in-progress is intentionally false. With true, two triggers for the
   # same PR (e.g. synchronize + reopened, or rapid pushes) cancel each other;
   # a cancelled run reports a "cancelled" conclusion, which turns this REQUIRED
@@ -52,8 +35,7 @@ concurrency:
   # non-required workflows to "stop the doom loop"; we do the same here so
   # concurrent runs run to completion instead of cancelling into a red status.
   # GitHub still evaluates the latest completed run per head SHA, so the gate is
-  # not weakened — only the self-cancel race is removed. A comment-triggered run
-  # must likewise never abort an in-flight push run for the same PR.
+  # not weakened — only the self-cancel race is removed.
   cancel-in-progress: false
 
 permissions:
@@ -66,21 +48,8 @@ jobs:
     name: aragora-merge-quorum
     runs-on: ubuntu-latest
     timeout-minutes: 15
-    # issue_comment events fire for BOTH issues and PRs. Only run when the
-    # comment is on a pull request (github.event.issue.pull_request is present);
-    # for pull_request / workflow_dispatch triggers there is no issue payload, so
-    # the guard short-circuits to true and the job runs as before.
-    if: >-
-      github.event_name != 'issue_comment' ||
-      github.event.issue.pull_request != null
 
     steps:
-      # For pull_request / workflow_dispatch this checks out the triggering ref;
-      # for issue_comment it checks out the default branch. The checkout is used
-      # ONLY to install the aragora CLI — the quorum is evaluated against the
-      # PR's CURRENT head SHA via `gh pr view` (live API lookup) in the
-      # merge-packet command below, never against the checked-out tree. So the
-      # comment-triggered run still reports status against the PR's current head.
       - uses: actions/checkout@v4
 
       - name: Set up Python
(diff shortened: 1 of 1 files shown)


an0mium (commented):
## Aragora review-pr: advisory findings

The issue_comment trigger does not actually update the PR head check, so the advertised auto re-evaluation/latch behavior will not work.

- Review mode: `advisory comment only`
- Final status: `changes_requested`
- Reviewer: `codex`
- Reviewed at: `2026-06-01T21:03:51.590056+00:00`
- Head SHA: `bd954f5b2a257c3ff8c2aa727b01d2439ede854c`
- Settlement note: machine review is advisory only; it does not approve or block merge on its own.
- Review route: `codex`

### Findings
- [P1] issue_comment runs are attached to the default-branch SHA, not the PR head SHA (.github/workflows/aragora-merge-quorum.yml): For issue_comment events, GitHub Actions creates the workflow check on the default branch commit from the event, not on the pull request head. This diff adds the trigger and resolves PR_NUMBER, but it does not checkout the PR head or create/update a check/status on the PR head SHA. As a result, posting evidence may run the quorum evaluation, but the required `aragora-merge-quorum` check on the PR head will remain stale, so the intended latch-to-green behavior will not occur. The workflow needs to explicitly report the result against the PR current head SHA, for example by resolving the head SHA and creating/updating a commit status/check run for that SHA or by otherwise ensuring the required check is associated with the PR head.

an0mium (commented):
## Aragora review-pr: advisory pass

Adds an issue_comment (created/edited) trigger to the merge-quorum workflow so posting evidence auto-re-evaluates the gate, plus PR-number/draft resolution across trigger types. The change is well-reasoned and safe: it only triggers RE-EVALUATION (never grants quorum), checks out the default branch (not PR code), interpolates no untrusted comment content into shell, and preserves draft-deferral and fail-closed behavior. A couple of low-priority observations only.

- Review mode: `advisory comment only`
- Final status: `passed`
- Reviewer: `claude`
- Reviewed at: `2026-06-01T21:04:28.233956+00:00`
- Head SHA: `bd954f5b2a257c3ff8c2aa727b01d2439ede854c`
- Settlement note: machine review is advisory only; it does not approve or block merge on its own.
- Review route: `claude:max-02`

### Findings
- [P3] Non-draft pull_request runs now incur an extra live `gh pr view` API call (.github/workflows/aragora-merge-quorum.yml): The IS_DRAFT default fallback changed from `|| 'false'` to `|| ''`. Because GitHub Actions short-circuits `github.event_name == 'pull_request' && github.event.pull_request.draft` to `false` for a NON-draft PR, the `|| ''` branch makes IS_DRAFT empty for every non-draft pull_request event, forcing the new live `gh pr view --json isDraft` lookup. Previously this path got `'false'` directly with no API call. Behavior is still correct (lookup returns false → evaluate; lookup failure → warn + false → evaluate), so this is purely a minor per-run latency/rate-limit cost, not a correctness regression. Worth confirming it's intentional given the gate already runs on every push/synchronize.
- [P3] issue_comment broadens who can trigger CI runs (.github/workflows/aragora-merge-quorum.yml): `issue_comment: [created, edited]` lets anyone who can comment on a PR trigger a quorum re-evaluation, increasing Actions-minute consumption (a mild resource/DoS vector on public repos). The design correctly bounds the security impact: the comment body is never interpolated into any script, PR_NUMBER derives from the integer `issue.number`, the checkout uses the default branch (no PR code execution), and quorum evidence is still head-grounded and author/dogfood/model-family filtered. Concurrency group sharing with cancel-in-progress:false also prevents comment runs from aborting in-flight push runs. No code change required; flagging the trigger-surface broadening for awareness. If this is a public repo, consider gating on commenter association or a comment marker to limit churn.

an0mium:
Closing as a no-op. Verified the issue_comment auto-re-evaluation approach is not viable (issue_comment runs attach checks to the default-branch SHA, not the PR head; and a same-named commit-status would force branch protection to require BOTH the check-run AND the status → permanent 'Expected — Waiting for status' hang). #7600 (merged) already fixes the dominant self-cancel race via cancel-in-progress:false. The proper follow-up — a head-SHA-bound signed quorum-settlement receipt surfaced through the existing check-run, with invalidation on evidence change — is documented here for a future Tier-4 design PR.

PR #7750, opened 2026-06-04
feat(review-queue): collect-evidence — genuine model-review evidence collector (B3)

an0mium:
## B3 — `review-queue collect-evidence` (genuine model-review evidence collector)

Phase 2 of the boss-loop merge-gate resilience design
(`docs/governance/BOSS_LOOP_MERGE_GATE_RESILIENCE.md`). Closes the loop opened by
B1 (#7742, re-runs the gate when evidence lands) and B2 (#7740, diagnose the
self-check) by **producing** the genuine evidence in the first place.

### What
A collector that:
1. Resolves a PR's **exact current head** and tier.
2. Runs **>=2 genuine, heterogeneous** model reviewers against the head diff.
3. Composes each reviewer's output into an evidence comment whose heading the
   canonical quorum parsers recognize.
4. Validates **every** comment with the same `review-queue evidence-lint` parser
   the gate uses — *before* anything is posted.

Surfaces: new `review-queue collect-evidence` subcommand + thin
`scripts/collect_quorum_evidence.py`, both delegating to the new
`aragora/swarm/quorum_evidence.py` module.

### Safety invariants (enforced in the module, not the caller)
- **Never fabricate** — a comment is only ever composed from a reviewer that
  actually returned non-empty output; failed/empty reviewers are recorded and
  produce no comment.
- **Tier-gated posting** — only Tier 0-2 may auto-post (and only with `--apply`);
  **Tier 3-4 (and unknown tier) always prepare** evidence for an operator and
  never post. `decide_action` fail-closes to `prepare` when tier is unknown.
- Default is a **dry run** (prepares + lints, posts nothing).

### Bug fix bundled
`merge_quorum_io.fetch_pr_tier` previously read `tier` off the merge-packet
top-level object and so always returned `None` for the real `--json` shape
(tier lives under `entries[]`). Fixed to read the envelope correctly; this also
benefits `scripts/settle_status.py` (A2), which calls the same helper.

### Tests (30 new)
- `compose_evidence_comment` output is validated against the **real**
  `_lint_evidence_comment` parser (would_count=True for claude + grok), so the
  collector stays bound to the gate's logic.
- Tier-gating: Tier 3+ never posts even with `--apply`; unknown tier fail-safes.
- Never-fabricate: a failed reviewer yields no comment.
- `fetch_pr_tier` across envelope/list/single/None/bad-json/timeout shapes.
- ruff clean; live dry-run smoke confirmed end-to-end (head resolve → genuine
[...]

PR #7765, opened 2026-06-05
fix(scripts): attribute Codex Desktop sessions in agent_bridge_sessions

an0mium:
## Problem

`scripts/agent_bridge_sessions.py` is documented (module docstring + `--help`) as listing **"Codex/Claude sessions"**, but `collect_sessions` only read **tmux + claude** transcripts — there was no codex reader (`--source` choices were just `all/tmux/claude`). 

Because a Codex Desktop automation run spawns `claude -p` reviewers, the only transcripts this tool could see were those `~/.claude/projects` children, which it tags `source="claude_jsonl"`. **Net effect: Codex Desktop activity (e.g. the #7750 Tier-4 settlement) was mislabeled as a claude session** (or omitted). The consolidated `list_active_agent_sessions.py` attributes correctly; this narrower tool did not, so the two disagreed.

## Fix

- Add a **codex source** to `agent_bridge_sessions.py`: reads `~/.codex/sessions/**/*.jsonl` rollouts, classifies **`codex_desktop` vs `codex_cli`** via the `session_meta` payload's `originator` field (`"Codex Desktop"`), agent=`codex`. Bounded scan (`--codex-scan-limit`, default 500) + tail-bounded summary so it stays cheap on large rollouts.
- New flags: `--source codex`, `--codex-home`, `--codex-scan-limit`.
- Capture `originator`/`thread_source` in `list_active_agent_sessions.read_codex_session_metadata` so both tools agree on attribution (additive; existing exact-dict tests unaffected).

## Verification

Live against this repo (formerly mislabeled): `--source codex` now yields **6 `codex_desktop` + 2 `codex_cli`, all `agent=codex`** (none as `claude_jsonl`), with correct branch/cwd/summary.

- 61 unit tests pass (4 new: desktop-not-claude, cli, repo-filter, absent-home; 2 isolated for the new param)
- ruff, mypy, portability guard: all green; pre-commit hooks passed

## Context

Surfaced while auditing Codex Desktop session activity on #7750 — the operator's settlement was driven by Codex Desktop engineering-autopilot (`originator="Codex Desktop"`), not claude.

### DIFF
--- a/scripts/agent_bridge_sessions.py
+++ b/scripts/agent_bridge_sessions.py
@@ -136,14 +136,23 @@ def _safe_repo_root(candidate: str | None) -> Path | None:
 
 
 def _tail_lines(path: Path, *, max_bytes: int = TAIL_BYTES) -> list[str]:
-    if not path.exists():
+    # Seek to the tail instead of slurping the whole file: Codex rollouts can be
+    # hundreds of MB, and only the trailing ``max_bytes`` are ever needed.
+    try:
+        with path.open("rb") as handle:
+            handle.seek(0, os.SEEK_END)
+            size = handle.tell()
+            if size > max_bytes:
+                handle.seek(size - max_bytes)
+                data = handle.read()
+                newline = data.find(b"\n")
+                if newline >= 0:
+                    data = data[newline + 1 :]
+            else:
+                handle.seek(0)
+                data = handle.read()
+    except OSError:
         return []
-    data = path.read_bytes()
-    if len(data) > max_bytes:
-        data = data[-max_bytes:]
-        newline = data.find(b"\n")
-        if newline >= 0:
-            data = data[newline + 1 :]
     return data.decode("utf-8", errors="replace").splitlines()
 
 
@@ -641,8 +650,10 @@ def _extract_codex_summary(path: Path) -> str:
     """Best-effort last visible message text from a Codex rollout tail.
 
     Bounded to the trailing ``TAIL_BYTES`` so it stays cheap even on multi-hundred
-    megabyte rollouts. Tool-call/token-count noise lines are skipped.
+    megabyte rollouts. Prefers the most recent assistant message, falling back to
+    the most recent visible text; tool-call/token-count noise lines are skipped.
     """
+    fallback = ""
     for line in reversed(_tail_lines(path)):
         if not line.strip():
             continue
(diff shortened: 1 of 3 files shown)


an0mium:

Reviewer: claude (anthropic) — independent adversarial model review via claude CLI (claude -p), adversarial re-review, grounded on the exact PR head.
Head: c9a777a (c9a777a2e553bbd9d92510357bd360efac6ae5e8), committed 2026-06-04T21:08:23-05:00.
PR: #7765.
Model family: claude

VERDICT: PASS

Walked the diff against each prior finding.

**1. P2 perf (memory) — resolved.** `_tail_lines` now does `open("rb") → seek(0, SEEK_END) → seek(size − max_bytes)` and reads at most `max_bytes`. Peak memory bounded regardless of rollout size. Edge cases check out: empty file (`size=0` → else branch → reads 0 bytes → `splitlines() == []`), file ≤ max_bytes (else branch, full read), file just over (seek+slice past first partial line). `OSError` catch subsumes the old `path.exists()` guard.

**2. P3 perf (stat fan-out) — resolved.** `load_codex_sessions` does a name-sort (no stat) → window of `scan_limit*4` (=2000) → mtime-sort only the window. Stat work is now O(limit), not O(total files). `_mtime` returning `-1.0` on OSError plus the second `path.stat()` in the per-record loop being wrapped in try/except means a single bad file no longer drops the whole listing.

**3. Classification containment — correct, not over-matching.** Lowercase + `in` covers "Codex Desktop", "Codex Desktop 1.2.3", "codex desktop"; rejects "Codex CLI". Only realistic over-match would be an originator that literally embeds "codex desktop" in its name, which would in fact be a desktop variant. Test `test_collect_sessions_codex_desktop_versioned_originator` pins this.

**4. Assistant-preferred summary — correct.** Walks the *reversed* tail; returns on first `role == "assistant"`, retains first non-assistant text as `fallback`. So the result is "most recent assistant, else most recent visible text" — exactly the stated intent.

**No new bugs introduced.** Reviewed potential regressions:
- The name-sort prefilter relies on the comment's claim that Codex filenames embed an ISO timestamp; that matches observed Codex rollout naming (`rollout-<iso>-<uuid>.jsonl`). If a future Codex version changes naming, the window could miss the truly-newest files — but mtime-sort within the window plus a 4× buffer makes that benign for any reasonable shift, and re-introducing full-tree stat would defeat the fix. Acceptable.
- `rglob("*.jsonl")` still lists the whole tree, but listing without stat is cheap on macOS/Linux (`scandir` + `d_type`).
- Seek/empty-file edges in `_tail_lines` are safe (verified above).
- Containment compares lowercased originator after `.strip()`; safe against whitespace and `None`.

Findings: none.

dogfood: yes

an0mium:

Reviewer: grok (xai) — independent adversarial model review via aragora xAI grok agent (independent model), grounded on the exact PR head.
Head: c9a777a (c9a777a2e553bbd9d92510357bd360efac6ae5e8), committed 2026-06-04T21:08:23-05:00.
PR: #7765.
Model family: grok

VERDICT: passed
- none

dogfood: yes

PR #7814, opened 2026-06-06
fix: update NEXT_STEPS_CANONICAL.md

an0mium:
(cherry picked from commit 3a77c1420d90aedfbc9b487c4bfd1c3c73a05c1c)

PR #7821, opened 2026-06-06
Fix quorum rerun after evidence collection

an0mium:
## Summary
- Trigger the same-PR merge-quorum reconciler after `review-queue collect-evidence --apply` successfully posts low-tier evidence.
- Keep Tier 3/Tier 4 behavior prepare-only; the default CLI reconciler is only wired in apply mode and only after successful posts.
- Preserve A1 safety by reusing the reconciler cooldown, max-rerun, and real-failure guards instead of adding a new rerun path.

## Context
- Live dogfood on #7814 showed the seam: collect-evidence posted countable Claude/Grok evidence, but stale `aragora-merge-quorum` still needed a separate manual rerun.
- This change closes that B3/A1 handoff gap without broad-draining or changing merge authority.

## Tests
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest -q tests/swarm/test_quorum_evidence.py::test_collect_low_tier_apply_triggers_same_pr_quorum_reconciler_after_posts`
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest -q tests/swarm/test_quorum_evidence.py`
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest -q tests/swarm/test_quorum_evidence.py tests/swarm/test_merge_quorum_reconcile.py`
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m ruff check aragora/swarm/quorum_evidence.py tests/swarm/test_quorum_evidence.py`
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m compileall -q aragora/swarm/quorum_evidence.py`
- `git diff --check`

### DIFF
--- a/aragora/swarm/quorum_evidence.py
+++ b/aragora/swarm/quorum_evidence.py
@@ -93,6 +93,7 @@ _MAX_DIFF_CHARS = 60_000
 _MAX_REVIEWER_CHARS = 32_000
 _CLAUDE_TIMEOUT = 300
 _REVIEWER_TIMEOUT = 300
+_REVIEWER_CLEANUP_TIMEOUT = 10
 
 
 def _cap_text(text: str) -> str:
@@ -395,7 +396,9 @@ async def _close_api_agent_resources(agent: Any) -> None:
         try:
             result = close()
             if inspect.isawaitable(result):
-                await result
+                await asyncio.wait_for(result, timeout=_REVIEWER_CLEANUP_TIMEOUT)
+        except TimeoutError:
+            logger.debug("collect-evidence API agent close timed out")
         except Exception as exc:  # noqa: BLE001 - cleanup must not mask reviewer results.
             logger.debug("collect-evidence API agent close failed: %s", exc)
 
@@ -410,7 +413,9 @@ async def _close_api_agent_resources(agent: Any) -> None:
         # loop, so the shared aiohttp connector must be released before that
         # loop is torn down. The collector dispatches reviewers serially; if it
         # ever fans reviewers out, cleanup must move outside the per-reviewer path.
-        await close_shared_connector()
+        await asyncio.wait_for(close_shared_connector(), timeout=_REVIEWER_CLEANUP_TIMEOUT)
+    except TimeoutError:
+        logger.debug("collect-evidence shared connector close timed out")
     except Exception as exc:  # noqa: BLE001 - cleanup must not mask reviewer results.
         logger.debug("collect-evidence shared connector close failed: %s", exc)
 
--- a/aragora/swarm/quorum_evidence.py
+++ b/aragora/swarm/quorum_evidence.py
@@ -29,11 +29,11 @@ from __future__ import annotations
 import asyncio
 import inspect
 import logging
+import multiprocessing
 import os
+import queue
 import re
-import signal
 import subprocess
-import threading
 import time
 from collections.abc import Callable, Iterator, Sequence
 from contextlib import contextmanager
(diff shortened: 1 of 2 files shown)


an0mium:

Head: 3b25012 (3b25012052f85f6948899f886899da30ad4e65ca), committed 2026-06-06T07:15:01Z.
PR: #7821.
Model family: claude

Verdict: PASS

No blocking issues. A few notes worth tracking:

- [P2] `quorum_evidence.py:_run_api_agent` spawn-per-reviewer reimports the whole `aragora` package in each child. Correct, but each API reviewer call now eats a multi-second cold-import cost on top of the actual work; over N families × N PRs this is a real wall-clock regression vs. the prior in-process path. Worth measuring once in CI.
- [P2] `quorum_evidence.py:_locked_quorum_reconcile_state` stale-lock TOCTOU: after `_quorum_state_lock_is_stale` returns true, another process can finish, recreate the lock, and have its fresh lock unlinked here. Bounded by the 15-min staleness threshold, so unlikely in practice, but consider rechecking mtime after unlink or using `os.replace`/fcntl flocks for a tighter guarantee.
- [P2] `quorum_evidence.py:_reviewer_verdict` is strict-prefix on the first non-empty line. A reviewer that prepends a markdown decoration (`**Verdict: PASS**`) or a code fence (```` ``` ````) will get `"unknown"` and be silently demoted from supportive even with valid evidence. Reviewers are instructed to comply, but a single rogue decoration now blocks the post path. Consider stripping leading non-alnum before the prefix test.
- [P2] `quorum_evidence.py` dissent gating: `EvidenceItem.dissenting` ignores `would_count`. A reviewer that emits `Verdict: CHANGES-REQUESTED` with a malformed body still triggers `action="prepare"`. Probably intentional fail-closed, but worth confirming — it means an unparseable dissent has more blocking power than a valid pass.
- [P3] `default_quorum_reconciler` reaches into `scripts.reconcile_merge_quorum._load_state` / `_save_state` (underscored). Easy to break on a refactor of that script; consider promoting those to a public API or pinning with a unit test on the contract.
- [P3] `_quorum_state_lock_is_stale` does a text `pid=…` regex match on the lock body — fine, but if you ever rotate the lock format the staleness check silently becomes "never stale" (regex miss returns `True`, but a corrupted/legacy file matches nothing and the lock is treated as stale, which is the opposite footgun). Either way the format is now load-bearing; a small format-version tag would be cheap insurance.

Reviewer-timeout/cleanup hardening, the supportive-quorum tightening, and the post→reconcile handoff all look correct, and the test diff exercises the new prepare-vs-post branches (dissent, incomplete quorum, reviewer failure, uncountable evidence, reconciler error, locked rerun cap) thoroughly.

dogfood: yes

an0mium:

Head: 3b25012 (3b25012052f85f6948899f886899da30ad4e65ca), committed 2026-06-06T07:15:01Z.
PR: #7821.
Model family: grok

Verdict: PASS

- No blocking issues found. Logic for supportive quorum, multiprocessing isolation (spawn+daemon+timeouts), file locking (O_EXCL+O_NOFOLLOW+stale PID/mtime), early returns on dissent, and error paths all appear correct with no races, injection, or regression risks vs. prior `would_count` behavior. Tests cover new paths.

dogfood: yes

scarmani:
Queue-drain close: this non-draft quorum-rerun substrate PR has been inactive since 2026-06-06, is currently dirty/conflicting against main, and has no active owner or steering lane. The current drain posture forbids more quorum/evidence/reconcile substrate repair on stale branches, so I am closing it rather than patching or rerunning CI.

Closing without deleting the branch. If this behavior is still needed, recut from current main as a fresh bounded PR with current validation and exact-head review evidence.

PR #7926, opened 2026-06-07
fix(similarity): validate prompt dedupe thresholds

an0mium:
## Summary
- fail closed on invalid prompt dedupe similarity thresholds in the prompt extraction/ranking scripts
- reject negative, boolean, non-numeric, and >=1.0 thresholds before Jaccard dedupe can collapse or disable results
- add focused regression coverage across all three prompt dedupe surfaces

## Validation
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest tests/scripts/test_prompt_dedupe_thresholds.py -q`
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m ruff check scripts/extract_and_rank_prompts.py scripts/extract_intent_prompts.py scripts/extract_user_prompts.py tests/scripts/test_prompt_dedupe_thresholds.py`
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m ruff format --check scripts/extract_and_rank_prompts.py scripts/extract_intent_prompts.py scripts/extract_user_prompts.py tests/scripts/test_prompt_dedupe_thresholds.py`

### DIFF
--- a/scripts/extract_and_rank_prompts.py
+++ b/scripts/extract_and_rank_prompts.py
@@ -203,6 +203,15 @@ def extract_user_voice(text: str) -> str:
 
 def deduplicate(prompts: list[dict], threshold: float = 0.65) -> list[dict]:
     """Remove near-duplicate prompts using word overlap."""
+    if (
+        isinstance(threshold, bool)
+        or not isinstance(threshold, int | float)
+        or not 0.0 <= threshold < 1.0
+    ):
+        raise ValueError(
+            "deduplication similarity threshold must be between 0.0 inclusive and 1.0 exclusive"
+        )
+
     seen = []
     unique = []
     for p in prompts:
--- a/scripts/extract_intent_prompts.py
+++ b/scripts/extract_intent_prompts.py
@@ -313,6 +313,15 @@ def extract_user_text(message: dict) -> str | None:
 
 def deduplicate(prompts: list[dict], threshold: float = 0.7) -> list[dict]:
     """Remove near-duplicate prompts."""
+    if (
+        isinstance(threshold, bool)
+        or not isinstance(threshold, int | float)
+        or not 0.0 <= threshold < 1.0
+    ):
+        raise ValueError(
+            "deduplication similarity threshold must be between 0.0 inclusive and 1.0 exclusive"
+        )
+
     seen = []
     unique = []
     for p in prompts:
--- a/scripts/extract_user_prompts.py
+++ b/scripts/extract_user_prompts.py
@@ -173,6 +173,15 @@ def extract_prompts(
 
 def deduplicate(prompts: list[dict], similarity_threshold: float = 0.8) -> list[dict]:
     """Remove near-duplicate prompts (same text repeated across sessions)."""
+    if (
+        isinstance(similarity_threshold, bool)
+        or not isinstance(similarity_threshold, int | float)
+        or not 0.0 <= similarity_threshold < 1.0
+    ):
+        raise ValueError(
+            "deduplication similarity threshold must be between 0.0 inclusive and 1.0 exclusive"
+        )
+
     seen_texts = []
     unique = []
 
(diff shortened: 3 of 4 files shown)


an0mium:
**Head:** 40fd0a6d26adbec1f4b3d4ca3f74adc170bc5282  
**Committed:** 2026-06-07T03:58:51Z  
**PR:** #7926  
**Model family:** Grok  
**Files reviewed:** scripts/extract_and_rank_prompts.py, scripts/extract_intent_prompts.py, scripts/extract_user_prompts.py, tests/scripts/test_prompt_dedupe_thresholds.py  

NO FINDINGS  
The supplied diff adds identical fail-closed guards (bool/non-numeric/negative/>=1.0 rejection) to all three `deduplicate` surfaces with matching error text, plus regression tests that exercise every invalid case plus a boundary-valid run. Boundary handling and coverage are complete for the stated scope.

Validation run/not-run notes: none executed by this reviewer.  
dogfood: yes  

an0mium:

- PR: #7926
- Head SHA: 40fd0a6d26adbec1f4b3d4ca3f74adc170bc5282
- Committed: 2026-06-07T03:58:51Z
- Model family: Claude
- Dogfood: yes

### Files reviewed
- `scripts/extract_and_rank_prompts.py`
- `scripts/extract_intent_prompts.py`
- `scripts/extract_user_prompts.py`
- `tests/scripts/test_prompt_dedupe_thresholds.py`

### Verdict: NO FINDINGS

The three `deduplicate` surfaces now fail closed on out-of-range or wrong-typed similarity thresholds before Jaccard overlap can produce degenerate results (everything collapsed at `>=1.0`, or partial collapse at negatives via Python's numeric comparison). The guard is implemented identically across all three scripts, which is the right choice for this dedupe family.

Boundary handling looks correct:
- `bool` is rejected explicitly with `isinstance(threshold, bool)` first — necessary because `bool` is a subclass of `int` and `0.0 <= False < 1.0` would otherwise pass. Catches both `True` and `False`.
- `int | float` typecheck rejects strings and other non-numerics.
- `0.0 <= threshold < 1.0` admits the inclusive low end, excludes the degenerate `1.0` (where all distinct prompts would be considered duplicates only if identical — but more importantly, this matches the documented contract and prevents silent meaning drift if callers pass `1.0` expecting "always dedupe").
- `NaN` is rejected transitively because all comparisons with NaN are False; not explicitly tested but the behavior is correct.
- Error message is consistent across all three sites and accurately describes the inclusive/exclusive bounds.

Test coverage is appropriately scoped:
- Parametrized across all three modules × all three threshold kwarg names (`threshold`, `threshold`, `similarity_threshold`) — catches any future drift in the keyword surface.
- Invalid set covers the four failure classes the validator targets: negative (`-0.01`), upper-bound (`1.0`), bool (`True`), string (`"0.5"`).
- Positive path verifies exact-duplicate collapse with a valid `0.99` threshold, confirming the guard does not regress the happy path.
- Dynamic module loading via `importlib.util.spec_from_file_location` is appropriate for scripts that aren't packaged.

Minor observations (not blockers):
- `False` is not in the invalid parametrization, but the `isinstance(..., bool)` check catches it identically to `True`, and adding it would only marginally increase coverage.
- The exact lower bound (`0.0`) and slightly-above-lower (`0.0`) aren't asserted as valid in tests; only `0.99` is. The guard's `<=` makes this correct by construction, but a single `0.0` positive case would lock in the inclusive boundary contract.
- Three near-identical copies of the guard is a small duplication, but lifting it into a shared helper would couple three independent scripts; keeping them in-line is the lower-risk choice here.

### Validation
Not run — review performed from PR body and diff only. The PR's listed validation (`pytest tests/scripts/test_prompt_dedupe_thresholds.py`, ruff check/format, `git diff --check`, `automation_pr_preflight.sh`, pre-push mypy + portability) was not independently re-executed by this reviewer.

This review is not merge authorization.

scarmani:

PR #7927, opened 2026-06-07
fix(proof): validate quality prose filters

an0mium:
## Summary
- add parser-level validation for quality prose extraction filters
- reject non-positive `--min-words` / `--top-n` and negative `--min-quality` before building a misleading corpus
- make `--user-only` and `--assistant-only` mutually exclusive
- add focused parser regression tests for invalid and valid filter combinations

## Validation
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest tests/scripts/test_extract_quality_prose.py -q`
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m ruff check scripts/extract_quality_prose.py tests/scripts/test_extract_quality_prose.py`
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m ruff format --check scripts/extract_quality_prose.py tests/scripts/test_extract_quality_prose.py`

### DIFF
--- a/scripts/extract_quality_prose.py
+++ b/scripts/extract_quality_prose.py
@@ -416,7 +416,27 @@ def extract_quality_prose(
 # =============================================================================
 
 
-def main():
+def _positive_int(raw: str) -> int:
+    try:
+        value = int(raw)
+    except ValueError as exc:
+        raise argparse.ArgumentTypeError("must be a positive integer") from exc
+    if value <= 0:
+        raise argparse.ArgumentTypeError("must be a positive integer")
+    return value
+
+
+def _non_negative_float(raw: str) -> float:
+    try:
+        value = float(raw)
+    except ValueError as exc:
+        raise argparse.ArgumentTypeError("must be a non-negative number") from exc
+    if value < 0:
+        raise argparse.ArgumentTypeError("must be a non-negative number")
+    return value
+
+
+def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
     parser = argparse.ArgumentParser(
         description="Extract high-quality prose from conversations",
         formatter_class=argparse.RawDescriptionHelpFormatter,
@@ -438,29 +458,30 @@ def main():
     )
     parser.add_argument(
         "--min-quality",
-        type=float,
+        type=_non_negative_float,
         default=0.5,
         help="Minimum quality score (default: 0.5)",
     )
     parser.add_argument(
         "--min-words",
-        type=int,
+        type=_positive_int,
         default=100,
         help="Minimum words per passage (default: 100)",
     )
-    parser.add_argument(
+    role_group = parser.add_mutually_exclusive_group()
+    role_group.add_argument(
         "--user-only",
         action="store_true",
         help="Only extract user messages",
     )
-    parser.add_argument(
+    role_group.add_argument(
         "--assistant-only",
         action="store_true",
         help="Only extract assistant messages",
     )
     parser.add_argument(
         "--top-n",
-        type=int,
+        type=_positive_int,
         help="Only output top N passages by quality",
     )
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


an0mium:
**Head:** c7dd5ce8c2052e6914882d571413314b7ff7deb7  
**Committed:** 2026-06-07T04:07:11Z  
**PR:** #7927  
**Model family:** Grok  
**Files reviewed:** scripts/extract_quality_prose.py, tests/scripts/test_extract_quality_prose.py  

NO FINDINGS  
Parser validation (_positive_int, _non_negative_float, mutually_exclusive_group) correctly implements the stated requirements for --min-words, --top-n, --min-quality, and role flags. New regression tests cover both rejection paths and valid combinations. Empty-input fail-closed behavior is not addressed in the supplied diff or body.  

Validation run/not-run notes: commands listed in PR body were not executed by this review.  
dogfood: yes  

an0mium:

- PR: #7927
- Head SHA: c7dd5ce8c2052e6914882d571413314b7ff7deb7
- Committed: 2026-06-07T04:07:11Z
- Model family: Claude
- Dogfood: yes
- Files reviewed:
  - `scripts/extract_quality_prose.py`
  - `tests/scripts/test_extract_quality_prose.py`

### Verdict: NO FINDINGS

The change moves the quality-prose CLI into a fail-closed posture before any corpus work happens. `_positive_int` rejects both non-integers and any `value <= 0`, so `--min-words 0`, `--min-words -1`, `--top-n 0`, and `--top-n -1` all exit at argparse time with `ArgumentTypeError` rather than silently producing a degenerate / unbounded corpus. `_non_negative_float` similarly bars negative `--min-quality` while still admitting `0.0`, which matches the intended "no quality floor" semantic without sneaking through `-0.1`. Both helpers re-raise via `from exc` so the original `ValueError` is preserved in the chain, and the `ArgumentTypeError` paths return `int`/`float` rather than mutating state, which is correct for argparse `type=` callables.

The mutually exclusive group for `--user-only` / `--assistant-only` is the right primitive — argparse will reject the conflict before the extractor sees an impossible filter, instead of letting one flag silently win. Splitting `parse_args(argv)` out of `main(argv)` is a clean seam for the new tests, and `main` still defaults `argv=None` so the production entry point is unchanged.

Regression coverage is appropriate for the scope:
- `test_parse_args_rejects_invalid_quality_filter_values` parametrizes the five fail-closed cases (`--min-words` 0/-1, `--top-n` 0/-1, `--min-quality` -0.1), all asserted via `pytest.raises(SystemExit)` which is the correct argparse exit signal.
- `test_parse_args_rejects_conflicting_role_filters` locks in the mutually exclusive contract.
- `test_parse_args_accepts_valid_quality_filters` pins the happy path including the boundary `--min-quality 0` → `0.0` and confirms `assistant_only` defaults to `False` when only `--user-only` is set, so a future regression that flipped the group semantics would be caught.

`_load_module()` uses `importlib.util.spec_from_file_location` against the on-disk script path, which is the right approach for a `scripts/` file that isn't on `sys.path`; each test re-execs the module, so there's no risk of cached `parse_args` state across cases.

Scope note (not a blocker): the PR title focus mentions "empty input fail-closed" — this change fails closed on empty/invalid *filter arguments*, not on an empty input conversations file. The downstream behavior of `extract_quality_prose()` when the input JSON yields zero qualifying passages is not exercised here, but that is outside the stated PR scope ("validate quality prose filters") and the existing extraction path is untouched.

### Validation
I did not execute the PR's validation commands — this is a tools-off independent read of the diff at `c7dd5ce8c2052e6914882d571413314b7ff7deb7`. The PR body lists `pytest tests/scripts/test_extract_quality_prose.py`, `ruff check`, `ruff format --check`, `git diff --check`, `automation_pr_preflight.sh`, and pre-push hooks (mypy + portability) as having run; I am relying on those as author-reported, not re-run.

scarmani:

PR #8208, opened 2026-06-11
feat(ops): per-pass GitHub App token routing for daemon gh calls

an0mium:
Tier: 2

## Summary

The Python automation paths already route GitHub calls through the GitHub App installation token (`aragora/swarm/github_app_auth.py`, separate budget). The gap: the shell daemons call `gh` directly with the operator's personal auth, draining the operator's API budget. Installation tokens expire after **one hour**, so the token must be minted **per pass**, never baked into daemon env at start.

- **`scripts/gh_app_env.py` (new)**: mints a fresh installation token. `--print-token` emits the bare token for command substitution; default mode emits an eval-able `GH_TOKEN=` assignment. **Silent-safe contract**: exit 0 and print nothing when App config is absent, the mint fails, or aragora is unimportable — daemons degrade to the operator's existing gh auth instead of crashing. The token is never logged or echoed outside the designated stdout mode; `--quiet` suppresses the stderr diagnostic.
- **`scripts/run_codex_automation_publisher.sh`**: opt-in per-pass mint at the top of the publish pass behind `ARAGORA_GH_APP_AUTH=1` (default OFF, mirrors the `ARAGORA_AUTO_EVIDENCE` / `ARAGORA_QUORUM_RECONCILER` wiring pattern). The script is exactly one pass per launchd invocation, so every pass gets a fresh token.
- **`scripts/run_merge_arbiter.sh`**: same flag, but scoped in a **subshell** to the bounded auto-evidence section only. The arbiter itself is a long-running `exec` that refreshes its own App token per call via `aragora.swarm.github_app_auth` — it must NOT inherit a one-shot token that would go stale after an hour.
- **`aragora/swarm/github_app_auth.py`**: `github_cli_env(prefer_app=False)` now drops a shell-exported App token (tagged via `ARAGORA_GITHUB_AUTH_SOURCE`) so write ops keep preferring the operator PAT over narrow-scope App auth. Previously a shell-exported token would silently leak into the write path.

## Identity note

`gh` CLI prefers `GH_TOKEN` over keyring auth, so this cleanly switches identity per-process: with the flag on, daemon reads are attributed to `aragora-automation-fable[bot]` in audit logs instead of the operator — better attribution and a fully separate rate-limit budget. Operators should verify quorum-evidence counting accepts bot-authored comments before flipping the flag on the arbiter's auto-evidence path.

## Live proof (budget independence)

Minted once via the new script and compared `gh api rate_limit` core buckets (token value never printed):

| auth | limit | remaining | used |
|------|-------|-----------|------|
| operator keyring | 5000 | 4851 | 149 |
| App installation token | 15000 | 15000 | 0 |

Fully independent pools; the App budget is 3x larger.

## Tests

- 10 new: `tests/scripts/test_gh_app_env.py` — config-absent silence, minter-crash + import-failure silence, bare/eval output shapes, no stderr leakage in any mode, whitespace-token-as-absent, subprocess integration with App auth disabled.
- 2 new: `tests/swarm/test_github_app_auth.py` — `prefer_app=False` drops tagged App tokens, keeps untagged user tokens.
- 61/61 passing across `test_github_app_auth.py` + `test_gh_app_env.py` + `test_auto_evidence_cycle.py`; `bash -n` clean on both daemons; ruff + mypy clean.
- Live wiring fragment exercised: mint OK, degradation path empty-no-crash, eval mode OK.

## Governance

Two-family adversarial gate (run-20260610 lane Z): grok **Verdict: MERGE**; mistral-api initially CHANGES with three findings, each declined with documented technical rebuttal (no `/proc`/history on macOS launchd scripts + same-UID trust domain; `/tmp` logging contradicts the silent-safe contract; subshell exports cannot reach the parent), clean re-review returned **Verdict: MERGE**. Receipts VALID (3/3): `.aragora/run-20260610/receipts/laneZ_{grok,mistral-api}.json` (+ `laneZ_mistral-api_dissent.json` preserving the dissent round).

### DIFF
--- a/aragora/swarm/github_app_auth.py
+++ b/aragora/swarm/github_app_auth.py
@@ -251,7 +251,12 @@ def github_cli_env(
 ) -> dict[str, str]:
     env = dict(os.environ if base_env is None else base_env)
     if not prefer_app:
-        return env
+        # Drop any App-sourced token a shell wrapper exported (tagged via
+        # ARAGORA_GITHUB_AUTH_SOURCE by the scripts/gh_app_env.py wiring) so
+        # callers that do not prefer App auth — e.g. write ops where the
+        # installation has narrow scopes — fall back to the operator's
+        # PAT/keyring auth instead of silently using the App token.
+        return _drop_app_token(env)
     token = get_github_app_installation_token(env)
     if not token:
         return env
--- a/scripts/gh_app_env.py
+++ b/scripts/gh_app_env.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python3
+"""Mint a per-pass GitHub App installation token for shell ``gh`` consumers.
+
+Shell daemons (codex-automation publisher, merge-arbiter wrapper) call ``gh``
+directly with the operator's personal auth, draining the operator's API
+budget. The Python paths already route through the App installation token via
+:mod:`aragora.swarm.github_app_auth`; this script extends that to shell:
+
+    tok="$(python3 scripts/gh_app_env.py --print-token 2>/dev/null || true)"
+    [ -n "$tok" ] && export GH_TOKEN="$tok"
+
+or, eval-able form::
+
+    eval "$(python3 scripts/gh_app_env.py)"
+
+Contract (load-bearing for the daemons):
+
+* **Silent-safe** — when App config is absent, the mint fails, or aragora is
+  not importable, exit 0 and print *nothing* so callers degrade gracefully to
+  the operator's existing gh auth instead of crashing the pass.
+* **No leakage** — the token is written only to stdout in the designated
+  output mode; it is never logged, echoed to stderr, or included in
+  diagnostics.
+* **Per-pass freshness** — installation tokens expire after one hour, so
+  callers must invoke this at the top of each pass, never once at daemon
+  start. Each invocation is a fresh process, so the token is always fresh.
+
+Callers that export the result should also export
+``ARAGORA_GITHUB_AUTH_SOURCE=github_app_installation`` so
+:func:`aragora.swarm.github_app_auth.github_cli_env` and
+:func:`gh_subprocess_run` can recognize (and drop, for write ops) the
+App-sourced token.
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+from pathlib import Path
+from typing import Callable
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+
+
+def _resolve_minter() -> Callable[[], str | None]:
+    """Import the App token minter; raises if aragora is unavailable."""
+    if str(REPO_ROOT) not in sys.path:
+        sys.path.insert(0, str(REPO_ROOT))
+    from aragora.swarm.github_app_auth import get_github_app_installation_token
+
+    return get_github_app_installation_token
+
+
+def main(argv: list[str] | None = None, minter: Callable[[], str | None] | None = None) -> int:
+    parser = argparse.ArgumentParser(
+        description=(
+            "Mint a GitHub App installation token for shell gh calls. "
+            "Prints nothing (exit 0) when App config is absent."
+        ),
+    )
+    parser.add_argument(
+        "--print-token",
+        action="store_true",
+        help="Print the bare token to stdout (for command substitution).",
+    )
+    parser.add_argument(
+        "--quiet",
+        action="store_true",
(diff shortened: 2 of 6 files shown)


an0mium:

Head: 3168d1a (3168d1adb5d6d054039e1f2a47b5702902e66070), committed 2026-06-11T20:10:41Z.
PR: #8208.
Model family: grok

Verdict: MERGE

**Tagged findings**

- [TOKEN LEAKAGE] Mitigated. Token material is emitted solely via designated stdout paths (`--print-token` or `GH_TOKEN=` assignment). All diagnostic paths, including the config-absent message, are routed to stderr only when `--quiet` is absent and explicitly exclude token content. The `unset tok` and subshell scoping in both daemon scripts prevent accidental expansion into `ps`, logs, or files. Tests explicitly assert absence from `captured.err` across all modes.
- [FAIL-SAFE DEGRADATION] Robust. Every failure path (absent config, mint exception, import failure, whitespace-only token) forces `token = None`, returns exit 0, and emits nothing on stdout. Callers wrap with `|| true` and `[ -n ]` guards; the arbiter section is further isolated in a subshell so a failure cannot propagate via `set -euo pipefail`. Default-off (`${ARAGORA_GH_APP_AUTH:-0}`) ensures the feature is inert until explicitly enabled.
- [STALE TOKEN] Prevented. Mint occurs at the head of each publish pass or bounded auto-evidence subshell; tokens are never persisted across launchd invocations or inherited by the long-running arbiter `exec`. The Python `github_app_auth` layer already refreshes per call, and the shell wrapper adds the one-hour lifetime boundary.
- [WRITE-OP IDENTITY] Addressed. The `github_cli_env(prefer_app=False)` change now calls `_drop_app_token` when `ARAGORA_GITHUB_AUTH_SOURCE=github_app_installation` is present, guaranteeing operator PAT/keyring precedence for write operations. Untagged user tokens are left untouched.
- [DEFAULT-OFF + BASH 3.2] Compliant. The guard uses only POSIX-portable parameter expansion and `[[` tests; no `&>` , associative arrays, or other post-3.2 features. Bash `-n` validation on both daemon scripts is included in the PR.

The changes are narrowly scoped, defensively implemented, and accompanied by targeted tests. No material objections remain.

dogfood: yes

an0mium:
## Mistral independent model review

Reviewer: mistral (mistral) — independent adversarial model review via the Aragora Mistral reviewer, grounded on the exact PR head.
Head: 3168d1a (3168d1adb5d6d054039e1f2a47b5702902e66070), committed 2026-06-11T20:10:41Z.
PR: #8208.
Model family: mistral

Verdict: MERGE

The rebuttal effectively addresses all three prior findings with technical precision and evidence:

1. **Environment/History Exposure**: The rebuttal correctly notes that:
   - macOS lacks `/proc` entirely (non-issue)
   - Linux `/proc/PID/environ` is UID-restricted (same trust domain as operator keys)
   - GitHub's CLI *requires* env-var token passing (no viable alternative)
   - "Unset after exec" is architecturally impossible (post-exec code never runs)
   *Weakness*: No mitigation for potential `ps e` leakage, but this is a systemic shell limitation, not a PR-specific flaw.

2. **/tmp Logging in --quiet**: The rebuttal clarifies:
   - `--quiet` is caller-optional (not a silent-safe violation)
   - The publisher already emits a non-sensitive degradation line
   - Stderr diagnostics are suppressed in quiet mode
   *Strength*: Explicit design contract adherence; no token material is ever logged.

3. **Token Lingering on Exec Failure**: The rebuttal demonstrates:
   - Exports occur in a subshell (parent env remains pristine)
   - Exec failure triggers parent's `|| echo` with no token exposure
   *Strength*: Process isolation is correctly implemented; the parent never sees the token.

**Additional Merits**:
- The PR's opt-in wiring (default OFF) and per-pass minting align with the codebase's conservative security posture.
- The `ARAGORA_GITHUB_AUTH_SOURCE` tagging enables precise token routing (write ops retain operator PAT).
- Test coverage (12 new tests) validates silent-safe degradation and no leakage.

**Residual Risks**:
- Shell environments are inherently less secure than Python's controlled execution. However, this is a systemic constraint, not a PR regression. The PR *reduces* risk by shifting from operator PAT to ephemeral App tokens.

The rebuttal's technical accuracy and the PR's design rigor justify merging. No unresolved security concerns remain.

dogfood: yes

PR #8253, opened 2026-06-12
fix(benchmark-truth): validate corpus freshness map schema

an0mium:
Summary:
- fail closed on invalid corpus freshness map schema_version values
- validate the schema version on both load and write paths
- add focused regression coverage for rejected and preserved schema versions

Validation:
- /Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest tests/scripts/test_build_benchmark_truth_artifact.py


scarmani:
Queue-drain close: this ownerless draft has been idle since 2026-06-12. The current loop cannot mark drafts ready, and there is no active lane or steering for this branch. Closing preserves the branch and keeps recovery reversible; reopen from head 58b9d15005b51696270bb1de7dee91a586434f13 if this corpus freshness schema guard should continue.

PR #8185, opened 2026-06-11
fix(review-queue): avoid merge-packet startup hangs

an0mium:
## Summary

- add a lightweight `review-queue` CLI fast path so `merge-packet` does not initialize the full CLI/modes/secrets surface
- defer heavy review/swarm/provider imports from merge-packet metadata paths while preserving existing monkeypatch/public seams
- add import-boundary regressions that block eager imports of full parser, modes, secrets, worktree, post-merge audit, agent factory, and asyncio for metadata-only paths

## Validation

- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest tests/cli/commands/test_review_queue_import_boundaries.py tests/cli/commands/test_review_queue.py -q` -> 228 passed
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest tests/swarm/test_pr_review_protocol.py::test_execute_live_reviewers_parses_structured_outputs tests/review/test_provider_slots.py::test_provider_slot_summary_counts_core_and_opt_in_slots -q` -> 2 passed
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m ruff check aragora/cli/commands/review_queue.py aragora/cli/main.py aragora/review/__init__.py aragora/review/provider_slots.py aragora/swarm/__init__.py aragora/swarm/pr_review_protocol.py tests/cli/commands/test_review_queue_import_boundaries.py` -> All checks passed
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m ruff format --check aragora/cli/commands/review_queue.py aragora/cli/main.py aragora/review/__init__.py aragora/review/provider_slots.py aragora/swarm/__init__.py aragora/swarm/pr_review_protocol.py tests/cli/commands/test_review_queue_import_boundaries.py` -> 7 files already formatted
- `git diff --check` -> clean
- default `python3 -m aragora.cli.main review-queue merge-packet --pr 7841 --json` -> rc 0, JSON returned in 4.73s, head `d61d14204cad763e8347dba9a893d63981081dda`, status `repair_or_wait`, verdict `not_ready_for_settlement`
- default `python3 scripts/settle_tier4_pr.py --check --pr 7841 --head d61d14204cad763e8347dba9a893d63981081dda --json` -> rc 1 with structured JSON blockers instead of a timeout/no-output hang
- default `python3 scripts/settle_one_pr.py --pr 7841 --json` -> rc 0 with structured JSON blockers instead of a timeout/no-output hang

## Caveat

[...]

PR #8286, opened 2026-06-12
fix(tests): debate-nz shard failures — #8185 smoke-test contract + rlm load_dotenv env injection (#8277)

### LINKED ISSUE #8277: test pollution: pr_review_protocol smoke fails in debate-nz shard on main (AgentRegistry loses registrations under xdist)
## Symptom
`tests/debate/test_pr_review_protocol_smoke.py::test_pr_review_protocol_packet_smoke` fails with `assert None == 'claude'` (provider_slots[0].selected_provider is None) in the **debate-nz** CI shard, while passing in isolation and in small groups.

## Evidence it is main, not any PR
Reproduced locally on a clean worktree of **origin/main** (3e8450d52b) with the exact CI shard command:

```
PYTHONPATH=.:aragora-debate/src pytest $(python scripts/ci_resolve_test_shard.py debate-nz)   -m "not slow and not load and not e2e and not integration and not integration_minimal and not benchmark and not performance"   --timeout=120 -n auto --dist loadscope -q
# => 1 failed, 5640 passed: test_pr_review_protocol_packet_smoke
```

Observed failing on PR #8255's runs twice (different heads), which is what prompted the bisection — the PR merely shifts xdist module distribution.

## Mechanism (hypothesis, partially verified)
`ProviderSlotResolver._evaluate_candidate` selects on `check.available`, which requires `AgentRegistry.get_spec('claude')` to return a spec. The test patches `shutil.which`, so the only way to get `selected_provider=None` for claude is the registry losing its registrations — i.e. some test in the same xdist worker leaves `AgentFactory._registry` empty/replaced. `register_all_agents()` cannot self-heal because `aragora.agents.cli_agents` stays cached in `sys.modules`, so its `@AgentRegistry.register` decorators never re-execute. No `importlib.reload`/`sys.modules.pop` of the registry exists in the shard's test files, so the pollution path is indirect (mock/patch leak on the class attribute, or a conftest fixture).

## Suggested fix directions
1. Track down the polluter with `--randomly-seed` from a failing CI run + bisection within the failing worker's module set.
2. Defense in depth regardless: make `register_all_agents()` self-healing (e.g. each agent module exposes an explicit `register()` that `register_all_agents` always calls, instead of relying on import side-effects), or snapshot/restore `AgentFactory._registry` in an autouse fixture like the existing mock-descriptor restoration.

an0mium:
Fixes #8277

## What the issue thought was happening
#8277 attributed the `test_pr_review_protocol_packet_smoke` failure to xdist test pollution emptying `AgentFactory._registry`. That diagnosis doesn't hold: the test fails **in isolation** at the issue's own repro commit (3e8450d52b). Verified by bisection: passes at 58251e67c4^, fails at 58251e67c4.

## What is actually happening (two independent bugs)

**Bug 1 — deterministic test break by #8185.** Commit 58251e67c4 routes `build_packet()` without live reviewer results through `metadata_provider_slots()`, which intentionally skips provider probing (`selected_provider=None`, `status="not_probed"`). The smoke test still asserted probed selections.

**Bug 2 — real worker-wide env pollution, but not the registry.** The official `rlm` package calls `dotenv.load_dotenv()` when `rlm.clients` is imported. Under pytest-xdist, execnet workers' `__main__` has no `__file__`, so python-dotenv's `find_dotenv()` treats the process as interactive and searches **from cwd upward** — loading the repository `.env` (including `ARAGORA_SECRETS_STRICT=true`) into `os.environ` of every worker at conftest-import time (chain: tests/conftest.py → slack handlers → rbac → control_plane → knowledge → documents.chunking → aragora.rlm.bridge → rlm). Strict secrets mode then makes `get_secret()` raise `SecretNotFoundError` for critical secrets instead of using env vars, which failed every test touching the provider-credential path. Single-process runs search from site-packages and find nothing — hence "fails in shard, passes in isolation". This explains all 9 debate-nz failures observed on origin/main locally (pinpointed by tracing `os._Environ.__setitem__` with stack capture).

## Changes
- **Smoke test updated** to assert the post-#8185 metadata-only contract; a second test passes `execution_failures` to exercise the probed branch and keeps the original provider-selection assertions.
- **`aragora.utils.env.preserve_environ()`** context manager wraps both official-rlm import sites (`aragora/rlm/bridge.py`, `aragora/server/handlers/features/rlm.py`), rolling back any import-time env mutation. This also fixes a production-grade issue: importing `aragora.knowledge` from any directory under a tree containing `.env` mutated process env.
- **Defense-in-depth requested by the issue:** `register_all_agents()` is now self-healing — `@AgentRegistry.register` records each spec on the agent class, and a healing pass re-applies specs from already-imported `aragora.agents.*` modules, so a cleared or rebound `AgentFactory._registry` recovers despite `sys.modules` caching. Existing entries are never overwritten; test-defined agents (in `tests.*`) are never resurrected.

## Verification
3 consecutive full debate-nz shard runs (exact CI command, pyenv 3.11, `-n auto --dist loadscope`):
- Run 1: **5642 passed, 2 skipped** (1:43)
- Run 2: **5642 passed, 2 skipped** (2:26)
- Run 3: **5642 passed, 2 skipped** (2:58)

Baseline on clean origin/main with the same command: **9 failed** (incl. structural_rhetoric ×4, post_debate_pipeline_bridge ×3, post_debate_workflow_fallback ×1 — all the same strict-mode pollution).

New unit coverage: `tests/utils/test_preserve_environ.py` (5 tests), 3 self-healing registry tests in `tests/agents/test_registry.py`.

### DIFF
--- a/aragora/agents/registry.py
+++ b/aragora/agents/registry.py
@@ -156,6 +156,15 @@ class AgentFactory:
                 accepts_api_key=accepts_api_key,
             )
             cls._registry[type_name] = spec
+            # Record the spec on the class itself so register_all_agents()
+            # can self-heal a cleared/rebound registry without re-importing
+            # (decorators never re-run while the module is cached in
+            # sys.modules). vars() lookup avoids inheriting a parent's specs.
+            own_specs = vars(agent_cls).get("__registry_specs__")
+            if own_specs is None:
+                own_specs = []
+                setattr(agent_cls, "__registry_specs__", own_specs)
+            own_specs.append(spec)
             return agent_cls
 
         return decorator
@@ -463,6 +472,11 @@ def register_all_agents() -> None:
 
     This function should be called once at startup to ensure
     all agents are registered before create() is used.
+
+    Safe to call repeatedly: if the registry was cleared (or rebound) after
+    the agent modules were first imported, the registrations are re-applied
+    from specs recorded on the agent classes, since the cached modules'
+    @AgentRegistry.register decorators never re-run.
     """
     # Import modules to trigger @register decorators
     # These imports are intentionally side-effect only
@@ -480,3 +494,24 @@ def register_all_agents() -> None:
         from aragora.agents import api_agents  # noqa: F401
     except ImportError:
         logger.debug("api_agents module not available for registration")
+
+    _heal_registry()
+
+
+def _heal_registry() -> None:
+    """Re-apply registrations recorded on already-imported agent classes.
+
+    Only production modules (``aragora.agents.*``) are scanned, so agent
+    classes registered ad hoc by tests are never resurrected. Existing
+    entries are never overwritten, so deliberate overrides survive.
+    """
+    import sys
+
+    for module_name, module in list(sys.modules.items()):
+        if module is None or not module_name.startswith("aragora.agents"):
+            continue
+        for obj in list(vars(module).values()):
+            if not isinstance(obj, type):
+                continue
+            for spec in vars(obj).get("__registry_specs__", ()):
+                AgentFactory._registry.setdefault(spec.name, spec)
--- a/aragora/rlm/bridge.py
+++ b/aragora/rlm/bridge.py
@@ -75,9 +75,14 @@ def _secret_configured(name: str) -> bool:
     return get_secret_presence(name).source in {"aws", "env"}
 
 
-# Check if official RLM is available
+# Check if official RLM is available.
+# preserve_environ: importing rlm triggers a load_dotenv() side effect that
+# can inject a repository .env into os.environ process-wide (#8277).
+from aragora.utils.env import preserve_environ
+
 try:
-    from rlm import RLM as OfficialRLM
+    with preserve_environ():
+        from rlm import RLM as OfficialRLM
 
     HAS_OFFICIAL_RLM = True
 except ImportError:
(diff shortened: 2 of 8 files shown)


an0mium:
**CI triage** (all 5 required checks pass; non-required failures investigated):

- `portability`, `governance`, `check` ×2 — runner cancellations during checkout at 14:22 (\"The operation was canceled\"), no test output. Re-run triggered.
- `test-fast (server-handlers-am)`, `test-fast (handlers-amk-no-features)`, `Integration Smoke (handlers)` — pre-existing order-dependent pollution, not from this diff:
  - The CI failures (`test_explainability*.py`, `'ExplainabilityHandler' object has no attribute '_handle_summary'`) pass 88/88 locally on both this branch and clean main, in isolation and in the local shard run.
  - Running the full `server-handlers-am` shard locally: clean main fails **23** tests; this branch fails **2** (order-dependent `test_audit_trail.py::TestAsyncSafety` flakes that pass in isolation on both branches). The `preserve_environ` guard in this PR is what removes the other 21.
  - The `Tests` workflow has been failing on main nightly since at least Jun 8.
- `aragora-merge-quorum` — review-quorum gate, expects the review flow, not a code failure.
- `Release Readiness` — downstream of the above.

scarmani:
Closing this stale conflicting PR to reduce queue pressure. The branch has been open since 2026-06-12, is now conflicting/dirty against current main, has no active owner or PR-specific steering, and cannot be merged or settled at this head. No branch deletion is requested; the debate-nz/rlm test fix can be deliberately revived or re-cut if still needed.

PR #8309, opened 2026-06-12
fix(reports): validate capability gap maps

an0mium:
Summary: fail closed when the capability gap report receives malformed capability catalog or surface map shapes instead of silently publishing empty data or crashing unclearly. Validation: /Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest -q tests/scripts/test_capability_gap_report.py; /Users/armand/.pyenv/versions/3.11.11/bin/python -m ruff check scripts/capability_gap_report.py tests/scripts/test_capability_gap_report.py; /Users/armand/.pyenv/versions/3.11.11/bin/python scripts/capability_gap_report.py --format json; git diff --check; bash scripts/automation_pr_preflight.sh origin/main HEAD.

### DIFF
--- a/scripts/capability_gap_report.py
+++ b/scripts/capability_gap_report.py
@@ -68,8 +68,21 @@ def _load_yaml(path: Path) -> dict[str, Any]:
     with path.open("r", encoding="utf-8") as handle:
         data = yaml.safe_load(handle) or {}
     if not isinstance(data, dict):
-        return {}
+        raise ValueError(f"{path} must contain a YAML mapping")
     return data
+
+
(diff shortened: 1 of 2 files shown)


scarmani:
Queue-drain close: this ownerless report-guard draft has been idle since 2026-06-12 and is now dirty/conflicting against current `main`. I am not spending a repair pass on a stale draft with no live owner in this drain loop.

This is reversible: the branch `codex/capability-gap-map-shape-guard-20260612` is preserved at `51a691f7f5b97e68ef70645202468ac87196aeaa` for deliberate revival if the capability gap map guard is still wanted.

Issue #6026, opened 2026-04-16
[DIC-16] Extend receipts and Knowledge Mound for claim/crux provenance


an0mium:
## Goal
Persist executable claim and CruxSet outputs into receipt and Knowledge Mound paths with provenance, evidence links, and later-settlement hooks.

## Why
The system should remember not only decisions, but the claims and cruxes that made them fragile. That is the bridge from debate output to organizational memory.

## Existing substrate
- aragora/export/decision_receipt.py has verified_claims and unverified_claims.
- aragora/knowledge/mound/adapters/receipt_adapter.py converts receipt claims into knowledge items.
- aragora/knowledge/mound/adapters/performance/reverse_flow.py already references crux_resolver patterns.

## Acceptance
- Document and/or extend receipt metadata for executable claims and cruxes.
- Ensure Knowledge Mound ingestion can preserve claim/crux IDs, evidence IDs, source receipt, and verification status.
- Add focused tests around receipt ingestion behavior.
- Keep backwards compatibility with existing receipt formats.

scarmani:
Closing as implemented on current origin/main (a6db90d251f3c449610e81be889370b96f818f82).\n\nAcceptance proof:\n- Merged #6376 / 27043afaa4 adds the signed CruxReceipt artifact for crux-finder runs.\n- Merged #6431 / b234ab0968 adds optional, backward-compatible claim_id, crux_id, evidence_ids, verification_status, and source_receipt_id fields plus claim/crux bridge helpers.\n- The standard receipt adapter preserves those fields in Knowledge Mound item metadata.\n- Merged #6635 / 6d48ae15e4 adds dedicated CruxReceipt ingestion preserving crux IDs, receipt/debate linkage, affected claims, load-bearing scores, and receipt checksum.\n- Focused current-main validation passed: 128 receipt export, backward-compatibility, CruxReceipt, and Knowledge Mound ingestion tests.\n- The issue still has no boss-ready or autonomous labels.\n\nThis closes the issue only; implementation history and branches remain preserved.

Issue #6025, opened 2026-04-16
[DIC-15] Add CruxSet contract and crux-finder consensus mode


an0mium:
## Goal
Introduce a CruxSet contract for 3-5 load-bearing disagreements and add an Arena consensus mode that produces ranked crux candidates instead of only an answer.

## Why
Aragora should locate where reasonable agents diverge, not merely synthesize answers. This complements Epistemic CI: claims say what must stay true; cruxes say what would change the conclusion.

## Existing substrate
- DebateProtocol consensus modes already route debate behavior.
- aragora/debate/convergence.py can be inverted from agreement detection into divergence scoring.
- aragora/interrogation/ already prioritizes questions from debate signal.
- aragora/explainability/builder.py already has counterfactual machinery.

## Acceptance
- Define Crux and CruxSet data contracts.
- Add a non-default crux-finder consensus path or adapter.
- Include ranking fields: statement, opposing positions, load-bearing score, evidence gaps, and candidate verifier.
- Add focused tests using deterministic mocked agent outputs.
- Do not replace existing consensus modes.

an0mium:
Concrete implementation plan now lives in [docs/plans/2026-04-16-crux-mode-design.md](https://github.com/synaptent/aragora/blob/main/docs/plans/2026-04-16-crux-mode-design.md) and epic #6035 with subtasks #6036–#6039. This issue can close when A1–A3 merge; proposing we leave it open as the originating design thread.

Key deltas from this ticket's original description:
- The `CruxDetector` machinery already exists (`aragora/reasoning/crux_detector.py`) — the work is wiring, not building.
- Approach A (thin wiring) ships first; protocol-shaping (divergence scoring, agent prompt shaping) is Track B, gated on measurable dogfood triggers.
- `CruxReceipt` is introduced as a distinct artifact from `DecisionReceipt` — the deliverable is explicitly *not* a verdict.

an0mium:
## Implementation-level plan available

Design doc [docs/plans/2026-04-16-crux-mode-design.md](https://github.com/synaptent/aragora/blob/main/docs/plans/2026-04-16-crux-mode-design.md) (PR #6041) has concrete A1/A2/A3 breakdowns for CruxSet contract, CruxReceipt with SHA-256 signing, and `aragora crux` CLI verb — all grounded in the existing `CruxDetector` machinery. It was filed under a parallel epic (#6035, now closed) before I had the latest `main` with the DIC-13..22 canonical sequence.

[...]

PR #8366, opened 2026-06-13
[DIC-15] CruxSet arena-path bridge — CruxFinderResult → CruxSet

an0mium:
## Slice

Adds `maybe_emit_cruxset_from_finder_result()` to `aragora/reasoning/cruxset_emission.py` — the missing direct bridge from a completed `CruxFinderResult` (Arena output) to the DIC-15 `CruxSet` contract. The `CruxSet` dataclass, `maybe_emit_cruxset`, and `consensus="crux_finder"` registration were already on `main`; what was absent was a single call-site for Arena callers that hold a `CruxFinderResult` and want the signed, ranked `CruxSet` output.

## Gating

- Default: **OFF** (flag: `ARAGORA_CRUXSET_EMISSION_ENABLED`)
- Live queue effect: **none**
- Advances issue: #6025

## Tests

- `test_returns_none_when_disabled` — flag off → function exits before any debate import
- `test_returns_cruxset_when_enabled` — flag on, result with cruxes → `isinstance(cs, CruxSet)`
- `test_decision_is_always_none` — crux-finder never produces a verdict; `cs.decision is None`
- `test_cruxes_sorted_by_load_bearing_score_desc` — ranking preserved from analyser output
- `test_provenance_carries_debate_id_and_mode` — provenance links `debate_id`, `mode="crux_finder"`, `approach="A"`
- `test_extra_provenance_merged` — `extra_provenance` kwarg merged without losing base provenance
- `test_receipt_id_threaded_through` — `receipt_id` kwarg passes through to `CruxSet.receipt_id`
- `test_checksum_valid_and_schema_version_set` — `cs.verify_checksum()` is True; schema version correct
- `test_returns_none_when_analysis_has_no_cruxes` — empty crux list → None (not a ValueError)
- `test_returns_none_for_wrong_type` — isinstance guard fires for non-`CruxFinderResult` inputs
- `test_convergence_barrier_in_counterfactual_notes` — `build_cruxset_from_analysis` embeds the barrier signal

## Validation

- `pytest tests/reasoning/test_dic15_cruxset_arena_bridge.py --noconftest` — **11 passed**
- `ruff check aragora/reasoning/cruxset_emission.py tests/reasoning/test_dic15_cruxset_arena_bridge.py` — **clean**
- `mypy aragora/reasoning/cruxset_emission.py --ignore-missing-imports` — **clean**

## Out of scope

- Wiring the bridge into the Arena's consensus phase (that's a follow-up once DIC-16 receipt persistence lands)
- Persisting the returned `CruxSet` to the receipt or KM stores (DIC-16 / #6026)
- Testing `DebateProtocol(consensus="crux_finder")` mode registration (already covered by `tests/debate/test_crux_mode.py::test_crux_finder_literal_accepted` on main)

---
_Generated by [Claude Code](https://claude.ai/code/session_01CXKiLCA2iiLhAepPtPktWk)_

### DIFF
--- a/aragora/reasoning/cruxset_emission.py
+++ b/aragora/reasoning/cruxset_emission.py
@@ -127,9 +127,65 @@ def maybe_emit_cruxset(
         return None
 
 
+def maybe_emit_cruxset_from_finder_result(
+    result: Any,
+    *,
+    receipt_id: str = "",
+    extra_provenance: dict[str, Any] | None = None,
+) -> CruxSet | None:
+    """Arena-path bridge for DIC-15: CruxFinderResult → CruxSet.
+
+    Converts the output of a ``consensus="crux_finder"`` debate run into the
+    DIC-15 :class:`~aragora.reasoning.cruxset.CruxSet` contract. Returns
+    ``None`` when emission is disabled (the default).
+
+    Call this after :func:`~aragora.debate.crux_mode.build_crux_finder_result`
+    to obtain the ranked, signed CruxSet that downstream receipt ingestion
+    (DIC-16) and the follow-up bridge (DIC-17) consume.
+
+    ``decision`` is always ``None`` — a crux-finder run never produces a
+    verdict by design. Provenance is seeded from ``result.debate_id`` and
+    can be extended via ``extra_provenance``.
+
+    Flag: ``ARAGORA_CRUXSET_EMISSION_ENABLED`` (default ``False``).
+    Live queue effect: none.
+    """
+    if not cruxset_emission_enabled():
+        return None
+
+    from aragora.debate.crux_mode import CruxFinderResult  # lazy — avoids import cycle
+
+    if not isinstance(result, CruxFinderResult):
+        logger.warning(
+            "maybe_emit_cruxset_from_finder_result: expected CruxFinderResult, got %s",
+            type(result).__name__,
+        )
+        return None
+
+    provenance: dict[str, Any] = {
+        "debate_id": result.debate_id,
+        "mode": "crux_finder",
+        "approach": result.metadata.get("approach", "A"),
+        "rounds": result.rounds,
+        "agents": list(result.agents),
+    }
+    if extra_provenance:
+        provenance.update(extra_provenance)
+
+    return maybe_emit_cruxset(
+        question=result.question,
+        analysis_payload=result.analysis.to_dict(),
+        decision=None,  # crux_finder never produces a verdict by design
+        receipt_id=receipt_id,
+        provenance=provenance,
+        top_k=len(result.analysis.cruxes) or 5,
+    )
+
+
 __all__ = [
     "CRUXSET_EMISSION_ENV_VAR",
     "cruxset_emission_enabled",
     "enable_cruxset_emission",
     "maybe_emit_cruxset",
+    "maybe_emit_cruxset_from_finder_result",
 ]
--- a/aragora/reasoning/cruxset_emission.py
+++ b/aragora/reasoning/cruxset_emission.py
@@ -153,7 +153,13 @@ def maybe_emit_cruxset_from_finder_result(
     if not cruxset_emission_enabled():
         return None
 
-    from aragora.debate.crux_mode import CruxFinderResult  # lazy — avoids import cycle
+    try:
+        from aragora.debate.crux_mode import CruxFinderResult  # lazy — avoids import cycle
+    except Exception as exc:  # noqa: BLE001 - degrade gracefully like the rest of this function
+        logger.warning(
+            "maybe_emit_cruxset_from_finder_result: could not import CruxFinderResult: %s", exc
+        )
+        return None
 
     if not isinstance(result, CruxFinderResult):
         logger.warning(
@@ -170,6 +176,7 @@ def maybe_emit_cruxset_from_finder_result(
         "agents": list(result.agents),
     }
     if extra_provenance:
+        # extra_provenance intentionally overwrites base keys — caller's responsibility
         provenance.update(extra_provenance)
 
     return maybe_emit_cruxset(
@@ -178,7 +185,7 @@ def maybe_emit_cruxset_from_finder_result(
         decision=None,  # crux_finder never produces a verdict by design
         receipt_id=receipt_id,
         provenance=provenance,
-        top_k=len(result.analysis.cruxes) or 5,
+        top_k=max(len(result.analysis.cruxes), 1),
     )
 
 
--- a/aragora/reasoning/cruxset_emission.py
+++ b/aragora/reasoning/cruxset_emission.py
@@ -168,24 +168,36 @@ def maybe_emit_cruxset_from_finder_result(
         )
         return None
 
-    provenance: dict[str, Any] = {
-        "debate_id": result.debate_id,
-        "mode": "crux_finder",
-        "approach": result.metadata.get("approach", "A"),
-        "rounds": result.rounds,
-        "agents": list(result.agents),
-    }
+    try:
+        metadata = result.metadata or {}
+        analysis = result.analysis
+        analysis_payload = analysis.to_dict()
+        cruxes = list(analysis.cruxes)
+        counterfactuals = list(result.counterfactuals or [])
+        provenance: dict[str, Any] = {
+            "debate_id": result.debate_id,
+            "mode": "crux_finder",
+            "approach": metadata.get("approach", "A"),
+            "rounds": result.rounds,
+            "agents": list(result.agents),
+        }
+    except Exception as exc:  # noqa: BLE001 - malformed finder output must fail closed
+        logger.warning("maybe_emit_cruxset_from_finder_result: malformed CruxFinderResult: %s", exc)
+        return None
+
+    if counterfactuals:
+        provenance["counterfactuals"] = counterfactuals
     if extra_provenance:
         # extra_provenance intentionally overwrites base keys — caller's responsibility
         provenance.update(extra_provenance)
 
     return maybe_emit_cruxset(
         question=result.question,
-        analysis_payload=result.analysis.to_dict(),
+        analysis_payload=analysis_payload,
         decision=None,  # crux_finder never produces a verdict by design
         receipt_id=receipt_id,
         provenance=provenance,
-        top_k=max(len(result.analysis.cruxes), 1),
+        top_k=max(len(cruxes), 1),
     )
 
 
--- a/aragora/reasoning/cruxset_emission.py
+++ b/aragora/reasoning/cruxset_emission.py
@@ -155,7 +155,7 @@ def maybe_emit_cruxset_from_finder_result(
 
     try:
         from aragora.debate.crux_mode import CruxFinderResult  # lazy — avoids import cycle
-    except Exception as exc:  # noqa: BLE001 - degrade gracefully like the rest of this function
+    except BaseException as exc:  # noqa: BLE001 - pyo3_runtime.PanicException is not an Exception
         logger.warning(
             "maybe_emit_cruxset_from_finder_result: could not import CruxFinderResult: %s", exc
         )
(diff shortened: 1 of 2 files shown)


scarmani:
## Focused adversarial dogfood

Model family: openai
PR: #8366
Head: f999fcd8d5deccdff7084c403c320ffb9012904a
Verdict: PASS

dogfood: yes

I ran focused current-head dogfood from a disposable PR-head worktree at `/private/tmp/aragora-8366-evidence.9z26bO`.

Command:
`PATH=/Users/armand/.pyenv/versions/3.11.11/bin:$PATH pytest -q tests/reasoning/test_dic15_cruxset_arena_bridge.py`

Result: `11 passed in 0.60s`.

Coverage checked the CruxFinderResult -> CruxSet arena-path bridge: default-disabled returns `None`, enabled emits `CruxSet`, `decision` remains `None`, cruxes rank by `load_bearing_score`, debate provenance/extra provenance/receipt id are threaded, checksum/schema validate, empty and wrong-type inputs fail closed, and convergence-barrier metadata is preserved.

No blocking issues found.

scarmani:

Head: f999fcd (f999fcd8d5deccdff7084c403c320ffb9012904a), committed 2026-06-13T08:19:11Z.
PR: #8366.
Model family: claude

Verdict: PASS

No blocking issues. Concrete findings:

- [P2] `tests/reasoning/test_dic15_cruxset_arena_bridge.py:50-58` — the autouse `_inject_crux_mode_stub` fixture replaces `aragora.debate.crux_mode` with a hand-rolled `_CruxFinderResult` dataclass. The bridge under test is therefore never exercised against the real `CruxFinderResult`. If the real class renames fields (`debate_id`, `question`, `metadata`, `agents`, `rounds`, `analysis.to_dict()`, `analysis.cruxes`) or its `to_dict()` shape diverges, prod breaks while these tests stay green. The header comment asserts "In CI the real class is used and the same tests pass" — that claim is unverifiable from this PR and the fixture is `autouse`, so CI will also get the stub. Either drop the stub when the real import succeeds (try real import first, fall back to stub only on ImportError), or add a single non-stubbed integration test gated on the cryptography backend being available.
- [P3] `aragora/reasoning/cruxset_emission.py:170-173` — `provenance.update(extra_provenance)` lets callers silently overwrite base keys (`debate_id`, `mode`, `approach`, `rounds`, `agents`). Probably fine for internal callers, but the docstring says provenance "can be extended via `extra_provenance`" which reads as additive. If overwrite is intentional, say so; otherwise reverse the order (`{**extra, **base}`) or reject conflicting keys. Tests don't cover this.
- [P3] `aragora/reasoning/cruxset_emission.py:177` — `top_k=len(result.analysis.cruxes) or 5` falls back to 5 when `cruxes` is empty. Currently masked because `maybe_emit_cruxset` short-circuits on empty cruxes (per `test_returns_none_when_analysis_has_no_cruxes`), but the `or 5` is a footgun if that upstream contract ever changes — surfacing 5 phantom cruxes from an empty analysis would be a silent regression. Prefer an explicit guard or pass `top_k=max(len(...), 1)` with the empty-list short-circuit kept upstream.
- [P3] `aragora/reasoning/cruxset_emission.py:153` — lazy `from aragora.debate.crux_mode import CruxFinderResult` runs after the flag check but is unguarded. If that import raises (the very condition the test fixture exists to work around), enabling the flag in a degraded environment raises through callers instead of degrading to `None` like the rest of the function. Wrap in try/except ImportError → log + return None, matching the other defensive returns in this function.

Flag is default-off and the surface is bounded to a one-direction conversion, so none of these block merge — but the test-stub divergence (P2) is worth fixing before this becomes load-bearing for DIC-16/DIC-17.

dogfood: yes

scarmani:

Head: f999fcd (f999fcd8d5deccdff7084c403c320ffb9012904a), committed 2026-06-13T08:19:11Z.
PR: #8366.
Model family: grok

Verdict: PASS

- No blocking issues found (correctness, security, or regressions).
- [P3] test_convergence_barrier_in_counterfactual_notes has stale/outdated docstring referencing non-existent `build_cruxset_from_analysis`.
- [P3] `top_k=len(...) or 5` in bridge could silently default to 5 on empty input (but `maybe_emit_cruxset` already guards this per tests).

dogfood: yes

an0mium:
## Follow-up fixes (f52cbfd)

Addressed the P2 and P3 findings from the Claude and Grok reviews:

**P2 — test stub no longer autouse-always:**
`_inject_crux_mode_stub` is now conditional. We try `from aragora.debate.crux_mode import CruxFinderResult` at module level; only when that raises (here via `pyo3_runtime.PanicException` from the broken cryptography backend — not a subclass of `Exception`, so the catch uses `BaseException`) do we fall back to the stub. In CI where the backend is healthy, `_USE_STUB = False` and the fixture is a no-op: tests exercise the real class.

**P3 — `top_k=max(len(cruxes), 1)` instead of `or 5`** (eliminates the phantom-5 footgun).

**P3 — lazy import wrapped in try/except** in `cruxset_emission.py` so a degraded environment degrades to `None` rather than raising through callers.

**P3 — `extra_provenance` overwrite semantics documented** in the docstring.

All 11 tests still pass locally.


---
_Generated by [Claude Code](https://claude.ai/code)_

scarmani:
## Focused adversarial dogfood

Head SHA: 326b697f8b (326b697f8bab3a6c3b170cc495cd1eeaccb5ca22) — adversarial dogfood validation grounded on the exact PR head.
PR: #8366.
Model family: openai

Dogfood runner: Codex/OpenAI current-head focused validation in the exact PR-head worktree `/private/tmp/aragora-8366-repair.xVfUSt`.

Validation command: `pytest -q tests/reasoning/test_dic15_cruxset_arena_bridge.py`
Result: **passed**

Output digest:
```
..............                                                           [100%]
14 passed in 0.84s
```

Coverage focus: CruxFinderResult-to-CruxSet arena bridge fails closed on malformed finder analysis/conversion and preserves counterfactual provenance.

dogfood: yes

This is evidence only, not merge authorization or Tier 4 settlement authorization.

scarmani:

Head: 326b697 (326b697f8bab3a6c3b170cc495cd1eeaccb5ca22), committed 2026-06-16T20:51:47Z.
PR: #8366.
Model family: claude

Verdict: PASS

Findings:

- [P3] `aragora/reasoning/cruxset_emission.py:163-166` — lazy `import CruxFinderResult` is wrapped in `except Exception`, but the test file's `try/except BaseException` block explicitly notes the import can raise `pyo3_runtime.PanicException` (not an `Exception` subclass) on containers with a broken Rust/cffi cryptography backend. If that environment ever runs the bridge, the panic propagates out of the bridge instead of degrading. Low-risk in normal prod, but worth widening to `BaseException` for symmetry with the test guard.

- [P3] `aragora/reasoning/cruxset_emission.py:189-191` — `top_k=max(len(cruxes), 1)`: when the finder yields zero cruxes, you still pass `top_k=1` into `maybe_emit_cruxset`. Behavior depends on whether the downstream emitter returns `None` for empty `cruxes`; the test `test_returns_none_when_analysis_has_no_cruxes` asserts it does, so this is fine today but is an implicit contract — a defensive `if not cruxes: return None` here would make the guarantee explicit at the bridge layer.

- [P3] `aragora/reasoning/cruxset_emission.py:177` — `metadata.get("approach", "A")` silently defaults to "A" when the upstream finder omits/renames the field. If "A" later gets replaced with other approach IDs, stale receipts will be mislabeled rather than flagged. Consider `None`/`"unknown"` instead, or emit a debug log on miss.

- [P3] `tests/reasoning/test_dic15_cruxset_arena_bridge.py:42-62` — when the real `CruxFinderResult` import succeeds, tests construct it via `_ResultClass(debate_id=..., metadata=..., raw_claims=[], ...)`. If the real dataclass adds a required field (or renames one), every test silently breaks even though the bridge is fine. Worth either pinning to the stub unconditionally or adding a single smoke test that uses the real public constructor (`build_crux_finder_result`) to catch field drift.

- [P3] `aragora/reasoning/cruxset_emission.py:183-186` — counterfactuals are stored in `provenance` as list-of-dicts. If `CruxSet.verify_checksum()` JSON-serializes non-deterministically (no `sort_keys`), checksum stability across runs depends on dict insertion order. The test `test_checksum_valid_and_schema_version_set` doesn't exercise counterfactuals, so this isn't covered. Worth one assertion that a CruxSet with counterfactuals round-trips its checksum.

No P1/P2 blockers. The fail-closed exception handling, no-verdict invariant, ranking, and provenance threading all look correct.

dogfood: yes

scarmani:

Head: 326b697 (326b697f8bab3a6c3b170cc495cd1eeaccb5ca22), committed 2026-06-16T20:51:47Z.
PR: #8366.
Model family: grok

Verdict: PASS

- No blocking issues found. Code is defensive (flag-gated, type-checked, fail-closed on import/malformed data, lazy import avoids cycles); tests cover the DIC-15 contract and edge cases; no security surface, no obvious regressions to existing `maybe_emit_cruxset` path.

dogfood: yes

scarmani:
## Focused adversarial dogfood (openai)

Head SHA: 099aefea3e (099aefea3ed5549bb4ce9a7c0be6e03e770a1a6f) - adversarial dogfood validation grounded on the exact PR head.
PR: #8366.
Model family: openai

This is an automated dogfood run by the openai auto-evidence cycle: the PR head was checked out in a disposable worktree and its own scoped validation was executed, including the CruxFinderResult-to-CruxSet bridge and the BaseException lazy import guard follow-up.

Validation commands:
- `pytest -q tests/reasoning/test_dic15_cruxset_arena_bridge.py`
- `python3 - <<'PY' ... patch("builtins.__import__", side_effect=guarded_import) ... maybe_emit_cruxset_from_finder_result(object()) is None`

Result: **passed**

Output digest:
```
099aefea3ed5549bb4ce9a7c0be6e03e770a1a6f
pytest: 14 passed in 1.01s
lazy import BaseException guard returned None
```

dogfood: yes

This is evidence only, not merge authorization or Tier 4 settlement authorization.

scarmani:

Reviewer: claude (anthropic) - independent adversarial model review via the Aragora Claude reviewer, grounded on the exact PR head.
Head: 099aefe (099aefea3ed5549bb4ce9a7c0be6e03e770a1a6f), committed 2026-06-16T21:50:57Z.
PR: #8366.
Model family: claude

Verdict: PASS

No blocking issues. The bridge is correctly flag-gated, the no-verdict invariant (`decision=None`) is enforced, malformed input fails closed, and provenance/receipt threading is verified by tests. Notes:

- [P3] `cruxset_emission.py:200` - `top_k=max(len(cruxes), 1)` substitutes `1` when `cruxes` is empty. Harmless today because `maybe_emit_cruxset` short-circuits on empty `payload["cruxes"]` (line 108), but the `max(..., 1)` floor is misleading - passing `len(cruxes)` directly (or returning early on empty) would be clearer and avoid future regression if the downstream guard ever changes.
- [P3] `cruxset_emission.py:190-192` - `extra_provenance.update` lets callers overwrite `debate_id`, `mode`, `approach`, `rounds`, `agents`. The comment delegates to caller responsibility, but those keys are receipt-traceability anchors signed into the CruxSet checksum. Worth considering a reserved-key guard or at least an explicit warn log if a caller clobbers `debate_id`/`mode`.
- [P3] `cruxset_emission.py:158` - `except BaseException` on the lazy import is unusual but justified by `pyo3_runtime.PanicException` (cryptography backend), matches the test's mirroring rationale, and is bounded to a single `import` statement. OK as-is.
- [P3] Tests `tests/reasoning/test_dic15_cruxset_arena_bridge.py:55-67` - the stub-injection fallback is sound; just note that if the real import later succeeds at runtime but failed at collection, the `_USE_STUB=True` branch still wins because of `monkeypatch.setitem`. Not a correctness issue, but the tests assert only against the dataclass shape - if `CruxFinderResult` ever grows required init fields, only the real-import lane catches the regression. Consider a `pytest.importorskip`-style real-only smoke test in CI.

dogfood: yes

scarmani:

Reviewer: grok (xai) - independent adversarial model review via the Aragora Grok reviewer, grounded on the exact PR head.
Head: 099aefe (099aefea3ed5549bb4ce9a7c0be6e03e770a1a6f), committed 2026-06-16T21:50:57Z.
PR: #8366.
Model family: grok

Verdict: PASS

- No blocking issues found. Code is defensive (fail-closed on import/type/malformed errors, flag-gated, provenance handling correct), tests cover the DIC-15 invariants + edges, and no security/regression vectors identified in the bridge. (Minor: last test docstring references non-existent `build_cruxset_from_analysis`.)

dogfood: yes

PR #8374, opened 2026-06-13
fix(automation): handle audit skip broken pipes

an0mium:
## Summary
- make `scripts/audit_test_skips.py` tolerate downstream pipe closure while writing JSON/summary output
- add focused regression coverage for BrokenPipe handling in audit skip output paths

## Validation
- `PYTHONDONTWRITEBYTECODE=1 PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest -q tests/scripts/test_audit_test_skips.py -p no:rerunfailures -p no:cacheprovider` (3 passed, 2 existing config warnings)
- `python3 -B -m py_compile scripts/audit_test_skips.py tests/scripts/test_audit_test_skips.py`
- `python3 -m ruff check scripts/audit_test_skips.py tests/scripts/test_audit_test_skips.py`
- `python3 -m ruff format --check scripts/audit_test_skips.py tests/scripts/test_audit_test_skips.py`
- `python3 scripts/audit_test_skips.py --count-only` (68)
- `set -o pipefail; python3 scripts/audit_test_skips.py --json | head -1` (emitted `{` and exited 0)
- `python3 scripts/report_code_quality.py --check`
- `git diff --check origin/main...HEAD`
- `git rev-list --left-right --count origin/main...HEAD` (0 2)
- `git cherry -v origin/main HEAD` (one non-merge patch)
- `git merge-tree --write-tree origin/main HEAD` (348ba088e18bdcdf3737baa4e2c19f34818651dc)

## Automation
- Opened from Primary Writer lane `Q525-primary-audit-test-skips-pr-open` after confirming branch steering had no messages and the previous lane was completed/unowned.


an0mium:
## Factory independent model review on head 73db9e2f8e7ebe2ace3cfd341cb1fb2e402a9e1b

**Reviewer harness:** droid
**Model family:** glm
**Model id:** glm-5.1
**Receipt artifact:** /private/tmp/aragora-8374-evidence-pl7UOb/droid-glm-review.md

Independent model review and adversarial dogfood recheck for PR #8374 at exact head 73db9e2f8e7ebe2ace3cfd341cb1fb2e402a9e1b.

## Model Review — PR #8374 (head `73db9e2f8e`)

### Verdict: **PASS**

### Blocking findings: None

### Non-blocking notes

1. **`--update-docs` output atomicity shift** — In the original code, `print(f"Updated {docs_path}")` ran *between* the two `write_text` calls; if `baseline_path.write_text` raised, the user had already seen the first message. The refactored version emits both messages *after* both writes succeed. This is arguably better (no misleading "Updated" if the second write fails) but is a behavioural change worth noting.

2. **`_mute_stdout_after_broken_pipe` not directly unit-tested** — The test file mocks `_mute_stdout_after_broken_pipe` and only verifies it is *called*. There is no test that confirms it actually closes old stdout and reopens `os.devnull`. Low risk since the function is trivial, but a defensive test could catch regressions if someone reworks it.

3. **No test for subsequent `_emit_output` succeeding after muting** — Verifying that a second call writes to devnull without raising would close the loop on the BrokenPipe recovery contract. Again, low risk given the simplicity of the reopen.

4. **Only `BrokenPipeError` is caught, not the broader `OSError`** — This is *correct* (we don't want to mask unrelated I/O errors), but it's worth noting that on some rare platform combinations (old glibc + SIGPIPE race) a plain `OSError(EPIPE)` can surface instead of `BrokenPipeError`. The standard library itself only catches `BrokenPipeError`, so this matches CPython convention.

5. **`--count-only`, `--update-docs`, `--gh_commands`, `--generate_issues` paths are not covered by tests** — The two tests that exist validate `--json` and the default summary path. The other branches are mechanical `print→_emit_output` refactors with identical structure, so the gap is low-risk, but a single parametrised test could close it cheaply.

### Dogfood/recheck summary

At `73db9e2f8e`, the BrokenPipe handling is safe and follows CPython's recommended pattern (close broken fd → reopen devnull), every `print()` in `main()` is correctly replaced with a pipe-safe `_emit_output` that preserves original output formatting, and the two provided tests cover the core mechanism and two representative code paths; no blocking regressions found.

an0mium:
## Claude independent model review on head 73db9e2f8e7ebe2ace3cfd341cb1fb2e402a9e1b

**Reviewer harness:** claude-code
**Model family:** claude
**Model id:** opus via Claude Code 2.1.177
**Receipt artifact:** /private/tmp/aragora-8374-evidence-pl7UOb/claude-review.md

Independent model review and adversarial dogfood recheck for PR #8374 at exact head 73db9e2f8e7ebe2ace3cfd341cb1fb2e402a9e1b.

## Independent Model Review — PR #8374 @ `73db9e2f8e7ebe2ace3cfd341cb1fb2e402a9e1b`

**Verdict: PASS**

### Blocking findings
None.

### Analysis (grounded at head SHA)

**BrokenPipe handling is safe.** `_emit_output` writes payload + newline + flush inside a single `try`, catching `BrokenPipeError`. The flush is what actually surfaces a closed-reader pipe (Python buffers stdout), so triggering it inside the guarded block is correct. On break it calls `_mute_stdout_after_broken_pipe`, which closes the broken stream and rebinds `sys.stdout` to `os.devnull` — the established CPython idiom that prevents the interpreter's exit-time stream flush from re-raising and printing an "Exception ignored" traceback. `close()` can itself raise (the buffered flush hits the dead pipe); that raise is `BrokenPipeError`, a subclass of `OSError`, and is caught by the `except OSError` there. No path leaves a closed `sys.stdout` in place to ValueError on a later write.

**No output-format regression.** I reconstructed each converted branch byte-for-byte:
- `count-only` / `json`: `_emit_output(str(...))` and `_emit_output(json.dumps(...))` reproduce `print(...)` exactly (single trailing `\n`).
- Default summary: the `lines` list with `""` separators joined by `\n` plus the emitter's trailing `\n` reproduces the original `print("\nBy category:")`-style blank-line layout exactly.
- `gh_commands`: header lines + `""` + `*commands` matches the old `print()`-blank-then-loop output, including the empty-`commands` case.
- `update_docs`: both files are still written in the same order; only the two "Updated …" messages moved to a single post-write emit — identical text, identical order, no behavioral change.

**Leftover `print()` calls are correct.** The three remaining `print()`s (lines 337, 646, 686) all target `sys.stderr`, which a downstream `| head` does not close, so they correctly stay outside the pipe-safe path. No inconsistency.

**Tests are sound.** Both `scripts/` and `tests/scripts/` carry `__init__.py`, so `from scripts import audit_test_skips` resolves cleanly. `test_emit_output_mutes_stdout_after_broken_pipe` correctly asserts `flush` is never reached after the write raises, and verifies muting fires.

### Non-blocking notes
- `_mute_stdout_after_broken_pipe` is mocked out in every test, so its actual close-and-reopen-to-devnull logic has no direct coverage. The riskiest line of the PR (the `OSError`-swallowing `close()` + devnull rebind) is unexercised.
- Coverage is limited to the `--json` and default-summary branches; `count-only`, `gh_commands`, and `update_docs` emit paths are untested. None changed semantics, so this is a gap, not a defect.
- Behavior intentionally shifts to exit 0 on broken pipe (mute + normal return) rather than the Python-docs `sys.exit(1)` convention. Reasonable for a `| head`-style consumer; worth a one-line code comment but not a blocker.

### Dogfood/recheck summary
At head `73db9e2f8e7ebe2ace3cfd341cb1fb2e402a9e1b`, the diff's pipe-safe emitter preserves every stdout branch's exact byte layout, correctly leaves stderr diagnostics untouched, and implements the canonical close-stdout-then-rebind-to-devnull BrokenPipe idiom — no behavioral regression found.

PR #8416, opened 2026-06-14
refactor(structex): split tests/conftest.py into tests/fixtures plugins (VAL-P1-007)

scarmani:
## Summary

Tier 0 (tests-only). Splits the 3,044-line `tests/conftest.py` into eight cohesive `tests/fixtures/*.py` plugin modules loaded via `pytest_plugins`, shrinking the root conftest to **561 lines** (< 800). Fulfills **VAL-P1-007** and **VAL-P1-008** of the structural-excellence mission (#8258).

## What changed

- All 46 fixtures extracted as **byte-identical verbatim source slices** into:
  - `tests/fixtures/autouse.py` — every autouse fixture, in original definition order (preserves autouse setup/teardown ordering)
  - `tests/fixtures/tempfiles.py`, `mocks.py`, `services.py`, `env.py`, `sample_data.py`, `api_responses.py`, `clients.py` — non-autouse fixtures grouped by theme
- `tests/conftest.py` retains only: monorepo/Slack bootstrap, the extended `pytest_plugins` list, optional-dependency skip markers, custom marker registration (`pytest_configure`), and skip-count reporting (`pytest_terminal_summary`).
- **Import paths stable:** the only symbols ever imported from `tests.conftest` repo-wide are the skip markers (`requires_no_ci`, `REQUIRES_NO_CI`, `requires_plugins`, `REQUIRES_PLUGINS`); these remain in conftest. No fixture is imported directly anywhere (fixtures are pytest-discovered), so no shim is needed.

## Validation (run on this worktree)

- **VAL-P1-007**: `wc -l < tests/conftest.py` -> **561** (<800); `pytest_plugins` references `tests.fixtures` -> `plugins=0`; non-init plugin modules under `tests/fixtures/` -> **11** (>= 1).
- **VAL-P1-008**: `PYTEST_BIN="python3 -m pytest" bash scripts/test_tiers.sh smoke` -> **exit=0** (138 passed, 228115 deselected, 219.70s).
- **Collect-only unchanged**: `PYTHONPATH=.:aragora-debate/src python3 -m pytest --collect-only -q tests/` -> **228533** on both this branch and current origin/main (rc=0, no collection errors); >= pinned baseline 228385.
- **Autouse ordering preserved**: `pytest -p no:randomly --setup-plan` diff (monorepo + handler tests) is **empty** vs origin/main.
[...]

Issue #8258, opened 2026-06-12
chore(repo): root hygiene — relocate screenshots/.docx/research files out of repo root


an0mium:
## Problem
The repo root carries 11 non-project files (10 FireShot RingRift screenshots + `Aragora Idea-to-Execution Strategy.docx`) plus `Idea-to-Execution-Pipeline-Research.md`. An outsider's first `ls` reads as internal-dogfood clutter and obscures intent; clone weight grows for zero value.

## Evidence (measured 2026-06-12, head a5cf5fc70b)
- `ls *.png *.docx | wc -l` → 11 at repo root.

## Scope
Move screenshots to `docs/assets/` (or delete if obsolete — they appear to be RingRift captures unrelated to Aragora), move the .docx content to a markdown doc under `docs/` or remove, relocate the research md under `docs/`. Update any references.

## Definition of done (falsifiable)
- `ls *.png *.docx 2>/dev/null | wc -l` → 0 at root; baseline `root_clutter` shrinks 11→0.
- Repo root contains only standard project files (code dirs, config, top-level docs).

**Tier:** 1. **Approach:** single small PR; verify nothing imports/links the moved files (`rg -l "FireShot|Idea-to-Execution"`).

**Source audit:** `docs/audits/2026-06-12-codebase-health-audit.md` + machine-readable `docs/audits/2026-06-12-codebase-health-baseline.json` (PR to follow). **Program plan:** `docs/superpowers/plans/2026-06-12-codebase-health-program.md`. Part of #8257.

scarmani:
Closing HEALTH-1 as part of the Structural Excellence epic (#8257) P1 (repo hygiene) phase, completed via the model-quorum merge gate. All P1 work is merged to `main` and the P1 definition-of-done is verified green on a fresh `origin/main` worktree (`b2faf4bb0e`, measured 2026-06-15).

### What landed (merged `structex/p1-*` PRs)
- **#8368** untrack repo-root clutter (10 FireShot screenshots + `*.docx`) + relocate root tests
- **#8369** repo-root allowlist doc (`docs/reference/ROOT_ALLOWLIST.md`) + CI checker (`scripts/ci/check_root_allowlist.py`)
- **#8375** pin `tests_collect_total` pre-migration floor (VAL-P1-009)
- **#8387 / #8404 / #8415** relocate the 635 loose `tests/` root files into mirrored, shard-referenced subdirs (batches 1–3)
[...]

PR #8417, opened 2026-06-14
fix(swarm): give quorum reviewers the complete changed-file list

scarmani:
## Summary

Fixes the deterministic reviewer diff-truncation false-positive in the
`aragora-merge-quorum` gate (P1 hygiene, HEALTH-1 / #8258; unblocks #8416).

`aragora/swarm/quorum_evidence.py` built the collect-evidence reviewer prompt by
slicing the first `_MAX_DIFF_CHARS=60_000` chars of alphabetical `gh pr diff`
output. When a PR's large DELETION sorts alphabetically before its ADDITIONS
(e.g. `tests/conftest.py` deleted before `tests/fixtures/*` added in #8416), the
additions were cut off entirely, so the API reviewer (grok) saw only deletions
and emitted a confident-but-FALSE `Verdict: CHANGES-REQUESTED` ("modules do not
exist / fixtures deleted"), failing the quorum even though the PR was correct
(the CLI reviewer claude, which inspects the repo directly, PASSED).

## Change (gate-correctness PRESERVING)

- `build_review_prompt` now ALWAYS prepends the complete changed-file list
  (`gh pr diff --name-status`, fetched best-effort by `default_prompt_builder`;
  falls back to file headers parsed from the diff when that fetch fails) BEFORE
  the unified-diff body, so a reviewer can never falsely claim a file/module is
  absent.
- The diff body is bounded **per file** (water-filling, smallest first) instead
  of a blind first-N-bytes slice, so every changed file contributes at least a
  hunk; a small addition that sorts after a huge deletion is now kept in full.
- No quorum thresholds, reviewer counts, or builder exit-code/return semantics
  change. The name-status fetch is best-effort and never adds a new raise path.

## Tests (`tests/swarm/test_quorum_evidence.py`)

- a synthetic diff whose DELETION sorts before its ADDITIONS yields a prompt
  containing every added file's path;
- the complete file-list header is never truncated even when the body is;
- the file list is derived from the diff when name-status is absent;
- `_bound_diff_body` water-filling gives every file a hunk and keeps small
  additions whole when a deletion is huge;
- `default_prompt_builder` prepends name-status, tolerates a name-status fetch
  failure, and still raises on an empty diff (return/exit semantics unchanged).

## Validation (local, all green)

- `make lint` -> All checks passed!
- `bash $MISSION/scripts/typecheck_differential.sh` -> TYPECHECK PASS (new=48 <= env-drift 50; `quorum_evidence.py` mypy-clean)
- `python3 -m pytest tests/swarm/test_quorum_evidence.py -q` -> 91 passed (red-first: 10 of the new cases failed before the fix)
- `make test-smoke` -> Smoke tests passed!

## Risks

Low. Tier 2 (`aragora/swarm/`), additive and behavior-neutral for normal PRs
(small diffs are emitted unchanged aside from the always-present complete
file-list header). This STRENGTHENS review accuracy; it does not weaken the gate.

Co-authored-by: factory-droid[bot] <138933559+factory-droid[bot]@users.noreply.github.com>

### DIFF
--- a/aragora/swarm/quorum_evidence.py
+++ b/aragora/swarm/quorum_evidence.py
@@ -95,6 +95,10 @@ QUORUM_STATE_LOCK_STALE_SECONDS = 15 * 60
 _REVIEWER_RESULT_QUEUE_TIMEOUT = 1.0
 # Cap the diff fed to reviewers so a huge PR cannot blow the model context.
 _MAX_DIFF_CHARS = 60_000
+# Marker left in place of a changed file's omitted hunk once the per-file diff
+# budget is exhausted. The complete changed-file list always precedes the body,
+# so a reviewer can still confirm every file is present.
+_PER_FILE_TRUNCATION_MARKER = "\n[hunk truncated; full changed-file list is above]\n"
 # Cap reviewer output so a runaway model cannot exceed GitHub's per-comment limit.
 _MAX_REVIEWER_CHARS = 32_000
 _CLAUDE_TIMEOUT = 300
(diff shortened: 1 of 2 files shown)


scarmani:

Head: 1fa2c1b (1fa2c1bb9359a171e1c5161742de3046131787b0), committed 2026-06-14T14:57:39Z.
PR: #8417.
Model family: claude

Verdict: PASS

No blocking issues. The change is well-scoped and the tests cover the core invariant (`#8416` regression: a large deletion sorting before later additions used to hide them with a blind `diff[:N]` slice). Minor non-blocking observations:

- [P3] `aragora/swarm/quorum_evidence.py:_bound_diff_body` — water-filling allocates equal share per remaining segment. When many files share a small `_MAX_DIFF_CHARS`, the per-file share can drop below the length of the `diff --git a/<path> b/<path>` header itself, so the header is cut mid-line in the body and only the marker remains. Identity is still recoverable from the `=== CHANGED FILES ===` block above, so this is graceful, but the docstring's "each segment keeps its own `diff --git` header" claim isn't strictly true in the degenerate case.
- [P3] `aragora/swarm/quorum_evidence.py:_file_list_from_diff` — path extraction uses `rest.rfind(" b/")`. Git emits C-style quoted paths (`diff --git "a/…" "b/…"`) for paths with quotes/UTF-8/specials; this heuristic will mis-parse those. Fine as a best-effort fallback when `gh pr diff --name-status` succeeds (which is the normal path), but worth a note.
- [P3] `aragora/swarm/quorum_evidence.py:default_prompt_builder` — `gh pr diff` and `gh pr diff --name-status` are two separate subprocess calls; in theory the head can move between them. The subsequent `headRefOid` pin compares against the originally-resolved `head_sha`, so any mismatch still raises before the prompt is produced — race is contained, no inconsistent prompt escapes.
- [P3] `_fetch_name_status` swallows non-zero `gh` exit and OSError/SubprocessError to "" silently — intentional per the docstring (fallback path covers it), but consider a `logger.debug` for diagnosability of stuck fleets.

Security/regression posture: subprocess invocation uses argv lists (no shell), no new untrusted inputs are interpolated into shell strings, and the `default_prompt_builder` raise/return contract is preserved (tests `test_default_prompt_builder_tolerates_name_status_fetch_failure` and `test_default_prompt_builder_empty_diff_still_raises` lock this in).

dogfood: yes

scarmani:

Head: 1fa2c1b (1fa2c1bb9359a171e1c5161742de3046131787b0), committed 2026-06-14T14:57:39Z.
PR: #8417.
Model family: grok

Verdict: PASS

- No blocking issues found. Changes are narrowly scoped, well-tested, and correctly address truncation ordering risks without altering error/return semantics or introducing new attack surface.

dogfood: yes

PR #8426, opened 2026-06-15
perf(swarm): concurrent quorum reviewers + App-token read routing

an0mium:
## Problem

Two issues jointly produced the recurring **"`collect_quorum_evidence` hangs / GraphQL starvation"** failure mode seen across the fleet:

1. **Reviewers ran serially.** `quorum_evidence.collect_evidence` looped over families one at a time, each `reviewer_runner` call blocking up to its own 300s timeout. Two reviewers made wall-time the **sum** of timeouts (~10 min) — long enough that the agentic loop interrupted before any timeout fired, *presenting as a hang* even though the timeout machinery was working.
2. **Reads drained the operator's per-user PAT quota.** `merge_quorum_io`'s `gh` calls shelled out with inherited (operator PAT) auth, so the high-traffic read path (PR view / tier / check-runs / open-PR list / merge-packet) competed for the same exhausted per-user GraphQL bucket as everything else.

## Change

**1. Concurrent reviewers** (`quorum_evidence.py`)
- Resolve the ordered, de-duplicated family list up front, run the supported reviewers concurrently in a bounded `ThreadPoolExecutor` (`_MAX_REVIEWER_WORKERS=4`), then process results in the caller's requested order.
- Each runner is *already* timeout-guarded and process-isolated (claude via subprocess; API agents via spawned process), so threads here only wait — turning serial wall-time (sum of timeouts) into roughly the slowest single reviewer.
- **Behavior-preserving:** item/failure ordering, dedup, unsupported-family handling, never-fabricate, and the dissent/quorum gating are all unchanged. A reviewer that *raises* is recorded as a failure instead of aborting the run.

**2. App-token read routing** (`merge_quorum_io.py`)
- Read-only `gh` calls (`fetch_pr_context`, `fetch_latest_quorum_run`, `fetch_human_settlement_present`, `list_open_prs`, and the merge-packet subprocess) now route through the GitHub App installation token via `github_app_auth.github_cli_env`, drawing on the App's **separate** API budget.
- **Read-only by design:** `run()`'s default (PAT) env is left untouched so write call sites are unaffected — the App installation 403s on classic branch-protection writes. Degrades transparently to ambient/operator auth when no App config is present (same silent-safe contract as `scripts/gh_app_env.py`, extending the #8407 pattern from shell wrappers into the Python read path).

## Validation
- `pytest tests/swarm/test_quorum_evidence.py tests/swarm/test_merge_quorum_io.py` → **87 passed** (69+3 and 12+3 new).
- New reviewer tests: deterministic barrier-based concurrency proof, completion-order vs requested-order preservation, raising-reviewer-recorded-as-failure.
- New routing tests: read-env prefers the App token, degrades to ambient auth without App config, and `fetch_pr_context` actually routes its `gh` calls through the App token.
- `ruff check` + `ruff format --check` clean on all four files.

## Tier
Touches `aragora/swarm/quorum_evidence.py` and `aragora/swarm/merge_quorum_io.py` — swarm/merge-gate-adjacent but **behavior-preserving for the evidence output** (same comments, same order, same lint results; only faster and on a different token bucket). No parser or gate semantics change.

---
_Generated by [Claude Code](https://claude.ai/code/session_018jfJj5gb9VoLs6VBMnzhrP)_

### DIFF
--- a/aragora/swarm/merge_quorum_io.py
+++ b/aragora/swarm/merge_quorum_io.py
@@ -17,6 +17,7 @@ import sys
 import tempfile
 from typing import Any
 
+from aragora.swarm import github_app_auth
 from aragora.swarm.merge_quorum_reconcile import (
     EvidenceComment,
     PacketClassification,
@@ -68,6 +69,19 @@ def aragora_env() -> dict[str, str]:
     return env
 
 
+def _read_env() -> dict[str, str]:
+    """Environment for *read-only* ``gh`` calls.
+
+    Prefer the GitHub App installation token so reads draw on the App's separate
+    API budget instead of starving the operator's shared per-user PAT quota (the
+    chronic GraphQL-exhaustion failure mode). Degrades transparently to the
+    ambient/operator auth when no App config is present. Read-only by design --
+    the App installation 403s on classic branch-protection writes, so write call
+    sites deliberately keep ``run()``'s default (PAT) env instead of this helper.
+    """
+    return github_app_auth.github_cli_env(aragora_env())
+
+
 def run(
     args: list[str], *, env: dict[str, str] | None = None, timeout: int | None = _GH_TIMEOUT
 ) -> subprocess.CompletedProcess:
@@ -105,7 +119,8 @@ def list_open_prs(repo: str, *, limit: int, author: str | None) -> list[int]:
                 str(limit),
                 "--json",
                 "number,isDraft,author",
-            ]
+            ],
+            env=_read_env(),
         )
         or []
     )
@@ -133,7 +148,8 @@ def fetch_pr_context(repo: str, pr: int) -> dict[str, Any]:
             repo,
             "--json",
             "headRefOid,commits,statusCheckRollup",
-        ]
+        ],
+        env=_read_env(),
     )
     head_sha = str(data.get("headRefOid") or "").strip()
     commits = data.get("commits") or []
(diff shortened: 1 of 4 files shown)


scarmani:

Head: 906df81 (906df81d96c93ad648e2ab363ac05d24772dfdc2), committed 2026-06-15T03:05:03Z.
PR: #8426.
Model family: claude

Verdict: PASS

- No blocking issues. The concurrency rewrite preserves caller-requested family order via the separate `ordered_families` loop, supported/unsupported branching keeps `reviews[family]` lookups safe, and per-future exception capture prevents one bad reviewer from aborting the run. `ThreadPoolExecutor` is `with`-scoped, so shutdown is guaranteed. The App-token routing is correctly limited to read paths; write call sites are intentionally left on the default PAT env (comment explains the branch-protection 403 reason). `fetch_merge_packet_classification` switching from `aragora_env()` → `_read_env()` is read-only and safe.

Nits (non-blocking):
- [P3] `tests/swarm/test_quorum_evidence.py:test_collect_preserves_family_order_despite_completion_order` uses `time.sleep(0.2)` to force completion order — order assertion is correct but slightly timing-dependent on loaded CI; consider an `Event` or counter to make it deterministic.
- [P3] `aragora/swarm/quorum_evidence.py` concurrent `reviewer_runner` calls assume per-call thread safety; comment justifies via subprocess isolation, but any shared cwd/tmp/env mutation inside a runner would now race. Worth a one-line assert in docs of the runner contract if not already stated.
- [P3] `quorum_evidence.py` exception capture uses `str(exc)[:200]` — fine for internal `ReviewerResult.error`, not user-facing, but be aware truncation can split UTF-8 mid-codepoint in extreme cases.

dogfood: yes

scarmani:

Head: 906df81 (906df81d96c93ad648e2ab363ac05d24772dfdc2), committed 2026-06-15T03:05:03Z.
PR: #8426.
Model family: openai

Verdict: PASS

- [P3] `aragora/swarm/merge_quorum_io.py:69` `_read_env()` switches reads to GitHub App auth but does not retry ambient/PAT auth if the App token is minted yet lacks a needed read permission. This can turn previously working `gh pr view` / Actions / statuses reads into hard failures in partially-permissioned App installs. No blocking correctness issue if the App is guaranteed to have all required read scopes.
- No blocking issues found in the concurrent reviewer collection path; ordering and per-reviewer failure isolation are preserved.

dogfood: yes

PR #8466, opened 2026-06-16
feat(nomic): relay-with-timeout decision core for native mission (PR #3 core)

scarmani:
**PR #3 (pure core) of the native mission orchestrator** (plan: [#8463](https://github.com/synaptent/aragora/pull/8463)). The piece that ends the 3-minute relay: a blocked item **parks and notifies** instead of halting the whole mission.

- **`aragora/nomic/mission_relay.py`**: `RelayPolicy` (`max_item_failures`, `item_timeout_seconds`, `relay_channel`) + `RelayContext` + `decide_relay_action() → RelayDecision`.
  - Global **budget/time** limits → `STOP_MISSION`.
  - Per-item **needs-human / repeated-failure / timeout** → `PARK_AND_NOTIFY` — **mission continues other items**; operator gets an async notification, no blocking wait.
  - Else → `CONTINUE`.
- **Hard invariant (tested):** the action vocabulary is exactly `{continue, park_and_notify, stop_mission}` — **no outcome can merge / settle / mark-ready.** Parking sets an item aside for the operator; the merge-quorum gate stays the sole authority. `STOP_MISSION` is reserved for global limits, never a single blocked item.

**Pure, deterministic, UNWIRED.** Does **not** touch `boss_loop.py` — that drives your live loop (running right now) and per the Factory-mission precedent its edits need a quiet window. The wiring (this core + the actual Slack/email send into the loop) is a separate quiet-window PR.

**Tests:** 13 (continue/park/stop paths, precedence, channel gating, policy validation, the no-merge invariant). mypy / ruff / typecheck-changed clean. Branches independently from main.

Companion to PR #1 contracts ([#8465](https://github.com/synaptent/aragora/pull/8465)). Next: `aragora mission` CLI (#4, stacks on #8465) + durable resume (#5); boss_loop wiring deferred to a quiet window.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8486, opened 2026-06-16
feat(nomic): relay executor — park-and-notify (native mission PR#3 executor)

scarmani:
## What

PR #3 (executor half) of the native mission orchestrator (`docs/plans/2026-06-16-native-mission-orchestrator.md`). The pure relay decision core (`mission_relay.decide_relay_action`, #8466) only *decides*; this adds the deferred I/O half that turns a decision into real effects.

- **`MissionStore.set_item_status(mission_id, item_id, status) -> bool`** — update one work item in place (`dataclasses.replace` on the frozen `WorkItem`) and re-save atomically via the existing single write path. `False` if the mission/item is absent. Needed by the executor + future resume/status.
- **`MissionRelay`** (`aragora/nomic/mission_relay_executor.py`) — `evaluate()` delegates to the core; `apply()` turns a `RelayDecision` into effects: **PARK_AND_NOTIFY** marks the item `PARKED` + notifies (when a channel is configured), **STOP_MISSION** notifies a wind-down (does *not* park the trigger — it's a global limit, not an item fault), **CONTINUE** is a no-op. The notifier is duck-typed (`NotificationService.notify`); channel from `policy.relay_channel`.

## Invariant (test-pinned)

The executor **never merges, settles, or marks-ready** anything — the only state change is `WorkItemStatus -> PARKED` plus an advisory notification. The merge-quorum gate stays the sole settlement authority; a parked item waits for the operator. `test_invariant_executor_never_merges_or_settles` enforces this at the source level.

## Safety

- **Zero live-loop edits.** The `boss_loop.py` call site (`evaluate_and_apply` at the per-item halt points, behind `enable_native_mission`) lands separately in a quiet window.
- Default-OFF posture preserved: nothing wires this into a running loop yet.

## Tests / gates

11 new tests (`tests/nomic/test_mission_relay_executor.py`) + 21 existing mission tests green; mypy clean (2 files); ruff clean; import smoke OK.

### DIFF
--- a/aragora/nomic/mission.py
+++ b/aragora/nomic/mission.py
@@ -25,7 +25,7 @@ import json
 import os
 import tempfile
 from collections.abc import Iterable
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from enum import Enum
 from pathlib import Path
 from typing import Any
@@ -232,6 +232,30 @@ class MissionStore:
     def list_missions(self) -> list[str]:
         """List all active mission IDs based on files in the state directory."""
         return [p.stem for p in sorted(self.state_dir.glob("*.json"))]
+
+    def set_item_status(self, mission_id: str, item_id: str, status: WorkItemStatus) -> bool:
+        """Update one work item's status in place and persist.
+
+        Returns True if the item was found and updated, False if the mission or
+        item does not exist. The whole mission is re-saved atomically (the only
+        write path), so a single item update can never corrupt the file.
+        """
+        loaded = self.load_mission(mission_id)
+        if loaded is None:
+            return False
+        spec, items = loaded
+        updated: list[WorkItem] = []
+        found = False
+        for item in items:
+            if item.item_id == item_id:
+                updated.append(replace(item, status=status))
+                found = True
+            else:
+                updated.append(item)
+        if not found:
+            return False
+        self.save_mission(spec, updated)
+        return True
 
 
 class NativeMissionRunner:
--- a/aragora/nomic/mission.py
+++ b/aragora/nomic/mission.py
@@ -239,6 +239,11 @@ class MissionStore:
         Returns True if the item was found and updated, False if the mission or
         item does not exist. The whole mission is re-saved atomically (the only
         write path), so a single item update can never corrupt the file.
+
+        Concurrency: load-modify-save with no lock — assumes a single writer per
+        mission (the single-tick boss_loop / NativeMissionRunner). If a mission
+        store is ever shared across concurrent writers, this needs a flock/CAS
+        (mirroring billing/budget_guard) to avoid lost updates.
         """
         loaded = self.load_mission(mission_id)
         if loaded is None:
--- a/aragora/nomic/mission_relay_executor.py
+++ b/aragora/nomic/mission_relay_executor.py
@@ -1,0 +1,125 @@
+"""Relay executor — the I/O half the pure decision core deferred.
+
+`mission_relay.decide_relay_action` is pure: it only *decides* CONTINUE /
+PARK_AND_NOTIFY / STOP_MISSION. This module turns that decision into real
+effects: PARK_AND_NOTIFY marks the work item ``PARKED`` in the ``MissionStore``
+and (when a relay channel is configured) sends an out-of-band notification;
+STOP_MISSION sends a wind-down notification; CONTINUE is a no-op.
+
+Hard invariant (inherited from the core): NOTHING here merges, settles, or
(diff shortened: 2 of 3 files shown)


scarmani:

Head: a8dc672 (a8dc672debe37d38320c42cb01d53c190a67c6be), committed 2026-06-16T19:41:48Z.
PR: #8486.
Model family: claude

Verdict: PASS

No blocking issues. Three lower-severity observations:

- [P3] `MissionStore.set_item_status` (aragora/nomic/mission.py:236-262) is load‑modify‑save with no lock. The docstring acknowledges the single‑writer assumption, but the surrounding repo already standardises on flock/CAS for billing — worth a TODO pointing at the future relay/boss_loop wiring so the assumption isn't quietly violated when callers fan out.
- [P3] `STOP_MISSION` in `MissionRelay.apply` (aragora/nomic/mission_relay_executor.py:96-103) only sends a notification and returns `stopped=True`; nothing in the store records that the mission is winding down. If a caller forgets to honour `outcome.stopped`, the mission keeps running silently. Since the boss_loop wiring lands later, please ensure that PR also adds a regression test that drops the `stopped` signal and asserts the mission halts.
- [P3] `test_invariant_executor_never_merges_or_settles` (tests/nomic/test_mission_relay_executor.py:178-186) is a substring grep on the source. It misses indirect merges (e.g. `subprocess.run(["gh","pr","merge", ...])`, imports of merge helpers). Useful smoke test, but don't lean on it as the actual safety boundary — the real guarantee is "no mutation API other than `set_item_status(... PARKED)` is reachable from this module."

Minor nit: `severity="warning"|"error"` is a bare string — fine if `Notification` accepts it (test exercises both paths via `FakeNotifier`, so the real constructor is not validated). Worth a one‑line integration smoke against the real `NotificationService` before the boss_loop wiring lands.

dogfood: yes

scarmani:

Head: a8dc672 (a8dc672debe37d38320c42cb01d53c190a67c6be), committed 2026-06-16T19:41:48Z.
PR: #8486.
Model family: grok

Reviewing the PR diff against surrounding mission/relay code for correctness, security, and regression risks.
Verdict: PASS

No blocking [P1] issues. The executor honors the stated invariant (park + optional notify only; no merge/settle), park-before-notify ordering is correct, and the “skip notify when park didn’t take” guard is the right failure mode.

- **[P2]** `aragora/nomic/mission_relay_executor.py` (`MissionRelay.apply` / `_maybe_notify`): Re-applying `PARK_AND_NOTIFY` on an already-`PARKED` item still sets `parked=True` and re-notifies (`set_item_status` returns `True` even when status is unchanged). Once boss_loop wires this at per-item halt points, a loop that re-evaluates parked items will spam operators unless the executor checks prior status or `set_item_status` reports “no change.”
- **[P2]** `aragora/nomic/mission_relay_executor.py` (`evaluate_and_apply`): `item_id` param and `ctx.item_id` are not cross-checked; `evaluate(ctx)` can reflect one item while `apply` parks/notifies another. Add an assert or derive the ID solely from `ctx`.
- **[P2]** `aragora/nomic/mission.py` (`MissionStore.set_item_status`): Invalid `mission_id` raises `ValueError` from `path_for`, contradicting the “returns False if mission does not exist” contract and risking an uncaught crash in callers; missing missions and invalid IDs should behave consistently.
- **[P2]** `aragora/nomic/mission.py` (`set_item_status`): No transition guard — can overwrite `DONE`/`RUNNING` → `PARKED` if relay fires at the wrong lifecycle point, silently regressing queue state.
- **[P2]** `aragora/nomic/mission_relay_executor.py` (`_maybe_notify`): Sets `notified=True` after `await notifier.notify(...)` without inspecting delivery results; a failed/circuit-open send still reports success to the caller.
- **[P3]** `aragora/nomic/mission.py` (`set_item_status` docstring): Load-modify-save without locking is acknowledged, but concurrent writers can still lose updates; acceptable only while single-writer assumption holds.
- **[P3]** `aragora/nomic/mission_relay_executor.py`: `STOP_MISSION` sets `stopped=True` in memory only — no mission-level persisted stop flag; safe only if the future boss_loop caller reliably honors `RelayOutcome.stopped`.
- **[P3]** Integration gap: `MissionSpec.relay` exists in `mission.py` but the executor reads `RelayPolicy.relay_channel` only; wiring must map spec → policy or notifications will silently drop.
- **[P3]** `tests/nomic/test_mission_relay_executor.py` (`test_invariant_executor_never_merges_or_settles`): Source-text substring guard is brittle cosmetic coverage; doesn’t substitute for behavioral/integration tests once wired.

dogfood: yes

PR #8487, opened 2026-06-16
harden(quorum): retry reviewers only on infra failure, never a verdict (Tier-4, preapproved)

scarmani:
## What (Tier-4, operator-preapproved 2026-06-16)

The merge gate's own **reviewer transport** is the dominant conveyor friction observed this session: the CLI reviewers intermittently **time out** (gemini, 300s) or **error** (grok, `unknown`), which is recorded as a failed/non-counting family and forces a **manual quorum re-roll** — even when the code is fine and another family already returned PASS.

This wraps each reviewer dispatch in `_run_reviewer_with_infra_retry`: re-run **only** while the result is an **infra failure** (`ok=False`: timeout / CLI-not-found / nonzero-exit / empty). A result that **returned a verdict** (`ok=True`) — PASS *or* `changes_requested` — is returned immediately and **never retried**, so a genuine adversarial dissent can never be "retried away." Default 1 retry; env-tunable via `ARAGORA_COLLECT_EVIDENCE_INFRA_RETRIES` (0 disables).

## Not a gate weakening

Counting/settlement rules are **untouched**: `FAMILY_PROVIDERS`, the 2-distinct-family minimum, Fusion-exclusion, and `settle_one_pr` authority are all unchanged. The retry only distinguishes *transport flakiness* (`ok=False`) from a *real review* (`ok=True`), and only ever re-runs the former.

## Why Tier-4

`aragora/swarm/quorum_evidence.py` is a merge-authority surface (it composes/validates the evidence the gate counts). Per `docs/REVIEW_AUTHORITY_PRINCIPLES.md` this is a merge-authority self-modification → operator preapproval (granted 2026-06-16) + head-bound operator settlement to merge.

## Tests / gates

5 new tests pin the behavior (retry-recovers-transient, **never-retry-real-verdict**, exhaust-returns-failure, zero-disables, env-count); 117 quorum_evidence tests green; ruff + mypy clean.

### DIFF
--- a/aragora/swarm/quorum_evidence.py
+++ b/aragora/swarm/quorum_evidence.py
@@ -121,6 +121,48 @@ _REVIEWER_CLEANUP_TIMEOUT = 10
 # (e.g. 2x300s). Run them concurrently instead; this cap bounds the fan-out.
 _MAX_REVIEWER_WORKERS = 4
 
+# Retry a reviewer ONLY on infra failure (ok=False: timeout / CLI-not-found /
+# nonzero-exit / empty output) — transport flakiness, not a review. A returned
+# verdict (ok=True), INCLUDING changes_requested, is never retried: that is a
+# real adversarial review and must stand. This does not touch counting rules
+# (FAMILY_PROVIDERS, the 2-distinct-family minimum, Fusion-exclusion) — it only
+# stops a transient CLI timeout from masquerading as a missing/ dissenting family
+# and forcing a manual re-roll. Default 1 retry; 0 disables (env-overridable).
+_REVIEWER_INFRA_RETRIES_ENV = "ARAGORA_COLLECT_EVIDENCE_INFRA_RETRIES"
+_REVIEWER_INFRA_RETRIES_DEFAULT = 1
+
+
+def _reviewer_infra_retries() -> int:
+    raw = os.environ.get(_REVIEWER_INFRA_RETRIES_ENV, "").strip()
+    if not raw:
+        return _REVIEWER_INFRA_RETRIES_DEFAULT
+    try:
+        return max(0, int(raw))
+    except ValueError:
+        return _REVIEWER_INFRA_RETRIES_DEFAULT
+
+
+def _run_reviewer_with_infra_retry(
+    runner: Callable[[str, str], ReviewerResult],
+    family: str,
+    prompt: str,
+    *,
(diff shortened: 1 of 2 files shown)


scarmani:

Head: 287030f (287030fbe90341bcc7bdcbcd746a314b76de34e3), committed 2026-06-16T20:41:59Z.
PR: #8487.
Model family: claude

Verdict: PASS

- No blocking issues. The retry is correctly gated on `not result.ok` so a real verdict (including `changes_requested`) is never re-rolled — counting/family rules are untouched, and the existing fan-out wrapping remains intact.
- [P3] aragora/swarm/quorum_evidence.py:163 — retries are immediate with no backoff/jitter. If the underlying failure is rate-limiting rather than pure transport flake, back-to-back retries can amplify; default of 1 keeps this minor, but a small fixed sleep would be cheap insurance.
- [P3] aragora/swarm/quorum_evidence.py:155 — worst-case wall time per family now doubles on persistent CLI/timeout failure (e.g., 2×300s), inside the 4-worker pool. Acceptable given the explicit tradeoff in the docstring, but worth flagging since `collect_evidence` callers may have tighter budgets.
- [P3] aragora/swarm/quorum_evidence.py:147 — `_reviewer_infra_retries()` is re-read per family submission; harmless but a one-time read in `collect_evidence` would be slightly cleaner and avoid mid-run env mutation surprising one family but not another.
- [P3] tests/swarm/test_quorum_evidence.py:2117 — relies on positional `ReviewerResult("grok", "", False, "timeout")`; if the dataclass field order ever changes this breaks silently. Keyword args (`text=`, `ok=`, `error=`) would be more robust, but not required.
- Verified: env parser clamps to `max(0, …)` and falls back on `ValueError`; zero disables retries; concurrency cap (`_MAX_REVIEWER_WORKERS=4`) and Fusion-exclusion are unaffected; no new security surface.

dogfood: yes

scarmani:

Head: 287030f (287030fbe90341bcc7bdcbcd746a314b76de34e3), committed 2026-06-16T20:41:59Z.
PR: #8487.
Model family: openai

Verdict: PASS

- No blocking issues found.
- [P3] `tests/swarm/test_quorum_evidence.py`: test-only import of private helper is acceptable here, but it does couple tests to internal implementation rather than validating through `collect_evidence`. Not a regression blocker.

dogfood: yes

PR #8498, opened 2026-06-17
harden(proof): consolidate proof/benchmark-truth input-validation guards (harvest of #8052/#8057/#8064/#7995/#8044)

scarmani:
## What
Consolidates the genuinely-valuable proof-integrity / benchmark-truth input-validation guards harvested from 5 stale single-commit drafts that were closed during the open-PR backlog drain (branches preserved). One clean PR instead of 5 stale ones.

Harvested commits (fail-closed on malformed/inconsistent inputs to proof & benchmark-truth tooling):
- `measure_b0_scorecard.py` — fail closed on malformed scorecard metrics (#8052)
- `capability_gap_report.py` — fail closed on invalid capability gap sources (#8057)
- `capability_matrix_delta.py` — fail closed on missing capability matrix base (#8064)
- `render_benchmark_truth_status.py` — reject inconsistent proxy counts (#7995) + validate status corpus identity (#8044)

## Why
These guard the proof-first / benchmark-truth surfaces against silently accepting malformed input — real robustness on surfaces the operator cares about. Drained the 5-PR noise; harvested the value into `main`.

## Tests
52 passing across the 4 affected `tests/scripts/` modules. (The pre-existing `agent_bridge_steering` + `capability_matrix` CLI-count failures on main are unrelated and not touched here.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8499, opened 2026-06-17
feat(quorum): normalize reviewer output so low-cost models reliably count

scarmani:
## Problem
Low-cost models (deepseek, qwen, kimi, grok) produce useful review content but don't reliably emit the requested format — reasoning traces (`<think>…</think>`), prose preamble, heavy markdown. That raw text broke identity/verdict parsing, so their evidence failed to **count** even on a clear PASS. Observed live: qwen3-235b-a22b-*thinking* PASSed #8498 but didn't count, forcing a manual evidence post. This is *the* bottleneck on autonomous settling / the whole PR-drain.

## Fix
`normalize_reviewer_output()` canonicalizes every reviewer's text **before** the evidence comment is composed — decoupling reviewer *capability* from format *reliability*:
1. **strip reasoning-trace blocks** (`<think>/<reasoning>/…`)
2. **re-anchor at the first verdict line** (drop preamble; findings preserved) — handles the overwhelming majority deterministically
3. **opt-in cheap-reliable-model fallback** (`ARAGORA_REVIEWER_NORMALIZER_MODEL`, e.g. a haiku/mini slug) for genuinely-malformed output — faithful, default-OFF, best-effort

Deterministic-first: zero cost, no new dependency, no new failure mode. Wired into `compose_evidence_comment`.

## Security bonus
The re-anchor also makes the identity-hijack defense **stronger**: a pre-verdict `Model family: X` injection is now *dropped* entirely (not just quoted). Post-verdict injection is still quoted by the neutralizer. Both covered by the updated hijack test.

## Tests
149 passing, incl. a real-parser integration test proving a thinking-trace-polluted qwen-style PASS now counts.

## Dogfood note
This PR's own evidence is collected with the normalizer active — it settles itself: the 2nd family's raw output is normalized → counts.

### DIFF
--- a/aragora/swarm/quorum_evidence.py
+++ b/aragora/swarm/quorum_evidence.py
@@ -463,6 +463,95 @@ def _reviewer_verdict(text: str) -> str:
     return "unknown"
 
 
+# ---------------------------------------------------------------------------
+# Reviewer-output normalization
+#
+# Low-cost models (deepseek, qwen, kimi, grok) produce useful review *content*
+# but do not reliably emit the requested *format*: they wrap the verdict in
+# reasoning traces (``<think>...</think>``), lead with prose preamble, or bury it
+# under heavy markdown. That raw text then breaks identity/verdict parsing, so the
+# evidence fails to COUNT even when the model substantively passed (observed live
+# with qwen3-*-thinking). We normalize every reviewer's output to canonical form
+# BEFORE composing the evidence comment — decoupling reviewer *capability* from
+# format *reliability*. Deterministic-first (zero cost, no new dependency); an
+# opt-in cheap-reliable-model fallback handles the rare genuinely-malformed case.
+# ---------------------------------------------------------------------------
+
+_THINKING_BLOCK_RE = re.compile(
+    r"<\s*(think|thinking|reasoning|thought|scratchpad|analysis)\s*>.*?<\s*/\s*\1\s*>",
+    re.DOTALL | re.IGNORECASE,
+)
+
+
+def _strip_thinking_traces(text: str) -> str:
+    """Remove well-formed reasoning-trace blocks some models emit before answering."""
+    cleaned = _THINKING_BLOCK_RE.sub("", text)
+    return re.sub(r"\n{3,}", "\n\n", cleaned).strip()
+
+
+def _reanchor_at_verdict(text: str) -> str:
+    """Return text from the first verdict line onward (drops pre-verdict preamble).
+
+    Findings conventionally follow the verdict, so they are preserved; only leading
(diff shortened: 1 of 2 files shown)


scarmani:

Head: 1bc5605 (1bc5605a0e5612238397cd42db734052bc07923a), committed 2026-06-17T16:25:57Z.
PR: #8499.
Model family: claude

Verdict: PASS

- [P3] `aragora/swarm/quorum_evidence.py:570` — `normalize_reviewer_output(text, *, family: str = "")` accepts `family` but never uses it; the caller in `compose_evidence_comment` passes `family=family` for no effect. Either wire it in (e.g., to skip normalization for trusted families) or drop the parameter to avoid dead surface.
- [P3] `aragora/swarm/quorum_evidence.py:484` — `_THINKING_BLOCK_RE` only matches well-formed `<tag>…</tag>` pairs; an unclosed `<think>` (some streaming/truncated outputs do this) leaves the entire preamble in place. The re-anchor then rescues it iff a `Verdict:` line follows; otherwise the LLM fallback eats it. Acceptable, but worth a one-line comment so a future reader doesn't "fix" the regex to be greedy and accidentally swallow post-verdict content.
- [P3] `tests/swarm/test_quorum_evidence.py:179` — the hijack test now asserts pre-verdict content is *dropped* rather than quoted. Net behavior is safer (no identity confusion), but the audit trail of what the reviewer literally emitted before the verdict line is silently lost from the comment. If that visibility was load-bearing for forensics, capture it in a separate `[raw-pre-verdict]` quoted footer. Not a regression vs. the stated PR goal — just a tradeoff worth acknowledging.
- No P1/P2 issues. Fence + nonce + reliable-family pin on the LLM fallback are sound; deterministic path stays zero-cost; existing parser contract is unchanged; tests cover the new hijack split (pre = drop, post = quote) and the off-by-default fallback.

dogfood: yes

scarmani:
## Qwen independent model review

Reviewer: qwen (qwen) — independent adversarial model review via OpenRouter API fallback harness, grounded on the exact PR head.
Head: 1bc5605 (1bc5605a0e5612238397cd42db734052bc07923a), committed 2026-06-17T16:25:57Z.
PR: #8499.
Model family: qwen

Verdict: PASS
* No blocking issues found. Normalization logic correctly handles format inconsistencies from low-cost models without introducing security or reliability regressions.

dogfood: yes

scarmani:

PR: #8499
Exact head: 1bc5605a0e5612238397cd42db734052bc07923a
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #8499 is non-draft and live exact-head checks/merge-packet remain otherwise green.

PR #8503, opened 2026-06-17
feat(swarm): unattended Tier 0-2 auto-merge on green quorum

scarmani:
## What & why

Adds an **unattended path that merges PRs the merge-quorum gate has *already* authorized** for Tier 0-2 settlement — without a human typing `gh pr merge`. This takes the operator out of the per-PR settlement loop for low-tier PRs, which is the throughput unlock: more review lanes only out-throughput the backlog once *merging* is unattended (otherwise lanes just grow WIP behind a serialized human gate).

It makes **no new risk judgment** — it executes a merge the gate already labels *"Model quorum satisfied; Tier 0-2 settlement authorized."* **Tier 3-4 (human risk settlement) PRs are never touched** and continue through `scripts/settle_tier4_pr.py` unchanged.

## How it's safe

The decision core re-checks, defense-in-depth, the exact conditions that already gate an admin squash today:
- merge-packet `status=satisfied`, `verdict=admin_squash_allowed`, `admin_squash_allowed=true`, `requires_human_risk_settlement=false`, `unresolved_dissent=false`, `tier ≤ 2` (mirrors `auto_merge_bucket_a._merge_packet_allows_blocked_state`)
- `aragora-merge-quorum` green **and** every required check green **and** `mergeable` (mirrors `boss_drain_pass._proxy_authorized`)
- **dry-run by default**; `--apply` required; **`--match-head-commit`** so a push between decision and merge aborts rather than squashing an unreviewed tree; `--max-merges` bounds a pass.

## Files
- `aragora/swarm/auto_merge_green.py` — pure decision core + injectable apply loop
- `scripts/auto_merge_quorum_green.py` — thin `gh` I/O CLI (cheap pre-filter gates the merge-packet subprocess)
- `tests/swarm/test_auto_merge_green*.py` — 33 tests (decision core + dependency-injected merge fn)

## Verification
- 33 tests pass; ruff + mypy clean.
- Live dry-run against 5 open PRs: 0 merged, all correctly skipped with specific blockers.
- Live packet→context→decision chain validated against a real PR (every merge-packet field name confirmed).

## Dogfood
This PR touches only `aragora/swarm/`, `scripts/`, `tests/` — all Tier ≤ 2 — so it lands through the very Tier 0-2 path it creates.

## Usage
```bash
# dry-run plan
python scripts/auto_merge_quorum_green.py --repo synaptent/aragora
# act, bounded
python scripts/auto_merge_quorum_green.py --repo synaptent/aragora --apply --max-merges 5
[...]

PR #8506, opened 2026-06-17
harden(swarm): auto-merge-green defense-in-depth (packet/PR cross-check + schema-lock test)

scarmani:
Follow-up to #8503 (the unattended Tier 0-2 auto-merge unlock), which merged before two defense-in-depth findings from its own adversarial review could be folded in. This lands exactly those two:

- **packet/PR-number cross-check** in `decide_auto_merge` — re-asserts the merge-packet entry actually belongs to this PR (symmetric with the existing head-SHA cross-check). The caller `fetch_packet_entry` already filters by `pr_number`, so this is pure defense-in-depth — but for a tool that `--admin`-merges to main, binding packet flags to the wrong PR would be severe, so the guard is worth it.
- **schema-lock test** — feeds a real-shaped `review-queue merge-packet --json` payload (exact producer key names, verified against a live PR) through `context_from_gh`, so a producer field rename surfaces loudly in CI instead of silently disabling the tool (every PR would fail-closed → "nothing merges" with no signal).

No behavior change to the happy path. 47 tests (was 44), ruff + mypy clean. Tier-2 (touches only `aragora/swarm/` + `tests/`), so it settles on the normal claude+qwen quorum — no admin bypass.

### DIFF
--- a/aragora/swarm/auto_merge_green.py
+++ b/aragora/swarm/auto_merge_green.py
@@ -58,6 +58,7 @@ class PRMergeContext:
     number: int
     head_sha: str
     packet_head_sha: str
+    packet_pr_number: int
     tier: int | None
     packet_status: str
     packet_verdict: str
@@ -129,6 +130,12 @@ def decide_auto_merge(
             f"packet head mismatch (packet={ctx.packet_head_sha[:7]} view={ctx.head_sha[:7]})"
         )
 
+    # Defense-in-depth: confirm the merge-packet entry actually belongs to this
+    # PR. The caller already filters by pr_number, but binding packet flags to
+    # the wrong PR would be a severe failure, so re-assert here when disclosed.
+    if ctx.packet_pr_number and ctx.packet_pr_number != ctx.number:
+        blockers.append(f"packet PR mismatch (packet=#{ctx.packet_pr_number} view=#{ctx.number})")
+
     if ctx.mergeable != "MERGEABLE":
         blockers.append(f"not mergeable (mergeable={ctx.mergeable or 'unknown'})")
 
@@ -198,10 +205,17 @@ def context_from_gh(view: dict[str, Any], packet_entry: dict[str, Any] | None) -
     except (TypeError, ValueError):
         tier = None
 
+    pr_raw = packet.get("pr_number")
+    try:
+        packet_pr_number = int(pr_raw) if pr_raw is not None else 0
+    except (TypeError, ValueError):
+        packet_pr_number = 0
+
     return PRMergeContext(
         number=int(view.get("number") or 0),
         head_sha=str(view.get("headRefOid") or ""),
         packet_head_sha=str(packet.get("head_sha") or ""),
+        packet_pr_number=packet_pr_number,
         tier=tier,
         packet_status=str(packet.get("status") or ""),
         packet_verdict=str(packet.get("verdict") or ""),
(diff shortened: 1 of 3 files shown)


scarmani:

Head: a97f25b (a97f25b5e4aa67d4c2afe26d2f48ab051c473a13), committed 2026-06-17T20:17:13Z.
PR: #8506.
Model family: claude

**Verdict: PASS**

No blocking issues. This is a security-positive, fail-closed defense-in-depth addition, and the one regression that *would* matter (a fail-closed check that silently disables every merge because the producer doesn't emit the field) is not present — the field is reliably produced.

Verification notes:

- **Producer emits `pr_number`** — confirmed in both packet-entry builders: hydrated entries (`review_queue.py:2949`, sourced from `packet.pr_number: int`) and the explicit-merged path (`review_queue.py:3063`, `int(pr.get("number") or number)`). So `test_real_merge_packet_schema_maps_to_context`'s schema-lock claim holds, and the `<= 0 → block` guard will not fail-close legitimate Tier 0-2 merges.
- **Consumer pairs view+packet by the same PR** — `fetch_packet_entry` only returns the entry where `int(entry["pr_number"]) == pr` (`auto_merge_quorum_green.py:132`), and `view` is fetched for the same `pr`, so in the real flow `ctx.number == packet_pr_number` always. The new check is genuinely redundant-but-defensive there; it can only trip for a future/alternate caller — which is its stated purpose.
- **No other constructors** — only `context_from_gh` and the test helper build `PRMergeContext`; both use keyword args, so inserting the non-default field mid-dataclass is safe (no positional breakage).
- **Parsing/edge cases sound** — `int(pr_raw) if pr_raw is not None else 0` with `(TypeError, ValueError)` fallback to `0`; `<= 0` correctly catches both missing (`0`) and `ctx.number == 0` cases (fails closed). All blockers are accumulated, not short-circuited, so an operator still sees the full picture.

Non-blocking:

- [P3] `aragora/swarm/auto_merge_green.py:135` — the new check duplicates the invariant already enforced by `fetch_packet_entry` (`auto_merge_quorum_green.py:132`). Intentional defense-in-depth and documented as such; flagging only so future readers don't mistake it for the primary binding.

dogfood: yes

scarmani:

Head: a97f25b (a97f25b5e4aa67d4c2afe26d2f48ab051c473a13), committed 2026-06-17T20:17:13Z.
PR: #8506.
Model family: grok

Verdict: PASS

No blocking issues. The change correctly binds merge-packet authorization to the viewed PR and fails closed on missing/invalid `pr_number`, which closes a real gap where matching `head_sha` alone could authorize the wrong PR if view/packet were ever mispaired.

- [P3] `aragora/swarm/auto_merge_green.py` (`context_from_gh`) — reads only `pr_number`, not `number`/`pr` aliases used elsewhere (`_entry_pr` in `scripts/settle_one_pr.py`, `scripts/drain_codex_automation_value.py`). Safe given canonical producer schema, but slightly brittle to alternate entry shapes.
- [P3] `tests/swarm/test_auto_merge_green_cli.py` (`test_real_merge_packet_schema_maps_to_context`) — schema-lock is a static fixture, not a live `review-queue merge-packet` invocation; comment overstates runtime verification.
- [P3] `tests/swarm/test_auto_merge_green_cli.py` — no assertion that `context_from_gh(_view(), None)` surfaces the new `"packet pr number"` blocker (only `tier unknown` is tested today); coverage gap, not a logic bug.
- [P3] `scripts/auto_merge_quorum_green.py` (`fetch_packet_entry`) — production path already filters entries by `pr_number`; new guard is defense-in-depth for `context_from_gh` direct callers, not a functional fix to a known live mispairing bug. Appropriate for unattended `--admin` merge.

dogfood: yes

PR #8508, opened 2026-06-17
fix(quorum): stop "[P1] None:" non-findings from de-counting a PASS vote

scarmani:
## The bug (root-cause of "qwen PASS but doesn't count")

`has_blocking_or_negative_verdict` (`aragora/cli/commands/review_queue_comment_verdicts.py`) is the blocking-language scan used at **both** collect-time (`_lint_evidence_comment` → whether evidence is posted) **and** gate-time (`_model_review_signals_from_comments` → whether posted evidence counts toward the merge quorum). It matched **any** line starting with `[P0]`/`[P1]` and returned "blocking" **without reading the text after the tag**.

Models following the `[P1]/[P2]/[P3]` prompt format emit `"[P1] None:"`, `"[P1] N/A"`, `"[P1] no issues: ..."` to say *"no Pn findings"*. The scanner mistook these for real blockers → `negative_verdict=True` → `counted_reviewer_ids=[]` → `would_count=False` → a genuine **PASS vote silently dropped** from the quorum. This affects **every** reviewer family (qwen, claude, codex), so it can stall even a single-signal tiered gate.

## The fix (conservative — adversarial integrity preserved)

A `[P0]`/`[P1]` line is blocking **unless** the text before its first colon is **exactly** a no-finding head (`none`, `n/a`, `no issues`, `no blocking findings`, …). Matched **exactly, never as a prefix**, so a real finding that merely *starts with* "none"/"no" still blocks:
- `[P1] None of the inputs are validated` → **blocks** ✓
- `[P1] no auth check on the admin endpoint` → **blocks** ✓
- `[P1] None:` / `[P1] N/A` / `[P1] no issues: clean` → **counts** (non-finding) ✓

## Verification
- New dedicated test file (13 tests) covering both the non-finding cases and the real-finding-preservation cases.
- The **23 existing verdict tests stay green** (real `[P0]`/`[P1]` findings still block; `[P2]` still non-blocking).
- ruff + mypy clean.

Complementary to the gate-tiering and slug-default fixes in flight: tiering reduces the *number* of required signals; this makes each signal actually *count* when the reviewer PASSes.

### DIFF
--- a/aragora/cli/commands/review_queue_comment_verdicts.py
+++ b/aragora/cli/commands/review_queue_comment_verdicts.py
@@ -4,6 +4,42 @@ from __future__ import annotations
 
 import re
 
+# Explicit "no Pn finding" heads. A `[P0]`/`[P1]` line is blocking UNLESS the text
+# before its first colon is EXACTLY one of these — models emit `"[P1] None:"`,
+# `"[P1] N/A"`, `"[P1] no issues: ..."` to declare the absence of a finding. Matched
+# exactly (not as a prefix) so a real finding that merely starts with "none"/"no"
+# (e.g. "[P1] None of the inputs are validated") still blocks.
+_NO_FINDING_HEADS = frozenset(
+    {
+        "none",
+        "none found",
+        "none identified",
+        "none noted",
+        "none here",
+        "n/a",
+        "na",
+        "nil",
+        "0",
+        "zero",
+        "false",
+        "[]",
+        "not applicable",
+        "no issues",
+        "no issue",
+        "no findings",
+        "no finding",
+        "no blockers",
+        "no blocker",
+        "no blocking",
+        "no blocking findings",
+        "no concerns",
+        "no concern",
+        "no critical issues",
+        "no critical issue",
+        "no critical findings",
+    }
+)
+
 
 def has_blocking_or_negative_verdict(body: str) -> bool:
     """Return True for explicit evidence comments that report blockers."""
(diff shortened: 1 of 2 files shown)


scarmani:

Reviewer: claude (anthropic) — independent adversarial model review via OpenRouter API fallback harness, grounded on the exact PR head.
Head: 234b181 (234b181bb18305d4b3ae4103f617de9a755c81a3), committed 2026-06-17T20:14:21Z.
PR: #8508.
Model family: claude

Verdict: PASS

- [P3] `aragora/cli/commands/review_queue_comment_verdicts.py:97` — A real blocker phrased as `"[P1] No issues found: auth is bypassable"` is now silently de-counted because the head (`"no issues found"`) is whitelisted regardless of the trailing clause. This is an inherent, bounded weakening of the gate; acceptable since it trusts the model's self-tag, but worth noting as an attack/sloppiness surface.
- [P3] The `continue` after a no-finding head relies on the rest of the loop not re-examining `priority_marker_line` via the generic `label: value` path; the unshown loop body isn't in the diff, so confirm the `continue` actually skips it and a whitelisted-head line cannot re-trigger blocking through value-matching (`"None: defense ... bypass"` would be a false block). Test `test_p1_none_colon_is_not_blocking` covers a benign case but not a value containing blocking keywords.
- [P3] No test asserts the `[P2]`-tagged lines in `test_full_pass_body_with_p1_none_lines_counts` are handled by separate logic vs. ignored entirely; passes here only because `[P2]` isn't in the `p0|p1` regex. Fine, but the test name implies broader coverage than it gives.

No blocking issues. The fix is conservative, exact-match on the head, and the adversarial test cases (`"None of the inputs"`, `"no authentication"`) correctly still block.

dogfood: yes

scarmani:

Head: 234b181 (234b181bb18305d4b3ae4103f617de9a755c81a3), committed 2026-06-17T20:14:21Z.
PR: #8508.
Model family: grok

Verdict: PASS

No blocking issues. The fix correctly narrows the `[P0]`/`[P1]` fast-path to exact no-finding heads, preserves conservative blocking for real findings (`None of…`, `no auth…`, bare tags), and tests cover both the regression and adversarial cases. All 27 related tests pass.

- **[P2]** `review_queue_comment_verdicts.py` (~L100): head extraction splits only on `:`; `[P1] None — …` / `[P1] None - …` still block (verified). The label/value path already treats `—`/`–`/`-` as separators, so dash-formatted no-finding lines remain vulnerable to the same false-positive de-count.
- **[P2]** `review_queue_comment_verdicts.py` (~L100–104): any exact no-finding head with trailing text is ignored (`[P1] None: settlement gate bypass` → non-blocking). Enables quorum evasion if a reviewer pairs `Verdict: PASS` with smuggled blocker prose after the head; acceptable tradeoff for model output, but worth documenting as a known limit.
- **[P3]** `review_queue_comment_verdicts.py` (~L12–40 vs ~L63–76): `_NO_FINDING_HEADS` duplicates `non_blocking_prefixes` with different matching semantics (exact vs prefix); drift risk on future edits.
- **[P3]** `test_review_queue_comment_verdicts.py`: no coverage for dash/em-dash no-finding variants, `**[P1]** None:`, or `[P1]None:` (all work manually); existing `TestHasBlockingOrNegativeVerdict` in `test_review_queue.py` still doesn't assert the new non-blocking `[P1] None:` cases.

dogfood: yes

Issue #8257, opened 2026-06-12
[Epic] Codebase health: macro-architecture, packaging, and gate-rigor remediation


an0mium:
## Context (codebase-health audit, 2026-06-12)

Three measured assessment passes (architecture/quality, tests/CI, external utility) at head a5cf5fc70b produced a consistent verdict: **file-level code quality is genuinely good and the governance machinery (quorum, drift ratchets, receipts, evidence loops) measurably works — but that machinery was never pointed at macro-architecture, packaging, or the outsider's first hour.** Full evidence: `docs/audits/2026-06-12-codebase-health-audit.md` (+ machine-readable `docs/audits/2026-06-12-codebase-health-baseline.json`); operational sequencing: `docs/superpowers/plans/2026-06-12-codebase-health-program.md`.

Headline measurements (all commands pinned in the baseline file):
- `aragora/server/` ≈ 27% of package LOC; **180** files outside it import it; `debate/orchestrator.py:40` imports `server.metrics` at module level; **910** TYPE_CHECKING files, **155** circular-mention files.
- mypy `check_untyped_defs=false`; ruff `BLE001`/`G004` neutralized package-wide — gates lag the code.
- Packaging: 3 PyPI names, no console entry points, `INSTALL.md` promises a CLI the built package excludes.
- Docs: **925** md files, **5** self-hosting guides, quickstart pins version `2.6.3` vs `2.9.0`.
- 2 P2 security defaults from #8163's two-family review evidence (public `/metrics` without token; `session.revoke` revokes arbitrary tokens).
- CI: **86** workflows; canceller masked 2 real failures on #8220 (failed only on manual-rerun attempt 2).
- Tests verified substantive: 220K+ test functions accurate to 0.2%, skip density ~1%, retries disabled.

## Thesis

Apply the repo's own proven **shrink-only baseline / ratchet pattern** (`tests/server/test_route_collisions.py` precedent) to the three unguarded surfaces: layering, gate rigor, and the external first hour. Every child carries a falsifiable definition of done; most are measurable by re-running the baseline commands.

## Child issues

- [ ] HEALTH-1 root hygiene (#8258)
- [ ] HEALTH-2 layering shrink-only baseline (#8259)
- [ ] HEALTH-3 security defaults (/metrics fail-closed, session.revoke ownership) (#8260)
- [ ] HEALTH-4 mypy/ruff per-package ratchets (#8261)
- [ ] HEALTH-5 docs canonicalization (#8262)
[...]

PR #8512, opened 2026-06-18
feat(swarm): wip_budget — count-based open-PR WIP cap (stop the queue refilling)

scarmani:
## What

A **count-based open-PR WIP ceiling** — the missing queue-pressure control. The backlog doesn't drain because one fleet **out-generates the merge gate** (generation ≈ settlement). This adds the cap as a companion to `loop_budget.py`:

- `loop_budget.py` (existing) governs a *running* loop's **dollar spend**.
- `wip_budget.py` (new) governs whether a generation lane should create **another PR** given how many it already has open.

## Design (mirrors the proven dollar-budget pattern)

- `WipPolicy` — operator-owned per-fleet ceilings from `.aragora/wip_budgets.json` (+ `ARAGORA_WIP_OPEN_PR_CAP` fleet-default env fallback), loaded **fail-open** exactly like `BudgetPolicy`.
- `resolve_wip_budget` / `classify_wip` — pure, unit-tested decision. **Only a real ceiling + real count can produce `over_cap`** (`allow_generation=False`). An unknown/fabricated count **degrades and keeps allowing generation** — blocking a legitimate fleet on a number we can't trust is precisely the fabricated-alarm class the control plane exists to avoid (matches loop_budget's "never halt on stale spend").

## Scope / safety

**Provides the rails, not the retirement.** A generation lane adopts `classify_wip` at its create-PR decision point. `boss_loop.py` is a hot god-file (under #8257 P5) and a live-loop file, so it is **deliberately not edited here** — wiring is a documented follow-up seam. Net-new files only; no hot or merge-authority surfaces touched → Tier-2.

## Why this matters

Every fleet's own diagnosis converges here: settlement throughput is necessary but insufficient while generation refills the queue. This is the lever that lets the backlog actually trend down.

## Verification

13 unit tests (policy precedence + fail-open load, `_as_count` edge cases, all 4 resolve quadrants, classify within/over/degraded/unavailable + the fail-safe "unknown count never over_cap" contract), ruff + mypy clean.

### DIFF
--- a/aragora/swarm/wip_budget.py
+++ b/aragora/swarm/wip_budget.py
@@ -43,6 +43,10 @@ from typing import Any
 POLICY_RELPATH = Path(".aragora") / "wip_budgets.json"
 ENV_FLEET_DEFAULT = "ARAGORA_WIP_OPEN_PR_CAP"
 
+# resolve_wip_budget source_status values (human-readable; classify_wip derives
+# its verdict from the data, not from this advisory label -- see classify_wip).
+WIP_OK = "ok"
+
 # classify_wip verdicts.
 WIP_WITHIN_CAP = "within_cap"
 WIP_OVER_CAP = "over_cap"
@@ -148,7 +152,15 @@ def resolve_wip_budget(
     policy: WipPolicy,
     fleet_id: str,
 ) -> dict[str, Any]:
-    """Compose one fleet's raw ``wip`` dict for :func:`classify_wip`."""
+    """Compose one fleet's raw ``wip`` dict for :func:`classify_wip`.
+
+    Unlike :mod:`loop_budget` -- which reads a persisted spend snapshot and tracks
+    its mtime freshness -- the open-PR count here is supplied live by the caller
+    (e.g. a just-run ``gh pr list``), so there is no freshness/age field: ensuring
+    the count is current is the caller's responsibility at the decision point. A
+    caller wiring this into a generation lane must measure the count immediately
+    before gating, not reuse a stale cached value.
+    """
     cap, cap_source = policy.cap_for(fleet_id)
     count = _as_count(open_pr_count)
 
@@ -181,7 +193,7 @@ def resolve_wip_budget(
         }
     return {
         "source": f"{cap_source} + count:supplied",
-        "source_status": "ok",
+        "source_status": WIP_OK,
         "open_pr_count": count,
         "ceiling": cap,
         "remaining": cap - count,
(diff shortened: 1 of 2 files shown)


scarmani:

Head: 51a5160 (51a516059889e890159a527878a4f547dc4406cb), committed 2026-06-18T05:28:08Z.
PR: #8512.
Model family: claude

Verdict: PASS

- [P3] `aragora/swarm/wip_budget.py:48-58` (`_as_count`): inconsistent numeric-string handling — `3.0` (float) is accepted as `3`, but `"3.0"` (str) raises `ValueError` and is rejected as unknown. Minor and fails safe (rejected → `degraded`/no-gate, never a fabricated `over_cap`), so harmless in practice; worth a one-line note since env values and JSON-string caps arrive as strings.

Otherwise solid. Confirmed correct on the points that matter for an adversarial pass:
- Fail-safe invariant holds: only a real parsed count **and** a real parsed ceiling can yield `over_cap`/`allow_generation=False`; every other shape (`degraded`/`unavailable`) keeps generation allowed. `classify_wip` re-derives from data and ignores the advisory `source_status`, so a mislabeled dict can't bypass the gate (test covers this).
- `_as_count` correctly rejects `bool` before the `int` branch, plus NaN/inf/negative/non-integral — no spoofed counts.
- Boundary semantics right: `count >= ceiling` blocks, so the cap is the max in-flight PRs; `cap=0` freeze case works (`0 >= 0`).
- `WipPolicy.load` degrades on missing/unreadable/non-dict JSON (`OSError`/`JSONDecodeError`/`ValueError` covers `IsADirectoryError`, `UnicodeDecodeError`), never raising into the fleet path; per-fleet > default > env precedence is correct and handles `0` caps via membership/`is not None` rather than truthiness.
- No security surface (read-only JSON load, no eval/injection) and purely additive new files with no existing importers, so regression risk is negligible.

dogfood: yes

scarmani:

Head: 51a5160 (51a516059889e890159a527878a4f547dc4406cb), committed 2026-06-18T05:28:08Z.
PR: #8512.
Model family: grok

Verdict: PASS

- [P2] `aragora/swarm/wip_budget.py` (`classify_wip` / `resolve_wip_budget`) — Unlike `loop_budget`, there is no freshness/age signal for `open_pr_count`; any sanitized integer is fully trusted, so a stale cached `gh pr list` result can spuriously hit `over_cap` and block generation (opposite bias from unknown-count fail-open). Docstring warns callers, but the framework does not degrade stale counts the way spend staleness does.
- [P2] `aragora/swarm/wip_budget.py` (`classify_wip`) — `cap=0` freeze only blocks when count is known (`0 >= 0`); unknown count always degrades to `allow_generation=True`, so a deliberate fleet freeze is bypassable if the caller omits or fails to supply a count.
- [P3] `tests/swarm/test_wip_budget.py` — Missing `test_loop_budget.py` parity cases: malformed `ARAGORA_WIP_OPEN_PR_CAP`, corrupt policy file + env fallback, and `{}` policy file supplemented by env default.

dogfood: yes

PR #8576, opened 2026-06-23
feat(lanes): record heartbeat finalizer receipts

scarmani:
Refs #8560.

## Summary
- Add append-only heartbeat finalizer receipt support to scripts/agent_heartbeat.py.
- Add CLI --finalize mode for terminal outcomes: completed, failed, cancelled, and handoff.
- Add focused regression coverage for heartbeat renewal, finalizer receipt append behavior, shared-state CLI writes, and invalid terminal outcomes.

## Validation
- python3 -m pytest tests/scripts/test_agent_heartbeat.py -q
- python3 scripts/report_code_quality.py --check
- git diff --check

This is the first narrow slice for #8560; launcher trap/finally wiring remains a follow-up slice.

Issue #8560, opened 2026-06-23
Lane leases: heartbeat renewal and finalizer receipts for owner lifecycle


scarmani:
## Problem
Lane ownership is currently too easy to leave indefinite: a process can disappear without a terminal owner receipt, forcing later agents to infer whether the row is stale.

## Desired behavior
Every launcher and lane runner should write lease heartbeats while active and a finalizer receipt when exiting, completing, failing, cancelling, or handing off.

## Acceptance criteria
- `scripts/agent_heartbeat.py` supports a finalizer/exit receipt mode.
- Finalizer receipts include lane id, owner session, PID/cwd/worktree/branch/PR where available, outcome, timestamp, and reason.
- Launchers/runners wire start heartbeat, lease renewal, and trap/finally finalization.
- Tests prove renewal updates lease state and finalizers write terminal outcome receipts.

## Notes
This is part of the leased lane ownership program. Receipts are append-only proof; they are not deletion authority by themselves.

scarmani:
Verified complete on origin/main. PR #8610 added agent_heartbeat.py finalizer/exit receipt mode, heartbeat renewal, and tmux launcher start/renew/finalize wiring; PR #8631 tightened pidless finalizer matching. Coverage includes tests/scripts/test_agent_heartbeat.py for renewal and terminal finalizer receipts plus tests/scripts/test_tmux_transport_scripts.py for launcher heartbeat/finalizer wrapper behavior. Closing as completed.

PR #8582, opened 2026-06-23
fix(lanes): suppress finalized heartbeat fallback freshness

scarmani:
Refs #8560.

## Summary
- Prevent terminal-marked heartbeat rows from falling back to copied lane/ledger heartbeat timestamps.
- Keep terminal finalized owners from being assessed live through stale duplicated heartbeat metadata.
- Add a focused regression for fresh lane and ledger heartbeat fallbacks alongside a terminal heartbeat row.

## Validation
- python3 -m pytest tests/scripts/test_agent_heartbeat.py tests/scripts/test_identify_lane_owner.py -q
- python3 scripts/report_code_quality.py --check
- git diff --check

Follow-up to #8576; this contains only the post-merge liveness repair.

### DIFF
--- a/scripts/identify_lane_owner.py
+++ b/scripts/identify_lane_owner.py
@@ -1807,7 +1807,9 @@ def assess_owner_liveness(
         if heartbeat_terminal and heartbeat
         else ""
     )
-    if heartbeat and heartbeat.get("last_seen_at") and not heartbeat_terminal:
+    if heartbeat_terminal:
+        last_heartbeat_at = None
+    elif heartbeat and heartbeat.get("last_seen_at"):
         last_heartbeat_at = str(heartbeat["last_seen_at"])
     elif lane.get("last_heartbeat_at"):
         last_heartbeat_at = str(lane["last_heartbeat_at"])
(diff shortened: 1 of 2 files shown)


scarmani:

Head: d9ea0c3 (d9ea0c31c1e8ad2443e32441eaf4f233c03f5b21), committed 2026-06-23T21:53:34Z.
PR: #8582.
Model family: claude

Verdict: PASS

Analysis:
- **Correctness of both halves is sound.** The first hunk sets `last_heartbeat_at = None` on a terminal heartbeat, so `heartbeat_recent` (line 1849-1852) can no longer fire, and it also prevents the old fall-through to `lane["last_heartbeat_at"]` / ledger heartbeat fields. The second hunk excludes `last_heartbeat_at`/`heartbeat_at` from the lease-anchor candidate sets when terminal, so a recent heartbeat timestamp on the owner record or ledger can't keep `anchor_ts` fresh. Both halves are necessary — the test exercises both by adding recent `last_heartbeat_at` to lane and ledger.
- **No regression in the non-terminal path.** `max(_updated_at_timestamp(ledger_entry.get(key)) for key in _LEDGER_TIMESTAMP_KEYS)` is exactly equivalent to the replaced `_ledger_entry_timestamp(ledger_entry)`, and `ledger_timestamp_keys`/`owner_timestamp_keys` are unchanged unless terminal.
- **Scope is contained.** Non-heartbeat activity signals (`updated_at`, `last_delivery_at`, etc.) still legitimately keep an owner "live"; only heartbeat-derived freshness is suppressed, matching the intent. The assessment is advisory-only per the docstring, so even the behavior change carries low operational risk.
- `terminal_heartbeat_outcome`/`terminal_heartbeat_at` are captured (lines 1802-1809) before `last_heartbeat_at` is nulled, so they remain populated — test confirms.

[P3] `scripts/identify_lane_owner.py:1840-1842` — `max(... for key in ledger_timestamp_keys)` has no `default=`. With the current constants the terminal-filtered tuple always retains 5 non-heartbeat keys, so it can't be empty today, but the coupling is implicit: if `_LEDGER_TIMESTAMP_KEYS` were ever reduced to only heartbeat fields, the terminal branch would raise `ValueError: max() arg is an empty sequence`. Defensive `default=0.0` on the generator would remove the latent fragility. (Non-blocking.)

dogfood: yes

scarmani:

Head: d9ea0c3 (d9ea0c31c1e8ad2443e32441eaf4f233c03f5b21), committed 2026-06-23T21:53:34Z.
PR: #8582.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #8609, opened 2026-06-24
fix(outbox): require current base for merged-pr proof

scarmani:
## Summary
- fail closed when an outbox handoff requests a base different from the reconciler base
- require merged-PR preservation proofs to show the desired head is still present or patch-equivalent on the current base before archiving
- add regressions for reverted desired heads and head-only base mismatches

## Validation
- `python3 -m pytest tests/scripts/test_reconcile_automation_outbox.py tests/scripts/test_identify_lane_owner.py -q`
- `pre-commit run --files scripts/reconcile_automation_outbox.py tests/scripts/test_reconcile_automation_outbox.py`

### DIFF
--- a/scripts/reconcile_automation_outbox.py
+++ b/scripts/reconcile_automation_outbox.py
@@ -352,6 +352,11 @@ def _requested_base_from_payload(payload: Mapping[str, Any]) -> str:
     return ""
 
 
+def _handoff_base_matches_reconciler(payload: Mapping[str, Any], base: str) -> bool:
+    expected_base = _requested_base_from_payload(payload) or base
+    return _normalize_base_ref(expected_base) == _normalize_base_ref(base)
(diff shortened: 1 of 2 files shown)


scarmani:
Queue-drain close: freezing this PR rather than continuing quorum/evidence churn.

Reason: exact head `f4534f5e69c29154f7ad234cad9abab39e9638f1` is still blocked only on model quorum after five branch-head commits. The previous owner lane is stale and eligible for guarded stale-claim handling, there are no blocking reviews or fresh steering messages, and the queue-drain anti-treadmill rule says PRs with >=3 heads that still fail quorum should be frozen or closed rather than patched again.

No branch was deleted. Reopen or supersede only if a human owner wants to substantially narrow or replace this current-base proof change.

PR #8637, opened 2026-06-26
fix(secrets): never hang on interactive MFA prompt in headless processes

scarmani:
## Summary

**Root-cause fix for "the automations never landed anything."** `.env` sets
`ARAGORA_USE_SECRETS_MANAGER=true` + `AWS_PROFILE=aragora-secrets`. Any **headless**
process (cron automations, the overnight conductor, a background evidence pass) that
loads secrets then resolves AWS credentials through an assume-role profile requiring
**MFA** — and botocore calls `getpass()`, which **blocks forever** with no TTY to
answer it. The process looks alive but is wedged on an invisible prompt, so it can
never collect/post quorum evidence → never satisfies the gate → never merges.

Confirmed live: 3 background `collect-evidence` runs this session hung on
`Enter MFA code for ...armand-bootstrap-totp` until killed. The keys are already in
`.env` (XAI, OpenRouter) — the AWS detour was pure misconfiguration for headless use.

## Fix

In a non-interactive process, neuter botocore's assume-role MFA prompter so it
**fails fast → falls back to env/.env secrets** instead of hanging.
- `_mfa_prompt_allowed()` — interactive (TTY) or explicit `ARAGORA_AWS_ALLOW_MFA_PROMPT` only.
- `_fail_fast_mfa_prompter()` — raises a clear "no TTY" error instead of `getpass`.
- `SecretManager._build_client()` — installs the fail-fast prompter on a custom botocore session when non-interactive.

## Scope Contract

**Goals:** stop headless secret-loading from hanging on MFA; fall back to env keys.
**Non-goals:** changing the interactive path, the production (instance-role/OIDC) path, strict-mode policy, or which secrets are critical. No new secret sources.

## Test plan
- 81 secrets tests pass (5 new: TTY/override matrix, fail-fast prompter raises, guarded-session path headless, default path interactive, real-botocore prompter swap).
- mypy clean; ruff clean.

### DIFF
--- a/aragora/config/secrets.py
+++ b/aragora/config/secrets.py
@@ -438,7 +438,12 @@ class SecretManager:
                 },
             )
             client = self._build_client(boto3, region, config)
-            self._aws_clients[region] = client
+            # Don't cache a fail-closed None: a later retry (e.g. after credentials
+            # become available, or in a different process state) must be able to
+            # re-attempt Secrets Manager rather than be pinned to None for the
+            # SecretManager instance's lifetime.
+            if client is not None:
+                self._aws_clients[region] = client
             return client
         except ImportError:
             logger.debug("boto3 not installed, AWS Secrets Manager unavailable")
--- a/aragora/config/secrets.py
+++ b/aragora/config/secrets.py
@@ -36,11 +36,43 @@ from __future__ import annotations
 import json
 import logging
 import os
+import sys
 import threading
+from collections.abc import Mapping
 from dataclasses import dataclass, field
 from typing import Any, cast
 
 logger = logging.getLogger(__name__)
+
+
+def _mfa_prompt_allowed(*, isatty: bool, env: Mapping[str, str] | None = None) -> bool:
+    """Whether an interactive AWS MFA prompt is permissible in this process.
+
+    A non-interactive process (cron automation, headless conductor, a background
+    evidence pass) must NEVER block on ``getpass`` for an MFA code — it hangs
+    forever with no TTY to answer it. This is the silent failure that wedged every
+    automated Secrets Manager load: the process appears to "run" but is stuck on an
+    invisible prompt. Allowed only when attached to a TTY, or when explicitly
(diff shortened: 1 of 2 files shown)


scarmani:

Head: 1cfd5cb (1cfd5cb126d0aca53e6a25a81d13cb308cabdb9f), committed 2026-06-26T14:23:56Z.
PR: #8637.
Model family: claude

Verdict: PASS

- [P3] `aragora/config/secrets.py` `_build_client` (the `except Exception` around the guarded path): the try wraps both prompter installation *and* `boto_session.client(...)` credential resolution. In the primary intended scenario (headless + assume-role-with-MFA profile), the fail-fast prompter's `RuntimeError` is raised *during* `.client()` resolution and caught here — so the operator-facing log reads "could not install non-interactive MFA guard ... refusing the default client" even though the guard installed and fired exactly as designed. Same broad catch also masks genuine credential/network errors during assume-role in the headless path as a "guard install failure." Outcome (return None → env/.env fallback) is correct; only the log/legibility is wrong. Consider narrowing the try to the install step, or distinguishing the prompter-raise from an install failure.
- [P3] `_has_controlling_tty` (secrets.py): on Windows `/dev/tty` never opens → `OSError` → always classified non-interactive, so a Windows interactive user with an MFA profile silently loses the getpass prompt they previously got. Niche for a server codebase, and `ARAGORA_AWS_ALLOW_MFA_PROMPT` is an escape hatch, but it's a behavior change worth noting.
- [P3] Test gap: no test exercises the prompter *actually firing* → `_build_client` returns None. `test_real_botocore_prompter_is_failfast_when_headless` installs the prompter but its fake session's `.client()` never resolves credentials, and the mock test returns a mock client. The genuine-MFA-raise → None path (the core fix) is unverified.

dogfood: yes

scarmani:

Head: 1cfd5cb (1cfd5cb126d0aca53e6a25a81d13cb308cabdb9f), committed 2026-06-26T14:23:56Z.
PR: #8637.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #8667, opened 2026-06-27
feat(strategy): bounded-mission cadence M0 + ODR v1.0 contract (M1) + quorum→receipt bridge (M2 core)

scarmani:
## Strategy as a bounded-mission cadence — M0 + M1

Turns the "make Aragora useful + uniquely valuable to others" strategy into a durable, exit-gated execution structure, and delivers the first bounded sub-mission.

**Spec:** `docs/superpowers/specs/2026-06-26-strategy-as-bounded-mission-cadence-design.md`
**Plan:** `docs/superpowers/plans/2026-06-26-mission-cadence-m0-m1.md`

### M0 — durable Mission Queue
- `## Strategy Mission Queue` section in the Intake Register (single-register rule) with rows M1/M2/M3, statuses, and per-row external-proof gates.
- Parser test enforcing the at-most-one-active invariant and required columns.

### M1 — ODR v1.0 contract + independent verification (docs + verification only)
- ODR §9 expanded into a **Versioning and Stability contract**: field-stability tiers, v1.0 backward/forward-compat guarantees (with `aragora-verify` as conformance authority), deprecation policy, native↔`odr_version` independence, and explicit path-to-GA status.
- `docs/specs/odr-native-mapping.md` — normative native `DecisionReceipt` → ODR field mapping, with a **drift-guard test** (every emitted top-level field must be documented).
- Committed example receipt (`docs/specs/examples/example-decision-receipt.odr.json`) + a regeneration/schema-conformance guard.
- **External proof:** `aragora-verify` (zero Aragora dependency) independently verifies the emitter-produced receipt — `schema_conformance` PASS, `canonical_digest` PASS, unsigned signature = warning only.

### Scope discipline
- **Tier 0–1 only.** No README, `action.yml`, `.github/workflows/`, server verify endpoint, signing-in-gate, or ODR format-version bump — all parked.
- The on-wire `odr_version` stays `"0.1"`; this PR authors the v1.0 *contract* and locks the round-trip. The version flip + signing + PyPI re-release are a separate Tier-3 release.

### Validation
- `pytest tests/docs/test_mission_queue_register.py tests/gauntlet/test_odr_native_mapping.py tests/gauntlet/test_odr_example_receipt.py` → 4 passed
- `aragora-verify` suite (`PYTHONPATH=src python -m pytest tests/`) → 50 passed
- `mypy aragora/gauntlet/odr_export.py` → clean; pre-commit green

### Parked for founder/operator (not in this PR)
[...]

Issue #8665, opened 2026-06-27
[Epic] Strategy mission: DecisionReceipt Action wedge and external proof


scarmani:
## Objective
Turn the project positioning assessment into a durable, bounded execution mission: Aragora becomes the audit layer for AI decisions, starting with PR governance.

## Mission queue
1. M0: Register the strategy mission in the Roadmap Intake Register and keep one active mission home.
2. M1: Promote Open Decision Receipt from v0.1 to v1.0 GA by extending #8223 with a versioning/stability contract, native-to-ODR mapping, and independent verification proof.
3. M2: Productize the GitHub Action wedge by bridging model-quorum/CollectOutcome evidence into a DecisionReceipt artifact and preparing approval-required workflow changes.
4. M3: Publish a proof corpus and legibility pass: receipts from real PR governance plus a short external proof writeup and outsider-readable docs.

## Guardrails
- One bounded sub-mission active at a time.
- No perpetual loop: each sub-mission advances only when its proof artifact exists and verifies.
- Tier 3/4, workflow, release, and protected-surface changes park for exact-head human settlement.
- The Mission Engine/Intake Register are the canonical runtime and durable queue; do not create a parallel orchestrator or register.

## Existing trackers to extend
- ODR: #8223
- Codebase health and legibility: #8257
- Roadmap intake audit: #8650

scarmani:
M0 implementation PR opened: #8666 (`codex/strategy-mission-cadence-20260627` @ `97dc6698dc`).

Validation from the branch worktree:
- `python3 scripts/validate_doc_links.py README.md docs/status/ROADMAP_INTAKE_REGISTER.md docs/status/NEXT_STEPS_CANONICAL.md docs/superpowers/specs/2026-06-26-strategy-as-bounded-mission-cadence-design.md`
- `git diff --check`
- `python3 -m ruff check scripts/validate_doc_links.py`

Current status: M0 captured as a draft PR; M1 should extend #8223 with ODR v1.0 GA after this lands.

MarkovianProtocol:
M1 (independent verification proof) and M3 (external proof) map cleanly onto a Bitcoin anchor, and it composes with ODR rather than replacing it.

[...]

PR #8669, opened 2026-06-27
feat(action): emit verifiable Decision Receipt as a PR artifact (M2 finish)

scarmani:
## M2 finish — the Action emits a verifiable Decision Receipt

**Stacked on #8667** (M0/M1/M2-core). Review/merge #8667 first; this branch targets `worktree-strategy-mission-cadence` so the diff shows only the M2-finish changes.

Completes the M2 wedge: the marketplace action (`action.yml`) now produces a **portable, independently-verifiable Open Decision Receipt** as a PR build artifact.

### What it does (new `Emit decision receipt` step)
1. Collects a heterogeneous-model merge-quorum review — **dry run, never posts** (no `--apply`; Tier-gated regardless).
2. Bridges the outcome → `DecisionReceipt` → portable ODR (reuses the M2-core bridge from #8667).
3. Verifies it (schema + JCS digest) and uploads `decision-receipt.odr.json` as an artifact.
4. Appends a receipt summary to the PR comment with `pip install aragora-verify && aragora-verify decision-receipt.odr.json`.

Gated behind a new **`emit-receipt`** input (default `true`) and **fully failure-tolerant** (`set +e`, always `exit 0`) — a receipt problem can never break the workflow. New `receipt-reviewers` input; new `receipt-*` outputs.

### Also fixes a latent bug
`action.yml` on `main` did **not parse** as valid YAML — a column-0 `python3 - <<'PY'` heredoc broke the `run:` block scalar (both `actionlint` and PyYAML fail on it). External consumers of `uses: synaptent/aragora@v…` would hit a parse error. Extracted that logic to `scripts/extract_review_counts.py` (stdlib-only, behavior-preserving, **tested**); the action is now valid YAML.

### Tests / validation
- `tests/scripts/test_emit_pr_receipt.py` (3) + `tests/scripts/test_extract_review_counts.py` (3) → pass; bridge slice still green (11 total)
- `mypy` clean; `action.yml` now `safe_load`s and `actionlint` parses it (remaining actionlint notes are workflow-schema checks that don't apply to an action-definition file)
- Simulated the Action's emit path locally (collector-shaped outcome → emitter → verified receipt: `verdict=PASS`, `verified=true`)

### Honest caveat
I **cannot run GitHub Actions locally**, so the live CI execution path (the source-install + collector dry-run inside the runner) needs one real CI run to confirm. The step is failure-tolerant by design so a problem degrades to "no receipt" rather than a red workflow.

### Design notes
- The receipt step installs the action's **own pinned source** (`pip install "$GITHUB_ACTION_PATH"`) because the bridge is newer than the published `aragora` PyPI package. The existing `aragora review` step is untouched (still uses PyPI).
- This is the marketplace action (public surface) — flagged as the approval-required change you green-lit.


scarmani:
Part of epic #8665 (DecisionReceipt Action wedge and external proof). M2-finish for the cadence; stacked on #8667.

scarmani:
M2 reshape status after this pass:

- Parked the operator-gated self-test workflow outside the mergeable PR path; this PR no longer changes `.github/workflows/*`.
- Kept root `action.yml` as the canonical merge target: `emit-receipt` remains opt-in/default-false, receipt outputs are exposed, and the receipt path is best-effort/artifact-producing without posting quorum evidence.
- Local validation passed: `pytest tests/scripts/test_extract_review_counts.py tests/scripts/test_emit_pr_receipt.py tests/swarm/test_quorum_receipt.py -q` => 22 passed; emitter dry-run produced a verified ODR digest; action interface parse/check passed; automation preflight passed.
- Strong live CI receipt proof remains blocked until model credential/runner access is operator-settled. Treat this PR as the mergeable action wedge, not as live strong-quorum proof.

scarmani:

Head: fc8da32 (fc8da32ec911070f2d93166060eda58e69c04f81), committed 2026-06-30T16:13:17Z.
PR: #8669.
Model family: claude

Verdict: PASS

The PR is solid: it fixes a genuine YAML-validity bug (the column-0 `python3 - <<'PY'` heredoc broke the block scalar) by extracting tested stdlib scripts, adds input validation (PR number `^[0-9]+$`, numeric max-diff-size, reviewers allowlist regex), guards GitHub-output newline injection, and makes receipt emission opt-in, backwards-compatible (default `false` → identical behavior for existing consumers), and fail-loud via a separate post-upload check step that never loses the artifact. Output key set matches the old heredoc exactly. No correctness, security, or regression blockers found.

Minor notes (non-blocking):

- [P3] `scripts/emit_pr_receipt.py:115` — stdout logs `digest=sha-256:{digest}` but the `receipt_digest` GitHub output is written as bare hex; harmless but mildly confusing for anyone diffing the two.
- [P3] `action.yml` receipt step — the degraded-verify branch in `emit_pr_receipt.main()` writes `receipt_path` + `receipt_verified=false` to `$GITHUB_OUTPUT`, then the action's `receipt_failed` `rm -f`s the receipt, leaving `receipt-path` pointing at a deleted file. Not reachable inside the action itself (it `pip install`s `jsonschema`), so it only affects standalone invocations; worth tightening so `receipt_path` isn't emitted when verification fails.
- [P3] `action.yml` receipt step — `pip install "$GITHUB_ACTION_PATH"` builds/installs the entire aragora source tree into a fresh venv on every `emit-receipt: true` run; for third-party marketplace consumers this is a heavy, slow, dependency-resolution-prone step. Opt-in and failure-handled, but consider a slimmer extras target.
- [P3] Test gap: `tests/scripts/test_emit_pr_receipt.py` feeds `CollectOutcome.to_dict()` directly; nothing asserts the real seam — that `collect_quorum_evidence.py --json` stdout is parseable by `collect_outcome_from_dict`. A future divergence there would only surface at action runtime (caught, but as a failed run rather than a unit test).

dogfood: yes

scarmani:

Head: fc8da32 (fc8da32ec911070f2d93166060eda58e69c04f81), committed 2026-06-30T16:13:17Z.
PR: #8669.
Model family: claude

Verdict: PASS

Verified against the actual referenced code: `hydrate_env_from_secrets`, `collect_outcome_from_dict`, `collect_outcome_to_decision_receipt`, `decision_receipt_to_odr`, `load_odr_schema`, `odr_content_digest` all exist with compatible signatures; provider keys are in `MANAGED_SECRETS`; `--reviewers` is `nargs="+"` so `--reviewers claude openai --json` parses correctly; `scripts/` is an importable package so the new tests collect. Security posture is solid — PR number / max-diff-size / receipt-reviewers are all validated before use, untrusted `${{ }}` values are routed through env vars (not interpolated into `run:` bodies), GITHUB_OUTPUT values are newline-guarded, the venv lives outside the uploaded artifact dir, and the receipt is written before verification so a verify hiccup never loses the artifact. The fail-closed / opt-in / "check receipt emission at end" flow is internally consistent (no `true`/`false` output collision).

Minor non-blocking notes:

- [P3] `scripts/collect_quorum_evidence.py:_hydrate_provider_secrets` — the `except (OSError, RuntimeError, ValueError)` around `hydrate_env_from_secrets(...)` is effectively dead: that function already swallows those same exceptions internally and returns, so the "hydration skipped" branch can never fire (and any genuinely unexpected exception type would escape uncaught). Cosmetic.
- [P3] `action.yml` "Emit decision receipt" step — `collect_quorum_evidence.py`'s exit status is ignored; only `[ -s outcome.json ]` gates `emit_pr_receipt.py`. A collector that exits nonzero but still wrote partial JSON to stdout would be fed to the emitter. Low likelihood (stderr is redirected to the log and `--json` writes complete output only on success), but checking `$?` would be more robust than the size check alone.

dogfood: yes

scarmani:

Head: fc8da32 (fc8da32ec911070f2d93166060eda58e69c04f81), committed 2026-06-30T16:13:17Z.
PR: #8669.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #8673, opened 2026-06-28
feat(quorum): reviewer transport hardening — Anthropic-API fallback + two-tier timeouts

scarmani:
Standalone extraction from the M2 work (broadly useful beyond it).

The merge-gate `claude` reviewer was **CLI-only** (then OpenRouter). In keyless-CLI environments — CI runners, headless servers — with `ANTHROPIC_API_KEY` available (e.g. via AWS Secrets Manager), it now falls back to the **direct Anthropic API** (the `anthropic-api` agent, relabelled to the `claude` family so it counts) **before** OpenRouter. This lets the gate form a western-family quorum from API keys alone, instead of depending on subscription CLIs or OpenRouter.

**Purely additive / safe:** the API path only fires when the CLI fails AND `ANTHROPIC_API_KEY` is set; if both CLI and API fail, the original failure is returned so the existing OpenRouter fallback still applies. **255 existing quorum tests unchanged**; 5 new tests cover CLI-success, CLI-fail+key→API, CLI-fail+no-key, both-fail, and the default-runner route.

Discovered while validating the M2 receipt self-test in real CI (claude reviewer couldn't use ANTHROPIC_API_KEY at all).

### DIFF
--- a/aragora/swarm/quorum_evidence.py
+++ b/aragora/swarm/quorum_evidence.py
@@ -303,9 +303,21 @@ _MAX_DIFF_CHARS = 60_000
 _PER_FILE_TRUNCATION_MARKER = "\n[hunk truncated; full changed-file list is above]\n"
 # Cap reviewer output so a runaway model cannot exceed GitHub's per-comment limit.
 _MAX_REVIEWER_CHARS = 32_000
-_CLAUDE_TIMEOUT = 300
-_CODEX_TIMEOUT = 300
-_REVIEWER_TIMEOUT = 300
+# A genuine review of a large diff can legitimately take many minutes, so the
+# ceiling is generous (10 min default; raise via the *_TIMEOUT_SECONDS env vars
+# to 1200/1800 for very large diffs). The cost of being patient is bounded by a
+# fast liveness probe below: a wedged/contended CLI fails in ~probe seconds
+# rather than burning the full ceiling.
+_CLAUDE_TIMEOUT = 600
+_CODEX_TIMEOUT = 600
+_REVIEWER_TIMEOUT = 600
+# Fast-fail liveness probe: before committing to the long review timeout, check
+# the CLI answers a trivial prompt within this short window. A timeout here means
+# "no valid response is coming" (binary wedged, auth prompt, contention) → skip
+# fast. Set ARAGORA_REVIEWER_PROBE_TIMEOUT_SECONDS=0 to disable probing.
+_CLI_PROBE_TIMEOUT = 90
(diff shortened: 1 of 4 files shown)


scarmani:

Head: f87980c (f87980c2af6f061d3f7c90cfab134fd645695ab9), committed 2026-06-29T19:19:55Z.
PR: #8673.
Model family: claude

Verdict: PASS

- [P3] `_cli_liveness_probe` / `_run_claude_cli` (quorum_evidence.py ~1085-1110): the probe runs the **full** `claude -p --strict-mcp-config` CLI (a real model invocation with prompt "Reply with exactly: OK") before every Claude review, so the healthy/common path now spawns the CLI twice and consumes extra subscription/rate-limit budget + latency. Its stated benefit — catching fast auth/config failures before the 600s ceiling — is largely already provided by the real review call, which itself fast-fails on a non-zero exit. The genuine residual benefit (avoiding ingestion of a ~32k-char prompt before an auth failure surfaces) only applies on the rare failure path. Consider making the probe opt-in rather than default-on, or dropping it.
- [P3] `_run_claude_cli` probe-error branch: a transient/flaky non-zero probe exit returns the probe error and skips the real review entirely, a behavior change from "always attempt the real review." Low risk in practice (the real call is the identical command moments later and would usually also fail, and the OpenRouter fallback in `default_reviewer_runner` still fires), but a one-off probe blip can now suppress an otherwise-runnable review.
- [P3] `_run_claude_reviewer` (quorum_evidence.py ~1161): the new API fallback routes through `anthropic-api` (`AnthropicAPIAgent`, which carries `QuotaFallbackMixin`), so when the CLI fails and the Anthropic key is quota-exhausted, the resulting OpenRouter-backed review is relabeled `family="claude"` and counted as the claude vote. This newly makes an OpenRouter transport reachable on the merge gate's claude path (previously CLI-only). It's consistent with the existing grok/gemini API-fallback behavior and isn't fabrication, but for a quorum module whose core invariant is genuine, disclosed heterogeneous reviewers, confirm that a quota-fallback "claude" vote is acceptable for quorum-diversity accounting.

dogfood: yes

scarmani:

Head: f87980c (f87980c2af6f061d3f7c90cfab134fd645695ab9), committed 2026-06-29T19:19:55Z.
PR: #8673.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

PR: #8673
Exact head: f87980c2af6f061d3f7c90cfab134fd645695ab9
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #8673 is non-draft and live exact-head checks/merge-packet remain otherwise green.

PR #8676, opened 2026-06-28
[codex] fix action.yml YAML parsing

scarmani:
## Summary

Carves the public-surface YAML parse fix out of draft PR #8669 without bringing the broader M2 receipt / mission-cadence stack or the #8667 dependency with it.

Current `origin/main` still fails PyYAML on `action.yml` because the embedded `python3 - <<'PY'` body starts at column 0 inside a composite-action `run:` block. This moves that review-count extraction into a stdlib helper and keeps `action.yml` as valid YAML.

## Scope

- replace the inline Python heredoc in `action.yml` with a call to `scripts/extract_review_counts.py`
- add focused tests for valid, missing, and malformed review JSON
- no receipt-emission, quorum, workflow, mission-cadence, or #8667 changes

## Validation

- `python3 -c 'import yaml; yaml.safe_load(open("action.yml", encoding="utf-8")); print("action.yml parses")'`
- `python3 -m pytest tests/scripts/test_extract_review_counts.py`
- `python3 -c 'import yaml; data=yaml.safe_load(open("action.yml", encoding="utf-8")); steps=data["runs"]["steps"]; review=next(s for s in steps if s.get("id") == "review"); assert "scripts/extract_review_counts.py" in review["run"]; assert "python3 - <<" not in review["run"]'`
- pre-commit and push hooks passed, including `check yaml`, ruff, format, mypy, gitleaks, and portability guard

`actionlint action.yml` now reaches normal workflow-schema complaints because this file is action metadata, not a workflow; it no longer fails on the YAML scanner issue.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -188,50 +188,8 @@ runs:
         # Action-level metadata and extracted counts from JSON output
         echo "review_log_path=./aragora-artifacts/review.log" >> $GITHUB_OUTPUT
 
-        python3 - <<'PY' >> "$GITHUB_OUTPUT"
-import json
-from pathlib import Path
-
-path = Path("./aragora-artifacts/review.json")
-if path.is_file():
-    print("review_json_path=./aragora-artifacts/review.json")
-    try:
-        data = json.loads(path.read_text())
-    except Exception:
-        data = {}
-else:
-    print("review_json_path=")
-    data = {}
-
-def count_list(key: str) -> int:
-    value = data.get(key, [])
-    return len(value) if isinstance(value, list) else 0
-
-unanimous = count_list("unanimous_critiques")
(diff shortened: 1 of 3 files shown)


scarmani:
Superseded by #8669. Line-diff confirms an identical fix: both extract the column-0 `python3 - <<'PY'` heredoc into `scripts/extract_review_counts.py` and call it from action.yml (the script bodies differ only in a docstring line). #8669 contains this same YAML fix **plus** the receipt-emission feature (CollectOutcome→DecisionReceipt→ODR artifact), validated live in CI. Nothing unique here to harvest. Recommend closing in favor of #8669 to avoid a double-merge conflict on these three files.

PR #8700, opened 2026-06-29
fix(version): derive aragora.__version__ from canonical source (2.8.0 → 2.9.0)

scarmani:
## Problem

`aragora/__init__.py` hard-coded `__version__ = "2.8.0"` while the authoritative `pyproject.toml` and the canonical `aragora/__version__.py` module both declare `2.9.0`. So `import aragora; aragora.__version__` **lied about the installed version**:

```
# before
$ python -c "import aragora; print(aragora.__version__)"
2.8.0
```

## Root cause

`scripts/check_version_alignment.py` (run in CI via `test.yml` / `sdk-parity.yml`) treats `aragora/__version__.py` as the **single source of truth** and validates every manifest + doc against it. Every checked source is already `2.9.0` — but the script **never checks the top-level `aragora/__init__.py`**, so its duplicated literal drifted unnoticed.

## Fix

Re-export `__version__` from the canonical `aragora/__version__.py` module instead of hard-coding a literal:

```python
from aragora.__version__ import __version__ as __version__
```

This is the same pattern `aragora/compat/openclaw/standalone.py:348` already uses. It ties the package attribute to the CI-enforced source of truth so it **can never drift again** — superior to a bare literal bump (drifts at the next release) or `importlib.metadata` (whose fallback literal is itself a second drift source and which depends on the package being pip-installed). The version module is stdlib-only and side-effect free, so importing it at package init is safe and does not pull in any heavy debate-surface submodule.

> ⚠️ **Protected file.** `aragora/__init__.py` is on the project's Protected Files list (CLAUDE.md). This change was explicitly approved for this single, narrow version-reconciliation edit; no other edits were made to the file. Likely **Tier 3–4** (touches a public surface) — leaving evidence-ready for human settlement, not self-merging.

## Verification

```
# after
aragora.__version__                          -> 2.9.0   (import aragora)
from aragora import __version__               -> 2.9.0   (error_monitoring / health consumers)
from aragora.__version__ import __version__   -> 2.9.0   (standalone.py pattern, unchanged)
```

- `scripts/check_version_alignment.py` → **All versions aligned!**
- `ruff check` / `ruff format --check` / `mypy` on the file → clean
- `pytest tests/server/test_error_monitoring.py tests/handlers/test_platform_config.py` → **92 passed**

## Other version literals (out of scope — flagged for follow-up)

Discovered while scoping, intentionally **not** changed here to keep this PR narrow:

- **`sdk/python/aragora_sdk/__init__.py:268`** is still `__version__ = "2.8.0"` while its own `sdk/python/pyproject.toml` (`aragora-sdk`) is `2.9.0` — a real secondary drift. The alignment script misses it because its configured path `sdk/python/aragora/__init__.py` **does not exist** (the actual package dir is `aragora_sdk`). Worth a follow-up: fix the literal *and* correct the path in `check_version_alignment.py`.
- Separate distributions with independent versioning (not drift): `aragora-debate` (`0.2.3`), `aragora-verify` (`0.1.0`), `aragora/genesis/__init__.py` (`0.1.0`).

### DIFF
--- a/aragora/__init__.py
+++ b/aragora/__init__.py
@@ -13,7 +13,13 @@ from __future__ import annotations
 import importlib
 from typing import Any
 
-__version__ = "2.8.0"
+# Re-export the package version from the canonical single source of truth
+# (``aragora/__version__.py``), which CI keeps aligned with ``pyproject.toml``
+# via ``scripts/check_version_alignment.py``. Deriving it here instead of
+# hard-coding a literal prevents ``aragora.__version__`` from silently drifting
+# away from the declared/installed version. The version module is stdlib-only
+# and side-effect free, so importing it at package init is safe.
+from aragora.__version__ import __version__ as __version__
 
 _EXPORT_MAP = {
     "Agent": ("aragora.core", "Agent"),


scarmani:

Head: 51aeeff (51aeeffa2efdf59bcca19abaec065984df44865d), committed 2026-06-29T23:42:47Z.
PR: #8700.
Model family: claude

Verdict: PASS

I verified the change against the actual repo state rather than the comment's claims, and it holds up as a genuine fix:

- `pyproject.toml` declares `version = "2.9.0"` and `aragora/__version__.py` (canonical, per `scripts/check_version_alignment.py`) computes `2.9.0`. The old hardcoded `"2.8.0"` literal in `__init__.py` was **stale/drifted** — it is *not* checked by the alignment script (which only validates pyproject/package.json/`sdk/python/...`), which is exactly how it drifted. Deriving it removes the drift. Runtime `aragora.__version__` correctly moves 2.8.0 → 2.9.0, matching the declared/installed version and `docs/releases/v2.9.0.md`.
- No circular import / side-effect risk: `aragora/__version__.py` imports only `from __future__ import annotations` and defines pure constants/functions. Comment's safety claim is accurate.
- No regression in consumers: `from aragora import __version__` (error_monitoring.py, health/detailed.py) gets a string; `aragora/compat/openclaw/standalone.py` already uses `from aragora.__version__ import ...` (with an explicit ambiguity comment) and is unaffected. The eager init-time import actually makes the string-vs-submodule attribute resolution *deterministic* (the `as __version__` rebind wins last), removing a latent footgun.
- No test pins the literal: `test_version_fallback_on_import_error` patches `_get_version` directly; `test_version_is_string`/`_not_empty` still hold.
- No packaging break: the published `aragora` dist ships `aragora/__version__.py` as a normal module; the `aragora-debate` wedge is a separate `aragora_debate` package and does not consume this file.

Findings:
- [P3] `aragora/__init__.py` is a CLAUDE.md-designated protected file ("Do NOT modify without explicit approval"). The change is benign and correct, but the operating contract requires explicit sign-off — confirm approval is recorded on the PR.

dogfood: yes

scarmani:

Head: 51aeeff (51aeeffa2efdf59bcca19abaec065984df44865d), committed 2026-06-29T23:42:47Z.
PR: #8700.
Model family: grok

Verdict: PASS

- [P3] `aragora/__init__.py` — Comment claims CI keeps `aragora.__version__` aligned via `scripts/check_version_alignment.py`, but that script only regex-scans literal `__version__ = "..."` assignments (e.g. `sdk/python/aragora/__init__.py`) and does not check `aragora/__init__.py` or perform a runtime `import aragora` assertion; the re-export fixes the current 2.8.0 vs 2.9.0 drift, but a future broken/missing import would still slip through CI.
- [P3] `aragora/__init__.py` — Prefer relative `from .__version__ import __version__` over absolute `from aragora.__version__ import __version__` in package `__init__.py` (idiomatic, marginally safer during partial package init); not a functional defect here.

dogfood: yes

PR #8730, opened 2026-06-30
feat(tooling): RLM-backed digest of sibling agent (Codex) sessions

scarmani:
Makes cross-agent visibility one command. Codex writes full per-session transcripts to `~/.codex/sessions/**/rollout-*.jsonl` (MBs each); reading one wholesale blows a reviewer's context. `scripts/agent_session_digest.py` extracts the reviewable signal (prompts, decisions, exec commands, referenced PRs) deterministically, and with `--rlm` runs it through Aragora's **own RLM** (`aragora rlm compress`+`query`) for a recursive summary.

Read-only; deterministic path needs no creds. `--latest` / `--session <id>` / `--path` / `--json`. 3 focused tests; live-smoked on a real rollout. Foundation for a single-coordinator cross-agent view to stop two agents editing the same core files.

### DIFF
--- a/scripts/agent_session_digest.py
+++ b/scripts/agent_session_digest.py
@@ -26,9 +26,50 @@ import time
 from pathlib import Path
 from typing import Any
 
-DEFAULT_SESSIONS_ROOT = Path.home() / ".codex" / "sessions"
+from aragora.swarm.agent_bridge.codex_source import default_codex_home
+
+
+def default_sessions_root() -> Path:
+    return default_codex_home() / "sessions"
+
+
+DEFAULT_SESSIONS_ROOT = default_sessions_root()
+MAX_ROLLOUT_SCAN = 200
 _PR_RE = re.compile(r"#(\d{3,6})\b")
 _CMD_RE = re.compile(r'"cmd"\s*:\s*"([^"]{0,160})')
+
+
+def _rollout_session_id(path: Path, meta_id: str | None = None) -> str:
+    if meta_id:
+        return meta_id
+    name = path.name
+    if name.startswith("rollout-") and name.endswith(".jsonl"):
+        stem = name[len("rollout-") : -len(".jsonl")]
+        match = re.match(r"\d{4}-\d{2}-\d{2}T\d{2}-\d{2}-\d{2}-(?P<sid>.+)", stem)
+        return match.group("sid") if match else stem
+    return path.stem
+
+
+def _iter_rollouts(root: Path, *, since_hours: float | None = None, now: float | None = None):
+    cutoff = (
+        None
+        if since_hours is None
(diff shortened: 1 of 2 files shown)


scarmani:

Head: 89dfe10 (89dfe102c629420043d622da7bdeec3a0066ec92), committed 2026-07-01T19:51:43Z.
PR: #8730.
Model family: claude

Verdict: PASS

Reviewed both new files (a standalone CLI script + its tests). No existing code is modified, so regression surface is limited to the new `--all`/parsing paths. The tricky bits hold up: `_iter_rollouts` sorts by `key=item[0]` (mtime only), so equal-mtime tuples never fall through to a `Path` comparison `TypeError`; the two glob patterns (flat + 3-deep canonical) are disjoint, so no double-yield; `prs` is a `set` so the "message in both event_msg and response_item" path can't double-count PRs; and `_pr_ids_from_call_args`' `isdecimal()` guard correctly excludes `isdigit()`-only chars (e.g. `²`) that would crash the terminal `sorted(prs, key=int)`. Subprocess calls use list-form args (no shell), so `--rlm`/rollout text can't inject. Tests cover the meaningful edges (streaming vs read_text, structured vs string call args, dedupe, unicode digits, mtime windowing, uncapped `--session` scan).

- [P3] `scripts/agent_session_digest.py:99` (`coordinator_view` / `--all`) calls `extract_turns()` — a full line-by-line read — on *every* rollout in the window (up to `MAX_ROLLOUT_SCAN=200`, each potentially MBs), which is exactly the "read the whole transcript" cost the tool's docstring says it exists to avoid. Fine for the 24h/default case, but a busy sessions dir makes `--all` heavy; consider a lighter scan (e.g. counts-only / early-out) for coordinator view.
- [P3] `scripts/agent_session_digest.py:225` dedupe of a turn appearing as both `event_msg` and `response_item` relies on byte-identical text after `.strip()[:300]`; list-joined assistant content ("a"+" b" → double space) or minor formatting drift silently defeats it and double-counts a decision. Cosmetic only (counts, not correctness), and acknowledged in the comment — worth a note, not a blocker.

dogfood: yes

scarmani:

Reviewer: openai (openai) — independent adversarial model review via the Aragora OpenAI reviewer, grounded on the exact PR head.
Head: 89dfe10 (89dfe102c629420043d622da7bdeec3a0066ec92), committed 2026-07-01T19:51:43Z.
PR: #8730.
Model family: openai

Verdict: PASS
- [P3] scripts/agent_session_digest.py:165-217 — Prompt/decision dedupe and `event_msg` PR extraction operate on 300-char truncated text, so two distinct long turns with the same prefix can be collapsed and PR refs appearing later in an `event_msg`-only prompt/decision are missed. Use full text for dedupe/PR scanning and truncate only when storing/rendering snippets.

dogfood: yes

PR #8740, opened 2026-07-01
refactor(events): registry-backed cross-subscriber dispatch + layered bootstrap (P4a Batch E1)

scarmani:
## P4a EventBus inversion — Batch E1 (ENABLER)

Domain-free cross-subscriber registry core. Reworks `aragora/events/cross_subscribers/{__init__,manager}.py` to registry-backed dispatch and adds an explicit `bootstrap()` plus layered composition roots, so event-subscriber home modules can self-register at import time without `events/` importing any domain code. Supersedes cancelled Batch 2a/2b. See `docs/architecture/P4A_EVENTS_QUEUE_INVERSION.md` §4.1, §4.4, §10 (E1).

### What changed
- **New** `aragora/events/cross_subscribers/registry.py`: domain-free registry (`register_subscriber` / `register_factory` / `get_registered_subscribers` / `registered_subscriber_names` / `reset_registry`) + a `CrossSubscriber` Protocol. **ZERO** domain imports (eager or lazy).
- Reuses the existing domain-free `subscribers/config.SubscriberStats` (no duplicate/rename).
- `__init__.py`: adds idempotent `bootstrap()` returning the registry-backed manager; `get_cross_subscriber_manager()` import path **unchanged** (`aragora.events.cross_subscribers`).
- `manager.py`: adds idempotent `apply_registered_subscribers()` wiring registry subscribers into the existing per-event dispatch/stats/retry machinery.
- **New** layered composition roots: domain-subset `debate.event_subscribers.bootstrap_debate_event_subscribers()` and interface-superset `server.startup.event_subscribers.bootstrap_event_subscribers()`.
- Repoint the **actual** callers (grok residual #5): `debate/extensions.py` + `debate/knowledge_manager.py` → domain-subset bootstrap; `cli/commands/stats.py` → interface-superset bootstrap.

### Acceptance (doc §10 per-batch)
1. `check_import_contracts.py --layers foundation,infrastructure`: no new foundation/infra edge (only documented ambient `utils → caching` #8672). E1 clears none (pure enabler).
2. Full-layer re-check: no new domain/application/interface edge (only the 4 documented ambient sibling-fleet edges).
3. `get_cross_subscriber_manager()` import path unchanged.
4. `make test-smoke` exit 0 (VAL-P4A-013); smoke tier 138 passed; targeted registry tests 11 passed.
5. CI changed-file typecheck green ("no issues found in 8 source files").

[...]

Issue #8739, opened 2026-07-01
advisory-settle: unreachable in enforcing merge-quorum job (blocks #8738 enable)


scarmani:
The advisory-settle capability (code merged in #8729, default-OFF) is **not yet reachable in the enforcing merge-quorum job**, so enabling the flag (#8738) alone is a no-op in CI. Verified via review (openai [P2] on #8738) and code trace.

**Root cause:** the enforcing `Aragora Merge Quorum` workflow job passes its check only on `status=="satisfied"`. Inside that job, the `aragora-merge-quorum` row is the *current self-check* and is excluded from the effective required set (`_is_required_pr_check_current_merge_quorum_self_check`). So for a PR whose ONLY failing required check is model-quorum, `required_failing_or_cancelled_checks` is empty → `required_quorum_only_failure` (review_queue.py ~L2578) is `False` → `advisory_settle_eligible` is `False` (it requires `quorum_only_required_failure`) → `status` is `needs_model_review_quorum`, not `satisfied` → the check fails. advisory_settle only computes for EXTERNAL consumers (settle tooling), where the row is visible.

**Fix (needs care — has broad interactions):** make advisory_settle reachable in-job WITHOUT disturbing the existing self-check semantics. A naive change (treat `ignored_self_check_count>0` as a quorum failure in `required_quorum_only_failure`) is TOO broad — it fires on every workflow run and **broke 2 tests**:
- `TestBuildQueueAndPacket::test_merge_quorum_self_check_uses_required_gate_despite_shadow_jobs`
- `TestBuildQueueAndPacket::test_required_pr_checks_gate_ignores_stale_self_row_inside_quorum_run`

Better direction: give advisory_settle its own reachability predicate — "quorum NOT satisfied AND zero non-quorum required failures AND no pending required checks" — computed independently of the self-check-row presence, rather than reusing `quorum_only_failure`. Reconcile the 2 tests above.

**Until fixed:** keep #8738 (flag) HELD; the code stays safely on main (OFF). This also blocks #8730 (digest tool) from settling via advisory-settle.

scarmani:
Resolved by #8741 (merged 2026-07-01T15:55Z): `advisory_settle_surface_clear` gives advisory_settle a self-check-independent reachability predicate inside the enforcing merge-quorum job, so #8738's flag is no longer inert. Verified today by tracing the live gate on origin/main (review_queue.py:2600/3301) and an end-to-end reviewer run on #8730. 

[...]

PR #8738, opened 2026-07-01
ci(gate): enable advisory-dissent settlement (default→ON)

scarmani:
One-line activation of the advisory_settle capability whose code merged (default OFF) in #8729 (e59d9a4459). Sets `ARAGORA_ENABLE_ADVISORY_DISSENT_SETTLE: "1"` alongside the existing severity-gated + tiered-gate flags.

Effect: a Tier 0-2 PR whose ONLY failing required check is model-quorum settles via `verdict=advisory_settle` when a **validated** western-frontier review is present at head, there is **genuine (validated-source) advisory dissent**, and **zero [P0]/[P1]** findings remain — surfacing advisory_findings for follow-up. Tier 3-4 unaffected. Revert = flip OFF.

PR #8729, opened 2026-06-30
feat(gate): opt-in advisory-dissent settlement (default OFF)

scarmani:
Tier-4 / merge-authority change. **Default OFF (`ARAGORA_ENABLE_ADVISORY_DISSENT_SETTLE`), behavior-preserving until enabled.**

When ON, a Tier 0-2 PR settles via `verdict=advisory_settle` when: all non-quorum required checks green, >=1 western-frontier review, and ZERO [P0]/[P1] blocking findings across all reviews; advisory findings are surfaced for follow-up issues. Tier 3-4 unaffected (keep human settlement).

Unblocks substantial new features that two thorough reviewers never both PASS (the structural gate finding). 496 tests pass (no regression). Spec: `docs/plans/2026-06-30-advisory-dissent-settlement-gate-packet.md`; advisory-findings follow-up #8725.

Enabling is a separate one-line workflow flip after sign-off.

PR #8741, opened 2026-07-01
fix(gate): make advisory_settle reachable in enforcing merge-quorum job (#8739) [Tier 4]

scarmani:
## Summary
Fixes **#8739** — the advisory-dissent settlement path merged in #8729 (default-OFF) is **unreachable inside the enforcing `Aragora Merge Quorum` job**, so enabling the flag (#8738) is currently a no-op in CI.

**Root cause:** `advisory_settle_eligible` gated on `quorum_only_required_failure`, which needs the merge-quorum row to be a *visible* failing required check. In-job that row is the excluded current self-check → failing set is empty → predicate False even though the quorum signal is the only thing missing. It only computed True for external settle-tooling.

**Fix:** add a self-check-independent predicate `advisory_settle_surface_clear` ("no NON-quorum required check is failing/cancelled/pending"), emitted on the `required_pr_checks` surface and consumed by the gate (OR'd with the old key for backward-compat, since `quorum_only_failure` implies a clear surface). `quorum_only_failure` semantics are untouched, so the two self-check guard tests named in #8739 stay green.

## Verification
- New in-job regression test: `test_flag_on_settles_in_enforcing_job_via_surface_clear` (RED→GREEN).
- The two guard tests from #8739 (`..._ignores_stale_self_row_inside_quorum_run`, `..._self_check_uses_required_gate_despite_shadow_jobs`) still pass — no reconciliation needed because the change is additive.
- `304 passed` (test_review_queue) + `196 passed` (test_quorum_evidence); ruff + mypy clean.
- Byte-identical behavior when `ARAGORA_ENABLE_ADVISORY_DISSENT_SETTLE` is OFF (default): `advisory_dissent_settle_enabled()` short-circuits first.

## ⚠️ Tier 4 — human risk sign-off required
This is merge-authority self-modification. Per docs/plans/2026-06-30-advisory-dissent-settlement-gate-packet.md it must settle via **human sign-off**, not the gate it changes. Leaving as **draft** for @scarmani.

Once this lands, #8738 (enable flag) stops being inert and the quorum-only-blocked pipeline (#8730, #8726, #8740, and future substantial features) can settle on advisory findings with the hard zero-[P0]/[P1] bar intact.

### DIFF
--- a/aragora/cli/commands/review_queue.py
+++ b/aragora/cli/commands/review_queue.py
@@ -2582,6 +2582,27 @@ def _build_packet(
             and not required_pending_checks
             and all(_is_merge_quorum_check(check) for check in required_failing_or_cancelled_checks)
         )
+        # Self-check-independent reachability predicate for advisory-dissent
+        # settlement (#8739). Unlike ``required_quorum_only_failure`` above, this
+        # does NOT require the merge-quorum row to be *visible* as a failing
+        # required check. Inside the enforcing Aragora Merge Quorum job the quorum
+        # row is the current self-check and is excluded from
+        # ``effective_required_checks``, so ``required_failing_or_cancelled_checks``
+        # is empty and ``required_quorum_only_failure`` is False even though the
+        # quorum signal is exactly what is missing. Externally the same row appears
+        # as a distinct failing entry. Both collapse to "no NON-quorum required
+        # check is failing/cancelled/pending" — the true condition advisory_settle
+        # needs. Computed separately so the in-job path is reachable WITHOUT
+        # disturbing the self-check semantics ``quorum_only_failure`` (and its guard
+        # tests) rely on. ``all(...)`` over an empty failing set is True (in-job);
+        # over a quorum-only failing set is True (external); over any non-quorum
+        # failure it is False (correctly keeps a real failing check blocking).
+        advisory_settle_surface_clear = (
+            required_available
+            and effective_required_count > 0
+            and not required_pending_checks
+            and all(_is_merge_quorum_check(check) for check in required_failing_or_cancelled_checks)
+        )
         # _rollup_non_green_diagnostics reports the raw GitHub rollup counts/sample
         # and is intentionally not filtered by ignore_own_quorum_check. The flag's
         # effect on the rollup is limited to the summary text computed above; the
@@ -2633,6 +2654,7 @@ def _build_packet(
                 :CHECK_SURFACE_DIAGNOSTIC_LIMIT
             ],
             "quorum_only_failure": required_quorum_only_failure,
+            "advisory_settle_surface_clear": advisory_settle_surface_clear,
         }
         if required_surface.get("error"):
             check_surfaces["required_pr_checks"]["error"] = str(required_surface.get("error"))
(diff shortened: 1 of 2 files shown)


scarmani:

Head: 0c1b26a (0c1b26afd783f39d2b24ca2ded40161d76c9ad5d), committed 2026-07-01T14:37:07Z.
PR: #8741.
Model family: claude

Verdict: PASS

Reviewed the full pre/post context of `_build_model_review_quorum` and `_build_packet` at head `0c1b26a`, not just the hunk. The change is correct, minimally scoped, and well-guarded.

- The new `advisory_settle_surface_clear` is identical to `required_quorum_only_failure` minus the `bool(required_failing_or_cancelled_checks)` clause, so it collapses to "no NON-quorum required check is failing/cancelled/pending." Real failing/pending non-quorum required checks still block: `all(_is_merge_quorum_check ...)` returns False on any non-quorum failure, and `not required_pending_checks` keeps pending checks blocking. Verified against `review_queue.py:2600` and `:2577`.
- Backward-compatibility holds: external settle-tooling surfaces that only populate `quorum_only_failure` leave `advisory_settle_surface_clear` absent → `False`, so `advisory_settle_reachable` (`review_queue.py:3305`) reduces to the old `quorum_only_required_failure` for those callers. The OR is a strict superset only where `_build_packet` populates both keys.
- The remaining `advisory_settle_eligible` guards are untouched (zero `[P0]/[P1]` findings, WF review present at head, genuine advisory dissent, `not unresolved_dissent`, `not quorum_satisfied`, Tier 0-2, flag default-OFF), so reachability was broadened without relaxing the settlement bar.
- Tests: `TestAdvisoryDissentSettleGate` (12) and the broader quorum/packet/settle slice (205) both pass against the PR head; the new `test_flag_on_settles_in_enforcing_job_via_surface_clear` correctly exercises the in-job shape (`quorum_only_failure=False`, `advisory_settle_surface_clear=True`).

- [P3] `aragora/cli/commands/review_queue.py:3305` — Behavior-change worth noting (not a bug): because `advisory_settle_surface_clear` is also `True` on a fully-green required surface (empty failing set) and `advisory_settle_eligible` never guards on `not has_failures`, advisory_settle is now reachable even when the merge-quorum required check is green/absent rather than strictly failing — i.e. whenever recomputed `quorum_satisfied` is False with genuine dissent present. This is the intended reframing ("no other required check blocks") and stays protected by the zero-blocking-finding + WF-review + genuine-dissent guards and the default-OFF flag, but it is a real widening from the prior "quorum check must be failing" precondition and should be understood as such.

dogfood: yes

scarmani:

Head: 0c1b26a (0c1b26afd783f39d2b24ca2ded40161d76c9ad5d), committed 2026-07-01T14:37:07Z.
PR: #8741.
Model family: grok

Verdict: PASS

- [P3] `tests/cli/commands/test_review_queue.py` — No end-to-end `_build_packet` test with `ARAGORA_ENABLE_ADVISORY_DISSENT_SETTLE=1` and in-job `GITHUB_WORKFLOW`/`GITHUB_JOB`; the new regression is only covered via synthetic `check_surfaces` on `_build_model_review_quorum`. `test_required_pr_checks_gate_ignores_stale_self_row_inside_quorum_run` should assert `advisory_settle_surface_clear is True` and, with the flag on, `verdict == "advisory_settle"`.
- [P3] `aragora/cli/commands/review_queue.py` (`_build_model_review_quorum`, ~3314) — Condition (3) comment still says `quorum_only_required_failure`; code now gates on `advisory_settle_reachable`. Audit/reason text also still says “only the model-quorum check is failing” on the in-job path where `failing_or_cancelled` is empty — misleading, not a logic bug.
- [P3] `tests/cli/commands/test_review_queue.py` (`TestAdvisoryDissentSettleGate`) — No case combining `_INJOB_QUORUM_SELF_CHECK_SURFACES` with a non-quorum required failure (e.g. `typecheck`); `test_required_pr_checks_gate_keeps_non_self_required_failure_blocking` covers `_build_packet` blocking but not the new advisory-settle gate with that surface shape.

dogfood: yes

scarmani:

PR: #8741
Exact head: 0c1b26afd783f39d2b24ca2ded40161d76c9ad5d
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #8741 is non-draft and live exact-head checks/merge-packet remain otherwise green.

PR #8839, opened 2026-07-04
fix(tooling): bound agent_session_digest --all so goal-cycle packets never time out

an0mium:
## Summary
- Bound the `agent_session_digest.py --all` coordinator path by sampling head+tail slices from each rollout instead of reading entire multi-hundred-MB transcripts.
- Add a `--time-budget-seconds` guard that reports explicit skipped rollout counts instead of silently dropping work.
- Mark sampled coordinator rows in text/JSON output while preserving exact `--latest`, `--session`, and `--path` streaming behavior.

## Motivation
The 2026-07-04 Fable goal-cycle packet builder reported:

`agent activity digest: TimeoutExpired: ... scripts/agent_session_digest.py --all --since-hours 24.0 timed out after 120 seconds`

The sessions root contains rollout JSONL files up to ~459 MB, so the coordinator view needs bounded per-file IO.

## Validation
- `python3 -m pytest tests/scripts/test_agent_session_digest.py -q` -> 21 passed
- `python3 -m ruff check scripts/agent_session_digest.py tests/scripts/test_agent_session_digest.py` -> passed
- `git diff --check` -> passed
- Before scoped command: `python3 scripts/agent_session_digest.py --all --since-hours 24` completed in 6.16s in the current environment, but the goal-cycle packet builder had just timed out at 120s.
- After patch: same command completed in 0.24s with sampled rows explicitly marked.

## Broader CI Note
- `make ci-required` was attempted. It passed initial `ruff check aragora/ tests/ scripts/` but failed during repo-wide `mypy aragora/ --ignore-missing-imports` on existing unrelated type errors across hundreds of files, starting in `aragora/server/commands/models.py` and connector/server modules. No failures were in this patch's touched files.

### DIFF
--- a/scripts/agent_session_digest.py
+++ b/scripts/agent_session_digest.py
@@ -73,6 +73,8 @@ def _iter_rollouts(
     since_hours: float | None = None,
     now: float | None = None,
     limit: int | None = MAX_ROLLOUT_SCAN,
+    deadline: float | None = None,
+    timeout_state: dict[str, bool] | None = None,
 ):
     """Yield rollout files newest-first.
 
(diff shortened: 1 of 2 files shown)


an0mium:
Closing as a frozen queue-drain churner, branch preserved. Exact head 0851581d47295615a5b5628fd243c917e8a6b501 had one repair attempt, then fresh current-head dry-run /tmp/ev_8839_openai_grok_dry_20260704T0748Z.json returned non-postable OpenAI and Grok CHANGES-REQUESTED with concrete P1/P2 findings in the same agent_session_digest sampling/deadline area. Per the anti-treadmill rule, this loop should not patch it again. Reopen only for a replacement approach that removes the fragile abstraction or with explicit operator authorization. No evidence was posted, no CI was rerun, and the branch is not deleted.

PR #8887, opened 2026-07-05
feat(scripts): render founder decision queue

an0mium:
## Summary

Adds a read-only `scripts/founder_decision_queue.py` helper that renders pending founder/operator one-word decision packets into a single markdown table. The helper scans local `.aragora/founder-decisions/*.md` packet files and optional exported issue-comment JSON, deduplicates equivalent rows, and emits either Markdown or JSON.

This supports the transport-loop goal by turning posted decision packets into a repeatable status surface instead of requiring chat archaeology.

## Validation

- `python -m pytest tests/scripts/test_founder_decision_queue.py -q` -> `3 passed, 8 warnings`
- `python -m mypy scripts/founder_decision_queue.py tests/scripts/test_founder_decision_queue.py` -> `Success: no issues found in 2 source files`
- `ruff check scripts/founder_decision_queue.py tests/scripts/test_founder_decision_queue.py` -> passed
- `git diff --check` -> passed
- `python scripts/founder_decision_queue.py --decisions-root /Users/armand/Development/aragora/.aragora/founder-decisions --now 2026-07-05T15:07:20Z` -> rendered #8756/#8406/#8886 decision rows
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` -> passed

## Safety

Read-only renderer only. No evidence apply, settlement, merge, status-setting, workflow edits, branch-protection edits, or protected-file edits.

PR #8756, opened 2026-07-01
feat(quorum): add operator-approved advisory settlement records

scarmani:
Implements M0a for #8747 / #8757 on top of the existing spec.

## Summary
- Adds an opt-in operator-approved advisory settlement record composer for advisory-only CHANGES-REQUESTED reviews.
- Gates posting on severity-gated dissent + advisory-settle flags, Tier 0-2, operator login, at least one supportive counting review, and linked follow-up issue refs.
- Wires --post-advisory-dissent / --operator-login / --followup-issue through collect_quorum_evidence.py and settle_pr.py.
- Keeps Tier 3-4 prepare-only and leaves the default path unchanged without explicit flags.

## Test Plan
- PYTHONDONTWRITEBYTECODE=1 PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTEST_ADDOPTS='-p no:cacheprovider' python3 -m pytest -q tests/swarm/test_quorum_evidence.py tests/swarm/test_review_adjudicator.py
- python3 -m py_compile aragora/swarm/quorum_evidence.py scripts/collect_quorum_evidence.py scripts/settle_pr.py tests/swarm/test_quorum_evidence.py
- python3 -m ruff check aragora/swarm/quorum_evidence.py scripts/collect_quorum_evidence.py scripts/settle_pr.py tests/swarm/test_quorum_evidence.py
- python3 -m ruff format --check aragora/swarm/quorum_evidence.py scripts/collect_quorum_evidence.py scripts/settle_pr.py tests/swarm/test_quorum_evidence.py
- bash scripts/automation_pr_preflight.sh origin/main HEAD

PR #8888, opened 2026-07-05
fix(scripts): tighten founder decision queue ruling matching

an0mium:
## Summary
- tighten founder decision queue ruling detection so one-word replies are exact-case matches
- require settlement comments to mention the target PR/issue number with a boundary instead of a substring
- add regressions for lowercase `b` vs expected `B`, exact `B`, and PR `#875` vs `#8756` settlement matching

## Validation
- `python -m pytest tests/scripts/test_founder_decision_queue.py -q`
- `python -m mypy scripts/founder_decision_queue.py tests/scripts/test_founder_decision_queue.py`
- `ruff check scripts/founder_decision_queue.py tests/scripts/test_founder_decision_queue.py`
- `git diff --check`
- `python scripts/founder_decision_queue.py --decisions-root /Users/armand/Development/aragora/.aragora/founder-decisions --now 2026-07-05T15:07:20Z`

## Notes
Follow-up to the non-blocking Claude `[P3]` comments from #8887 evidence on `scripts/founder_decision_queue.py`.

### DIFF
--- a/scripts/founder_decision_queue.py
+++ b/scripts/founder_decision_queue.py
@@ -245,13 +245,24 @@ def _target_number(target: str) -> str | None:
 
 
 def _comment_resolves_item(body: str, item: DecisionItem) -> bool:
-    first_line = _strip_markdown_code(_first_nonempty_line(body)).lower()
-    expected = item.expected_reply.lower()
+    first_line = _strip_markdown_code(_first_nonempty_line(body))
+    expected = item.expected_reply
     if first_line and first_line == expected:
         return True
     number = _target_number(item.target)
+    if number is None:
+        return False
     lowered = body.lower()
-    return bool(number and number in lowered and "settlement" in lowered)
+    if "settlement" not in lowered:
+        return False
+    number_pattern = re.escape(number)
+    target_patterns = [
+        rf"#\s*{number_pattern}\b",
+        rf"/(?:issues|pull|pulls)/{number_pattern}\b",
+        rf"\b(?:issue|pr|pull request)\s+#?\s*{number_pattern}\b",
+        rf"\b{number_pattern}\b",
+    ]
+    return any(re.search(pattern, lowered) for pattern in target_patterns)
 
 
 def _item_resolved_after_packet(source: DecisionSource, item: DecisionItem) -> bool:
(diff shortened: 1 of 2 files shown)


an0mium:

Head: b899733 (b899733777c5f94b5e6e2a379d2eed8a22dc0d0b), committed 2026-07-05T14:03:33Z.
PR: #8888.
Model family: claude

Verdict: PASS

- [P3] `scripts/founder_decision_queue.py:250` — the exact-reply match was changed from case-insensitive to case-sensitive (`first_line`/`expected` no longer `.lower()`). This is deliberate and tested, but it introduces false negatives on the primary detection path: a founder who replies `approve`/`yes`/`b` when the packet's `expected_reply` is `Approve`/`Yes`/`B` will no longer resolve the item, leaving it stuck in the queue and re-surfaced. Direction is safe (re-prompt, never a wrong auto-resolve), and code-fence stripping still handles `` `B` ``, but it degrades matching for in-flight packets generated before this change. Consider a case-insensitive compare (or normalizing `expected_reply` at packet-parse time) unless case truly disambiguates options.

- [P3] `scripts/founder_decision_queue.py:259-265` — the first three settlement `target_patterns` (`#\s*N`, `/(?:issues|pull|pulls)/N`, `(?:issue|pr|pull request)\s+#?\s*N`) are fully subsumed by the trailing catch-all `\b{number}\b`: any position where 1–3 match also has word boundaries around the number, so the fourth pattern already matches there. Net effect is "settlement" + a word-bounded occurrence of the number anywhere, so the specific issue/PR-reference forms are dead code. The real fix (substring → word boundary, e.g. #875 no longer matched by #8756) is correct and well covered by the new boundary test; just note the three patterns add no discrimination if the intent was to require an actual issue/PR reference context.

Core change is sound: `_target_number` always yields a non-empty digit string (so `re.escape`/`\b` construction is safe), the `number is None` early return is correct, and the added tests cover the boundary and case behavior.

dogfood: yes

an0mium:

Head: b899733 (b899733777c5f94b5e6e2a379d2eed8a22dc0d0b), committed 2026-07-05T14:03:33Z.
PR: #8888.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #8934, opened 2026-07-06
fix(tests): repair debate-am main-red in test_memory_manager_extended

an0mium:
## Summary
- Fixes current `origin/main` debate-am failures in `tests/debate/test_memory_manager_extended.py` by making best-effort memory/spectator side effects fail closed.
- Centralizes the broad side-effect catch in one helper so memory outcome updates and spectator notifications cannot break debate flow.
- Links the queue-conductor classification from PR #8912: https://github.com/synaptent/aragora/pull/8912#issuecomment-4889402505

## Root Cause
Two tests expected non-critical memory side effects to be isolated from debate execution, but `MemoryManager.update_memory_outcomes()` and `_notify_spectator()` only caught narrower exception classes. Plain adapter exceptions from mocks, and potentially third-party side-effect adapters, escaped and failed the debate-am shard on current main.

## Verification
- `PYTHONPATH=.:aragora-debate/src /Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest tests/debate/test_memory_manager_extended.py -k "test_handles_notification_failures_gracefully or test_partial_update_failures_handled" -v --timeout=90` -> 2 passed
- `PYTHONPATH=.:aragora-debate/src /Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest tests/debate/test_memory_manager_extended.py --timeout=90 -q` -> 22 passed
- `/Users/armand/.pyenv/versions/3.11.11/bin/python -m ruff check aragora/debate/memory_manager.py tests/debate/test_memory_manager_extended.py` -> passed
- `python3 scripts/report_code_quality.py --check` -> passed, except_exception ratchet at 900/900
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` -> passed
- Push hook changed-file mypy -> passed

## Safety
No workflow, branch protection, settlement, evidence, or outbox changes. This is a normal source repair PR for a current-main test failure.

PR #8941, opened 2026-07-06
fix(tests): reset handler global pollution

scarmani:
## Summary
- reset handler lazy-import caches when prior tests populate them with synthetic handlers
- restore social OAuth re-export identity after tests rebind module globals
- add regression coverage for both pollution paths

## Validation
- PYTHONPATH=.:aragora-debate/src /Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest tests/fixtures/test_autouse_global_pollution.py -q --tb=short
- PYTHONPATH=.:aragora-debate/src /Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest tests/handlers/test_handlers_selection.py::TestSelectionHandlerImport::test_handler_in_all_handlers -q --tb=short
- PYTHONPATH=.:aragora-debate/src /Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest tests/fixtures/test_autouse_global_pollution.py tests/handlers/test_handlers_selection.py::TestSelectionHandlerImport::test_handler_in_all_handlers -q --tb=short
- PYTHONPATH=.:aragora-debate/src /Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest tests/handlers -x --timeout=60 -n auto --dist loadscope -q --ignore=tests/connectors -k 'not benchmark and not slow' (61848 passed)

Context: addresses the current-main handler-smoke order-pollution failure observed while classifying #8934.

### DIFF
--- a/tests/fixtures/autouse.py
+++ b/tests/fixtures/autouse.py
@@ -1620,6 +1620,7 @@ except ImportError:
     _GlobalBaseHandler = None
 
 # Capture the side_effect property descriptor
+from unittest.mock import Mock as _GlobalMock
 from unittest.mock import NonCallableMock as _GlobalNCMock
 
 _global_side_effect_descriptor = None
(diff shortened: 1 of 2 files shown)


an0mium:

Head: 2f7426e (2f7426e366ba4ac69ae95d3d9201f3dd087f9532), committed 2026-07-06T14:31:24Z.
PR: #8941.
Model family: claude

Verdict: PASS

I verified against the local tree (which is at pre-PR main, confirming the surrounding context) that the refactor is faithful and the referenced internals all exist.

What I checked:
- **All prior repairs preserved.** The consolidated `_repair_global_mock_pollution` keeps side_effect, `extract_path_param`, `run_async`, and `Agent.__init__` repairs (the `extract_path_param` block is unchanged context sitting between the two diff hunks), plus the oauth_impl restore. Setup and teardown both call it — behavior-equivalent to the old duplicated inline copies, plus the two new repairs.
- **Exception safety (the key blast-radius risk).** Both new helpers are fully guarded: imports wrapped in `try/except (ImportError, AttributeError)`, None-guards before use, `.clear()` in `try/except AttributeError`. Since this is an autouse fixture over ~216k tests, an unguarded raise would nuke the whole suite — none present.
- **Internals exist and heuristics are valid.** `handlers._all_handlers_cache`/`_handler_cache`/`_get_all_handlers`/`_lazy_import`/`__getattr__("ALL_HANDLERS")` and the social `_oauth_states`/`_oauth_states_lock`/`_store_oauth_state`/`_validate_oauth_state` re-exports all present. `SelectionHandler` is a registered handler name, so "SelectionHandler absent ⇒ polluted" is a sound clean-cache invariant; `isinstance(..., Mock)` also catches MagicMock (subclass). The `__dict__.pop("ALL_HANDLERS")` correctly un-shadows the lazy `__getattr__` after `patch.object` leaves a real attribute behind.
- **New tests are correct.** Both regression tests pollute, repair, and assert recovery accurately; the social test correctly demonstrates re-export identity divergence and restoration, and cleans up after itself. Import path `tests.fixtures.autouse` is valid (registered via `pytest_plugins`; both `__init__.py` files exist).
- **Security:** test-only; no secret/injection surface. The social-OAuth alias repair actually tightens isolation of OAuth-state tests.

[P3] `tests/fixtures/autouse.py:1690` (`_repair_handler_lazy_cache_pollution`) clears `handlers._all_handlers_cache` but not the `_registry.ALL_HANDLERS` mirror that `_populate_registry()` fills from the same cache; if a test pollutes the cache and `_populate_registry()` then copies mocks into the registry list, that mirror stays polluted. This is a pre-existing gap and out of this PR's scope (the old guard didn't touch handler caches at all, so the change is strictly additive) — noting it only so the coverage boundary is explicit, not as a blocker.

dogfood: yes

PR #8996, opened 2026-07-07
Guard required contexts against path-filtered main pushes

an0mium:
## Summary
- add required-context policy inspection for mapped workflow `push: main` path filters
- fail when a branch-protection required context maps to a workflow whose main push trigger can be skipped by `paths` or `paths-ignore`
- add regression coverage for unfiltered and path-filtered required workflow triggers

## Validation
- `python3 -m pytest tests/scripts/test_check_required_check_priority_policy.py -q`
- `python3 scripts/check_required_check_priority_policy.py`
- `python3 -m ruff check scripts/check_required_check_priority_policy.py tests/scripts/test_check_required_check_priority_policy.py`
- `git diff --check`

### DIFF
--- a/scripts/check_required_check_priority_policy.py
+++ b/scripts/check_required_check_priority_policy.py
@@ -128,6 +128,46 @@ def _inline_yaml_value_mentions(value: str, expected: str) -> bool:
     )
 
 
+def _flow_mapping_value(value: str, key: str) -> str | None:
+    stripped = value.strip()
+    if not (stripped.startswith("{") and stripped.endswith("}")):
+        return None
+
+    match = re.search(
+        rf"(?<![A-Za-z0-9_-]){re.escape(key)}\s*:\s*(?P<value>\[[^\]]*\]|[^,}}]+)",
+        stripped[1:-1],
+    )
+    if match is None:
+        return None
+    return match.group("value").strip()
+
+
+def _flow_push_targets_main(value: str) -> bool:
+    branches = _flow_mapping_value(value, "branches")
+    if branches is not None:
+        return _inline_yaml_value_mentions(branches, "main")
+
+    branches_ignore = _flow_mapping_value(value, "branches-ignore")
+    if branches_ignore is not None and _inline_yaml_value_mentions(
+        branches_ignore,
+        "main",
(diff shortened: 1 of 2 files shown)


an0mium:

Head: bf7092a (bf7092a35441b00b38ea73a83f70f5a7b2d24599), committed 2026-07-08T15:37:03Z.
PR: #8996.
Model family: claude

Verdict: PASS

Reviewed the hand-rolled YAML parser and its integration into the CI-lint script. Traced block-style, inline-flow, top-level-flow, multiline-flow, and comment cases against the test suite and the four real mapped workflows (`lint.yml`, `sdk-parity.yml`, `openapi.yml`, `sdk-test.yml`). All use unfiltered `push: branches: [main]` with path-scoping inside internal jobs, so the new check does not fire on the current tree, and `test_repo_..._passes_for_current_tree` (`assert violations == []`) guards against regression. `mapped_workflows` is keyed off `REQUIRED_CONTEXT_TO_WORKFLOW_PATH` (not `path_set`), `paths` detection is correctly scoped to the `push` block via `_nested_yaml_block`, and the loop-var reuse of `context`/`rel` is benign. No correctness/security/regression defects found.

Minor, non-blocking:

- [P3] `_inline_yaml_value_mentions` (via `_flow_push_targets_main` / `_push_block_targets_main`) only matches the literal token `main`, so wildcard branch filters that include main (`branches: ['*']`, `['**']`) are read as *not* targeting main — a path-filtered required workflow using a wildcard branch filter would silently escape detection (false negative).
- [P3] Scope gap in `_main_push_trigger`: only the `push` trigger is inspected. A path-filtered `pull_request` trigger — the classic "required check never reports → PR stuck pending" footgun, which is arguably the higher-value case since required checks report on PR events — is not covered. Fine if intentional; worth a comment noting the deliberate push-only scope.

dogfood: yes

an0mium:

Head: bf7092a (bf7092a35441b00b38ea73a83f70f5a7b2d24599), committed 2026-07-08T15:37:03Z.
PR: #8996.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

Issue #8942, opened 2026-07-06
Charter enforcement wiring: charters.yaml consumer for gates (R8) + Appendix A machine encoding


an0mium:
Follow-up from advisory P2 dissent on PR #8927 round 2. See PR evidence comments at head f079d8f1: (1) wire a charters.yaml-reading checker for CHR-*/ARCH-* enforcement before RATIFIED status; (2) encode Appendix A package states in charters.yaml for the UNMAPPED freeze.

scarmani:
Additional requirement from advisory P2 dissent on PR #8956 (charter v0.4, merged): **symbol-level granularity**. CHR-X-040 parks `aragora/control_plane/registry.py` at path level while the prose exempts the kept health/liveness surface (consumed by `debate/team_selector.py`) — with `symbols: []` in charters.yaml, automation cannot distinguish the kept API from the parked remainder and could wrongly freeze legitimate edits. The checker (item 1) needs to support a `symbols`/`kept_symbols` field, and CHR-X-040 should enumerate the kept registry symbols once that field is consumable. Until then, gates should treat path-level PARK entries with a documented kept-surface as advisory-only for edits touching the kept symbols.

scarmani:
Post-merge verification update: PR #8968 landed the advisory charter-compliance checker on main at merge commit 1390b9a27aaa6ffa04800fa20728e53ee51ac1a3.

Verified from live main:
- `scripts/check_charter_compliance.py` exists on `origin/main`.
- `docs/architecture/charters.yaml` parses as `status=DRAFT` with 55 registry rows.
- `CHR-X-040` now has kept-symbol metadata for the registry health/liveness surface: `AgentRegistry`, `AgentStatus`, and `AgentInfo.is_alive`.
- `pytest -q tests/scripts/test_check_charter_compliance.py` passes: 8 passed.
- The checker passes on the merged #8909 range: `787a9562eb5f72438df23af3784f6d069b0ca00f^..787a9562eb5f72438df23af3784f6d069b0ca00f`.
- The checker passes on benign docs-only #8868 range: `589d955ea5fc55afba7f184c32ac5f428ad7650f^..589d955ea5fc55afba7f184c32ac5f428ad7650f`.

Scope boundary remains explicit: #8968 is advisory tooling only. CI / required-check wiring, branch-protection integration, merge/evidence/settlement gate integration, and charter ratification remain separate operator-approved follow-ups.

PR #9074, opened 2026-07-09
feat(governance): encode Appendix A package states

scarmani:
## Summary

Refs #8942.

- encode all 144 Intended Architecture Appendix A package states in `charters.yaml`
- flag new Python modules under UNMAPPED or previously unknown top-level packages
- preserve maintenance edits and DRAFT proposed-only semantics
- keep the YAML map, Appendix A table, and live top-level package inventory in parity

## Validation

- `python -m pytest -q tests/scripts/test_check_charter_compliance.py` (21 passed)
- `python -m ruff check scripts/check_charter_compliance.py tests/scripts/test_check_charter_compliance.py`
- `python -m ruff format --check scripts/check_charter_compliance.py tests/scripts/test_check_charter_compliance.py`
- `git diff --check`
- unmapped-new-file and maintenance-edit CLI dogfood

## Scope

Advisory checker and machine-readable charter data only. No workflow, required-check, branch-protection, settlement, or merge-authority changes.


scarmani:

Head: c076b03 (c076b03bf1cbd7c5191de33e09dc4211ae7b05be), committed 2026-07-21T04:53:02Z.
PR: #9074.
Model family: claude

Verdict: PASS

Verified against the repo, not just the diff: fetched head c076b03 into a temp worktree and ran the full test file — 27/27 pass, including the live-tree parity test. Confirmed `package_states` has exactly 145 entries (64 MAPPED / 81 UNMAPPED, status DRAFT), matching the updated Appendix A totals and the actual 145 top-level `aragora/*` directories on both the PR head and current main (`aragora/models` exists). Grepped `.github/`, `scripts/`, and the Makefile: no CI or script consumers of `check_charter_compliance.py`, so the new hard requirement that `charters.yaml` define `package_states` (ValueError otherwise) is self-consistent — script and data file ship together.

- [P3] `scripts/check_charter_compliance.py:529` (`check_diff`) — `load_charter_entries` and `load_package_states` each read and `yaml.safe_load` the same charters.yaml; the file is parsed twice per invocation. Harmless but easy to fold into one load.
- [P3] `scripts/check_charter_compliance.py` (`_coalesce_multiline_imports`) — an inline `#` comment on an intermediate continuation line (e.g. `from x import (\n  a,  # keep\n  b,\n)`) coalesces to a statement where `_parse_imported_names` truncates at the `#`, so `b` and later names are silently missed (false negative). Still a strict improvement over main, which missed all multiline imports.
- [P3] `tests/scripts/test_check_charter_compliance.py` (`test_main_package_states_match_appendix_a`) — asserts `status == "DRAFT"` and hardcodes 145/64/81; ratifying the charter or adding any package in an unrelated PR breaks this test. The count ratchet is clearly the intended enforcement mechanism, but the DRAFT assert is an extra tripwire that will read as an unrelated failure at ratification time.
- [P3] same test — the live inventory uses filesystem `iterdir()`, so any untracked non-dot/non-dunder directory under `aragora/` on a dev machine fails the test; deriving the set from git-tracked directories would be more robust.
- [P3] `scripts/check_charter_compliance.py` (`load_package_states`) — the ValueError for missing/invalid `package_states` propagates out of `main()` as a raw traceback rather than a clean error message; fail-closed is correct, presentation is rough.

dogfood: yes

scarmani:

Head: c076b03 (c076b03bf1cbd7c5191de33e09dc4211ae7b05be), committed 2026-07-21T04:53:02Z.
PR: #9074.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:
## Exact-head optional-check decision request

PR: #9074  
Exact head: `c076b03bf1cbd7c5191de33e09dc4211ae7b05be`  
Current main observed: `c6654778a0d894a6fb48281d118a0153030e966c`  
Observed: `2026-07-21T10:40Z`

Live revalidation is otherwise settlement-complete for this Tier-2 PR:

- open, non-draft, `MERGEABLE/UNSTABLE`;
- all 6 required checks pass;
- exact-head Claude/OpenAI evidence is counted with focused dogfood and no unresolved dissent;
- merge-packet's sole blocker is `mergeStateStatus=UNSTABLE`;
- no active branch lease or live owner process exists; the historical lane is released. Its preserved worktree was not touched.

The sole non-green check is non-required **Metrics Drift**, run `29802654220`, job `88546592275`. The validator ran fully and reported inventory movement: `@pytest.mark.parametrize` decorators `879 -> 895` (1.8%). This is a real legacy-validator result caused by added focused test coverage, not a product defect, contract regression, or ratchet regression. Re-running the deterministic job cannot change the result. Restacking and regenerating exact counts would churn valid exact-head evidence without removing the branch's intentional test inventory. Issue #9283 tracks replacing this exact-count friction with categorized inventory reporting and policy gates.

**Smallest operator decision:** explicitly accept this sole optional inventory signal for the exact head, or choose a restack/metrics-refresh round despite the evidence churn. No CI rerun is requested.

This comment is a decision request only. It is not authorization, settlement, or merge approval. No branch, check, evidence, or protection state was changed.

scarmani:

Head: 4cab279 (4cab2792161854c054c57666e123818a15ca4b89), committed 2026-07-24T21:57:46Z.
PR: #9074.
Model family: claude

**Verdict: PASS**

Verified positively: Appendix A parity is real — I parsed `docs/architecture/INTENDED_ARCHITECTURE.md` on main (144 rows, 64 MAPPED / 80 UNMAPPED, no duplicate paths) and diffed against the live `aragora/` directory set; `aragora/models` was the sole missing row, so the added row and the 145/64/81 totals are correct. I also simulated the new `parse_new_files` over `git diff --unified=0 HEAD~30...HEAD` and it matched `git diff --diff-filter=ACR --name-only` exactly (18/18, zero false positives, zero misses). Metrics churn (224,872 → 224,885) is consistent across all 7 doc copies and `METRICS.md`.

- **[P3]** `scripts/check_charter_compliance.py` `_top_level_package` — contradicts the Appendix A footnote this PR leaves in place. For `aragora/foo.py` it returns `aragora/<stem>`, a key the `package_states` regex (`aragora/[A-Za-z0-9_]+`) can never hold, so every new top-level single-file module is reported as "absent from Appendix A" — even the ones the footnote says "follow their nearest package's state" and that are explicitly chartered (`resilience_patterns.py`, `resilience_config.py`, CHR-X-006). Verified against history: commit `e80ba41c3` adds `aragora/api_errors.py`, which keys to a nonexistent package. Consider resolving `aragora/*.py` to `aragora` (or the nearest chartered row) instead of a fabricated package.
- **[P3]** `scripts/check_charter_compliance.py:main` — exit-code conflation, amplified by this change. `main()` returns 1 whenever `result.ok` is False, and `_render_text` prints "Charter compliance: FAIL", so a PROPOSED-only Appendix-A finding under a DRAFT charter is indistinguishable at the process level from a binding failure. I measured the blast radius: 14 of the last 200 commits would now exit 1. Verified this is advisory-only today — `grep -rl check_charter_compliance` over the repo hits only the test and `CHARTER_COMPLIANCE_CHECKER.md`, no workflow — but the doc's "advisory" framing no longer matches the exit code for the common case.
- **[P3]** `scripts/check_charter_compliance.py` `_coalesce_multiline_imports` ships with no tests in this PR (all added tests cover package growth), and has two behaviors worth covering: (a) `" ".join(part.strip() ...)` strips the opening line's indentation, so an indented multi-line import reaches `_line_reexports_or_defines_export`, whose `line[:1].isspace()` guard exists precisely to skip non-module-level lines — the coalesced form can report a define/re-export its single-line equivalent would not; (b) a trailing comment on the opening line (`from x import (  # noqa`) makes `_parse_imported_names` reduce the joined statement to `["("]`, silently dropping detection (`_parse_imported_names` splits on `#` before the paren strip).
- **[P3]** `scripts/check_charter_compliance.py:load_package_states` — `check_diff` now hard-raises `ValueError` on any charters file lacking `package_states`, where it previously degraded gracefully; it also re-reads and re-parses the whole YAML a second time after `load_charter_entries` already did. Threading `data`/`status` through one parse avoids both. No in-repo caller breaks (only the test and the doc invoke the script), so impact is latent.
- **[P3]** `tests/scripts/test_check_charter_compliance.py:test_main_package_states_match_appendix_a` enumerates `(repo_root/"aragora").iterdir()` rather than `git ls-files`, so any untracked or gitignored non-dot directory under `aragora/` fails the suite for reasons unrelated to the change under test. Verified it passes on this tree (`ls -d aragora/*/` = 146, the only extra being `__pycache__`, which the `startswith(("." , "__"))` filter drops), but `docs/METRICS.md` computes the same count via `git ls-files` — worth matching.
- **[P3]** `scripts/check_charter_compliance.py:parse_new_files` — the `shlex.split(raw_line)` branch on the `diff --git` header is redundant and only adds a mis-parse path: git does not quote paths containing spaces there, so `a/aragora/foo bar.py` splits into `a/aragora/foo` + `bar.py` and `remember()`s a bogus path, and an apostrophe in a path raises `ValueError` (caught, silently yielding nothing). Renames and copies are already covered by the `rename to `/`copy to ` handlers and new files by `new file mode` / `+++`, which is why my simulation showed no real-world divergence.

dogfood: yes

scarmani:

Head: 4cab279 (4cab2792161854c054c57666e123818a15ca4b89), committed 2026-07-24T21:57:46Z.
PR: #9074.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

Issue #9239, opened 2026-07-12
[quality-program C3] M-TRUST: restore test trustworthiness (randomized guard + nightly-core)


an0mium:
Part of #9039 (Signal-First Quality Recovery, campaign 3). Boss-ready mission for Factory/Droid.

## Files
- .github/workflows/test.yml (job test-pollution-randomized, ~line 1171) — workflow changes as SEPARATE Tier-4 draft PR
- tests/ (pollution fixes only), aragora/cli/ (the demonstrated sys.argv leak)
- Issues to land/close: #9122 (pollution quarantine), #8277 (verify #8286 fixed it, close with fixed-by evidence)

## Acceptance
- Randomized guard decoupled from test-fast; split into bounded domain slices, one fixed + one rotating seed; failing seed/node IDs persisted as artifacts
- The demonstrated CLI sys.argv pollution failure fixed with a regression test
- A small nightly-core signal exists and is green; full matrix retained as advisory debt
- Exit gate: guard produces a verdict nightly and nightly-core green >=12 of 14 nights

## Constraints
- Do NOT make pollution-suspect failures non-terminal until nightly-core is reliable
- Workflow edits ship as Tier-4 draft PRs (human approval); test/source fixes are Tier 0-2
- Two-attempt limit per unit; park with terminal-block reason
- Validation: run the affected test slice with the randomized seed locally before PR

an0mium:
### Progress terminal block — droid pass 1 for #9239

Three draft PRs opened; work split per the "workflow changes = separate Tier-4 draft PRs" constraint and the two-attempt limit per unit.

| Unit | PR | Tier | Status |
|------|----|------|--------|
| CLI `sys.argv` pollution fix + regression test | [#9244](https://github.com/synaptent/aragora/pull/9244) | 0-2 (source+test) | draft, tests green locally (5/5) |
| Randomized guard: bounded slices, fixed+rotating seed, failing-node artifacts | [#9245](https://github.com/synaptent/aragora/pull/9245) | 4 (workflow) | draft, awaits human approval |
| `nightly-core.yml` small nightly trust signal with verdict artifact | [#9246](https://github.com/synaptent/aragora/pull/9246) | 4 (new workflow) | draft, awaits human approval |

**Acceptance-criteria coverage:**
[...]

PR #9244, opened 2026-07-12
fix(cli): restore sys.argv after doctor dispatch (#9239)

an0mium:
Refs #9239 (M-TRUST: restore test trustworthiness).

## Problem
`python -m aragora doctor …` mutated `sys.argv` in place at `aragora/__main__.py:17` and never restored it, leaking process-wide state to any subsequent caller. Under pytest-xdist workers this compounded into cross-test pollution — the "demonstrated CLI sys.argv leak" called out in M-TRUST.

## Fix
- **aragora/__main__.py** — save/restore `sys.argv` around the doctor sub-CLI call via `try/finally`. The doctor branch still receives an argv without the `doctor` token, but the caller/worker's `sys.argv` is preserved on both normal return *and* exception propagation.

## Regression test
- **tests/cli/test_main_argv_pollution.py** — 5 tests pinning the contract:
  - `sys.argv` unchanged after doctor branch normal return
  - `sys.argv` unchanged after doctor sub-CLI raises
  - inner doctor CLI sees argv **without** the `doctor` token
  - non-doctor path never mutates `sys.argv`
  - bare invocation (`python -m aragora`) never mutates `sys.argv`

## Verification
```
$ python -m pytest tests/cli/test_main_argv_pollution.py -v --timeout=60
5 passed in 0.32s
```

## Scope
- Tier 0-2 (source + test only, no workflow edits)
- Small (2 files, +128/-4)
- No API surface change; behavior externally identical on happy path

### DIFF
--- a/aragora/__main__.py
+++ b/aragora/__main__.py
@@ -11,13 +11,25 @@ import sys
 
 
 def main():
-    """Route to appropriate command."""
+    """Route to appropriate command.
+
+    ``doctor`` is dispatched by consuming the token from ``sys.argv`` so the
+    inner command sees an argv that no longer contains it. The mutation is
+    restricted to a ``try/finally`` block so callers (tests, programmatic
+    embedders, ``pytest-xdist`` workers) never observe a permanently rewritten
+    ``sys.argv``. See #9239 (M-TRUST) for the pollution regression that
+    motivated the guard.
+    """
     if len(sys.argv) > 1 and sys.argv[1] == "doctor":
-        # Remove 'doctor' from args and run doctor command
-        sys.argv = [sys.argv[0]] + sys.argv[2:]
         from aragora.cli.doctor import main as doctor_main
 
-        sys.exit(doctor_main())
+        saved_argv = sys.argv
+        try:
+            # Remove 'doctor' from argv for the doctor sub-CLI
+            sys.argv = [saved_argv[0]] + saved_argv[2:]
+            sys.exit(doctor_main())
+        finally:
+            sys.argv = saved_argv
     else:
         # Fall through to main CLI
         from aragora.cli.main import main as cli_main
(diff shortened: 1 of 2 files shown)


an0mium:

Head: ff7cccd (ff7cccd0e944aaf06c324cc69ffce0c451d82bfd), committed 2026-07-12T02:04:03Z.
PR: #9244.
Model family: claude

Verdict: PASS

The fix itself is correct: `saved_argv` holds the original list object, the `try` block only rebinds (never mutates) `sys.argv`, and the `finally` restore runs on normal `sys.exit`, on `SystemExit` propagation, and on arbitrary exceptions — all three paths are pinned by tests. I verified against the repo that the unchanged `else` branch calls `cli_main()` without `sys.exit`, so tests 4–5 correctly expect a plain return, and that the lazy `from aragora.cli.doctor import main` at call time makes the `patch("aragora.cli.doctor.main", ...)` targets effective. Moving the import above the argv mutation is safe (doctor.py has no module-level argv reads).

- [P3] `tests/cli/test_main_argv_pollution.py:48-53` (test_doctor_dispatch_does_not_leak_sys_argv) — the comment "We inspected `sys.argv` from inside the mock by reading it before the `finally` restores it" is false for this test: the mock uses `return_value=0` and never reads argv. The inner-argv contract is actually verified in `test_doctor_dispatch_hides_doctor_token_from_inner_cli`; delete or reword the misleading comment.
- [P3] `tests/cli/test_main_argv_pollution.py:33` — `_import_dispatch_main`'s docstring claims a "fresh" import ensuring deterministic patching, but it performs a normal (cached) import with no reload; patching is deterministic only because `main()` lazy-imports at call time. Comment should say that instead.
- [P3] `aragora/__main__.py:15` (context, not a regression) — the stripped argv is largely ceremonial: `aragora.cli.doctor.main(validate_keys: bool = False)` never parses `sys.argv`, so `python -m aragora doctor --validate-keys` silently ignores the flag today. Pre-existing, but worth a follow-up since the new tests now enshrine the argv-hiding contract.

dogfood: yes

an0mium:

Head: ff7cccd (ff7cccd0e944aaf06c324cc69ffce0c451d82bfd), committed 2026-07-12T02:04:03Z.
PR: #9244.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9257, opened 2026-07-13
refactor(billing): invert external sink dependencies

scarmani:
## Summary
- invert billing budget-alert, notification, KM cost, and MFA audit dependencies through billing-owned protocols/providers registered from server startup
- replace the billing calibration `TYPE_CHECKING` agents dependency with local structural protocols
- move client-IP extraction to `aragora.utils.request_ip` with warning compatibility wrappers at both server middleware paths
- hand-shrink the seven fully cleared billing/security foundation-infrastructure baseline edges while preserving `frozen_from_ref`

## Architecture evidence
- live import-linter computation reports none of these edges: billing -> agents/connectors/knowledge/server and security -> connectors/gateway/server
- `aragora/billing` has no direct imports of agents, connectors, knowledge, server, audit, or notifications with TYPE_CHECKING imports included
- `aragora/knowledge/mound/adapters/cost_adapter.py` and `aragora/server/initialization.py` were intentionally not edited
- `docs/METRICS.md` and `aragora/module_tiers.yaml` were intentionally not regenerated (path-frozen structural drift)

## Validation
- `pytest -q tests/billing` (2666 passed)
- focused regression suite (706 passed)
- `bash scripts/test_tiers.sh smoke` (138 passed)
- `make test-smoke` (passed)
- changed-file Ruff format/lint (passed)
- changed-file mypy with CI flags (passed)
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` (passed)
- shrink-only baseline audit: exactly seven removals, zero additions, frozen ref preserved

## Known ambient drift
`python3 scripts/ci/check_import_contracts.py --layers foundation,infrastructure` still reports the pre-existing `aragora.utils -> aragora.caching` edge through the required `aragora.utils.redis_cache` compatibility shim. This PR does not grow/freeze that baseline and does not modify the caching shim.


scarmani:

Head: 9645c89 (9645c895db27658d102fecac3936399825131ba3), committed 2026-07-13T03:47:55Z.
PR: #9257.
Model family: claude

Verdict: PASS

Verified against head 9645c89 (not just the diff): the moved `extract_client_ip` is behavior-identical to the pre-refactor version (diffed vs `5fdddd090a~1`), `Notification` accepts the newly-passed `workspace_id` (aragora/notifications/models.py:58), `audit_admin(admin_id, action, **kwargs)` matches the sink call signature, Teams `tenant_id` lookup semantics survive the sink hop, the new startup exception tuple exactly matches the old `get_cost_tracker` KM-wiring tuple, `CostAdapter.__init__` is in-memory-only (no I/O to escape that tuple), and the missing-sink `RuntimeError` is caught and reported as a FAILED delivery. Audit-before-mutate in `approve_mfa_bypass` fixes the old partial-mutation-on-audit-failure bug, and no production callers of approve/revoke exist yet, so the fail-closed registration requirement breaks nothing today.

- [P3] `aragora/server/middleware/auth.py:44` and `user_auth.py:44` — module-level `warnings.warn(DeprecationWarning)` fires on *import of the module*, not on use of the deprecated symbol. `aragora/server/middleware/__init__.py` imports `user_auth` unconditionally, so every middleware import emits both warnings even for callers that never touch `extract_client_ip`. Benign at runtime (default filters, and the `try/except DeprecationWarning` guards error-mode), but semantically wrong and noisy in test output; drop the module-level warns and keep only the per-call shim warnings.
- [P3] First-party code still calls the deprecated shims: `aragora/server/middleware/impersonation.py:60,444,519`, `aragora/server/handlers/admin/billing.py:480`, `aragora/server/handlers/billing/core.py:692`. These now emit DeprecationWarning on live request paths (impersonation, admin billing); migrate them to `aragora.utils.request_ip` in this PR or a fast follow.
- [P3] `aragora/billing/notification_sink.py:57,80` — when no sink is registered, cost-anomaly and budget-runway notifications silently no-op at `logger.debug`. Previously these worked in any process where `aragora.notifications` was importable; now non-server entry points (CLI, workers, library use) drop them with near-zero visibility. Same for KM cost persistence since `get_cost_tracker()` no longer auto-wires `CostAdapter`. Intentional inversion and registered on both server startup paths, but consider a warn-once instead of debug so degraded deployments are diagnosable.

dogfood: yes

PR #9264, opened 2026-07-13
refactor(events): move gauntlet notification ownership

scarmani:
## Summary
- relocate the live `GAUNTLET_COMPLETE` notification reaction from the infrastructure event manager to the server interface subscriber
- preserve the registration name, payload, asynchronous scheduling, delivery call, and expected failure handling
- remove the unwired plan-notification mixin and barrel export without a compatibility shim
- keep the aggregate `aragora.events -> aragora.server` baseline unchanged while separately owned routes remain

## Validation
- `python3 -m pytest tests/events/test_cross_subscriber_registry.py tests/server/test_event_subscribers.py -q --timeout=120` (`50 passed`)
- `make lint`
- `ruff format --check aragora/ tests/ scripts/`
- `python3 -m mypy --ignore-missing-imports --follow-imports=skip --show-error-codes aragora/events/cross_subscribers/manager.py aragora/events/subscribers/__init__.py aragora/server/event_subscribers.py`
- `make test-smoke`
- `PYTEST_BIN="python3 -m pytest" bash scripts/test_tiers.sh smoke` (`138 passed`)
- grimp assertion: no `aragora.events -> aragora.notifications.service` chain; aggregate events-to-server baseline retained with five remaining routes

## Baseline notes
- the mission differential typecheck reports the same `new=104 > recorded env drift=66` result on this head and clean `origin/main`
- the foundation/infrastructure import checker reports the same ambient `aragora.utils -> aragora.caching` violation on this head and clean `origin/main`


scarmani:

Head: df4bfd9 (df4bfd91553197975dab42d0dbafe14654543a6a), committed 2026-07-13T09:57:59Z.
PR: #9264.
Model family: claude

Verdict: PASS

Verified statically (sandbox denied running pytest, so no test execution — findings below are from code inspection):

- Handler relocation is verbatim: same registration name (`gauntlet_to_notification`), same event type, same error-handling envelope, so name-keyed feature-flag gating via the manager is preserved. `StreamEventType` is imported at runtime in the new home (`aragora/server/event_subscribers.py:40`).
- The deleted `NotificationHandlersMixin` was genuinely dead: never inherited or registered anywhere; only reference was the package `__init__` re-export the PR also removes. Plan lifecycle notifications are independently served by `aragora/server/handlers/plans.py:_fire_plan_notification`, so no behavior loss.
- No orphaned emitters: the only `StreamEventType.GAUNTLET_COMPLETE` producer is server-tier (`aragora/server/stream/gauntlet_emitter.py:154`), which only runs where the interface-superset bootstrap wires `ServerEventSubscriber` — the domain-subset path never receives this event, so relocating the handler drops no reachable behavior.
- The new guard test's grep assertion holds: zero `aragora.notifications` references remain under `aragora/events/`.

- [P3] `aragora/server/event_subscribers.py:136` — `loop.create_task(coro)` is fire-and-forget without retaining a task reference; per asyncio docs the task can be garbage-collected before completion, silently dropping the notification. Pre-existing pattern moved verbatim (also present in the old manager code), but this move was the opportunity to fix it (e.g., a module-level task set with done-callback discard).
- [P3] `tests/server/test_event_subscribers.py:184` — the parametrized failure test only asserts "does not raise"; the ImportError case is artificial (raised at await time via `side_effect`, whereas real ImportError occurs at import inside the outer `try`). Both paths happen to route to the same outer handlers, so coverage is nominally correct but weaker than it looks.

dogfood: yes

PR #9166, opened 2026-07-10
fix(control-plane): narrow optional types

an0mium:
Main-red remediation slice for #9099.

## Scope

- narrows optional values at existing control-plane runtime contracts
- preserves runtime behavior and public signatures
- repairs the isolated pre-push optional-import aliases in the already claimed scheduler file
- does not modify tests, schemas, baselines, dependencies, workflows, readiness, settlement, or the active main-red halt

Pinned base: `3fe2e5cf561fc094221008d064040ac84625bd4e`
Exact head: `a8d10276b2e3ddab3dfd13949bda16a1fdef0e01`

## Typecheck evidence

Pinned Python 3.12.12 / mypy 2.2.0 full run:

- before: 2,634 errors, 330 notes
- after: 2,619 errors, 330 notes
- removed: exactly the 15 claimed `aragora/control_plane` diagnostics
- added: 0
- after-output SHA-256: `9f062aba63d7812b0fef4468319aeacf0e100f60d5151c729070010232216cb3`

The mypy 1.19.1 `typecheck-changed` pre-push hook also passes.

## Validation

- `python3 -m pytest tests/control_plane tests/integration/test_control_plane_redis.py tests/integration/test_control_plane_e2e.py tests/integration/test_control_plane_km_integration.py -q -p no:cacheprovider` — 1,640 passed, 12 skipped
- `ruff check <8 claimed files>`
- `ruff format --check <8 claimed files>`
- scoped pinned mypy on all eight claimed files
- full pinned mypy before/after identity comparison
- `pre-commit run --hook-stage pre-push typecheck-changed --files <8 claimed files>`
- `bash scripts/automation_pr_preflight.sh origin/main HEAD`
- commit and push hooks

## Settlement boundary

Draft only. This PR does not re-arm main, settle the campaign, authorize merge, or alter the active halt.

## Current-base restack (2026-07-11)

- Merged current `origin/main` `a0f0c417916203d7ed0232ebd178a318495d1c23` normally; no rebase or history rewrite.
- Exact restack head: `38c13a30ae8fc3c5836d7ca3aa5a12899b0965ee`.
- Canonical Python 3.12.12 / mypy 2.2.0 / mypy-baseline 0.7.4 / PyJWT 2.13.0 proof: `2634 -> 2619` errors, exactly 15 removed identities in the eight claimed `aragora/control_plane` files, and 0 added identities.
[...]

PR #9265, opened 2026-07-13
refactor(layers): invert channel boundary contracts

scarmani:
## Source issue
P4a corrective inversion for the remaining channel-boundary routes found by the final shim sweep. This does not edit the path-frozen `aragora/control_plane/channels.py`, adopt PR #9166, or shrink aggregate import baselines.

## Behavior
- Adds an events-owned notification contributor hook and a side-effect-free control-plane adapter, composed before sequential and parallel control-plane startup.
- Adds an observability-owned SLO notification sink provider and extends `ControlPlaneSLOAlertSink` composition with a direct dispatcher adapter.
- Preserves notification event precedence, every `NotificationEventType` value, SLO formatting, priority mapping, metadata, dynamic dispatcher replacement, and fail-soft delivery.
- Removes the targeted Grimp chains from `events.registry` and `observability.slo`.

## Validation
- `make lint` (pass)
- `python3 -m ruff format --check aragora/ tests/ scripts/` (10,706 files formatted)
- changed-file mypy (7 source files, pass)
- targeted pytest (125 passed)
- 5-seed randomized targeted sweep (all passes)
- `AWS_CONFIG_FILE=/dev/null AWS_SHARED_CREDENTIALS_FILE=/dev/null AWS_EC2_METADATA_DISABLED=true make test-smoke` (pass)
- `PYTEST_BIN="python3 -m pytest" bash scripts/test_tiers.sh smoke` (138 passed)
- targeted Grimp chains (all `None`)
- `python3 scripts/ci/check_import_contracts.py --layers infrastructure` (pass)
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` (pass)

The full import-contract and differential-typecheck wrappers are currently red on clean `origin/main` for unrelated pre-existing violations. This branch adds no changed-file mypy errors and the affected infrastructure import-contract layer passes.

## Risks
Startup composition is explicit and adapter imports remain side-effect free. The generated module metrics files are intentionally unchanged because `docs/METRICS.md` and `aragora/module_tiers.yaml` are path-frozen by open PRs.

### DIFF
--- a/aragora/control_plane/event_registry.py
+++ b/aragora/control_plane/event_registry.py
@@ -1,0 +1,25 @@
+"""Control-plane adapter for the events-owned notification registry contract."""
+
+from __future__ import annotations
+
+from aragora.control_plane.channels import NotificationEventType
+from aragora.events.registry import (
+    register_notification_event_contributor as register_events_contributor,
+)
+
+
+def get_notification_event_names() -> tuple[str, ...]:
+    """Return every control-plane notification event name."""
+    return tuple(event_type.value for event_type in NotificationEventType)
+
+
+def register_notification_event_contributor() -> bool:
+    """Compose notification event discovery into the events registry."""
+    register_events_contributor(get_notification_event_names)
+    return True
+
+
+__all__ = [
+    "get_notification_event_names",
+    "register_notification_event_contributor",
+]
--- a/aragora/control_plane/slo_alert_sink.py
+++ b/aragora/control_plane/slo_alert_sink.py
@@ -6,10 +6,7 @@ import logging
 from typing import Any
 
 from aragora.control_plane.channels import (
-    ChannelConfig,
-    NotificationChannel,
     NotificationEventType,
-    NotificationManager,
     NotificationPriority,
 )
 from aragora.control_plane.notifications import get_default_notification_dispatcher
@@ -35,6 +32,12 @@ class ControlPlaneSLOAlertSink:
 
     def _get_manager(self) -> Any:
         if self._manager is None:
+            from aragora.control_plane.channels import (
+                ChannelConfig,
+                NotificationChannel,
+                NotificationManager,
+            )
+
             manager = NotificationManager()
             manager.add_channel(
                 ChannelConfig(
@@ -56,6 +59,11 @@ class ControlPlaneSLOAlertSink:
         metadata: dict[str, Any],
     ) -> Any:
         """Deliver an observability alert through the channel manager."""
+        from aragora.control_plane.channels import (
+            NotificationEventType,
+            NotificationPriority,
+        )
+
         return await self._get_manager().notify(
             event_type=NotificationEventType(event_type),
             title=title,
--- a/aragora/control_plane/slo_alert_sink.py
+++ b/aragora/control_plane/slo_alert_sink.py
@@ -3,10 +3,25 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING, Any
+from typing import Any
 
-if TYPE_CHECKING:
-    from aragora.observability.slo_alert_bridge import SLOAlertConfig
+from aragora.control_plane.channels import (
+    ChannelConfig,
+    NotificationChannel,
+    NotificationEventType,
+    NotificationManager,
+    NotificationPriority,
+)
+from aragora.control_plane.notifications import get_default_notification_dispatcher
+from aragora.observability.slo import (
+    SLOBreach,
+    SLONotificationSink,
+    register_slo_notification_sink_provider,
+)
+from aragora.observability.slo_alert_bridge import (
+    SLOAlertConfig,
+    register_channel_alert_sink,
+)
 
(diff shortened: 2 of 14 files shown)


scarmani:

Head: c446082 (c44608205818586a63fee673df77b4331362d43a), committed 2026-07-13T11:53:20Z.
PR: #9265.
Model family: claude

Verdict: PASS

- [P3] `aragora/events/registry.py:365` — Notification events are now only registered when a contributor has been composed, and composition lives solely in server startup (`_init_all_components` / `parallel_init`). Any non-server entry point (CLI, MCP, direct library use) constructing `EventRegistry`/`get_event_registry()` silently loses all `NotificationEventType` entries versus the old self-registering behavior. I verified there are currently zero production callers of `get_event_registry()` outside tests, so this is latent rather than live — but it's an invisible contract: nothing warns if a future consumer reads the registry before composition.
- [P3] `aragora/server/startup/control_plane.py:18` — `register_notification_event_contributor()` has no effect on an already-instantiated registry singleton (contributor is only read in `EventRegistry.__init__`, and registration doesn't reset/rebuild). The guard test `test_composes_events_before_control_plane_startup` only checks textual ordering inside `_init_all_components`; it can't catch an import in startup phases 0–2 instantiating the registry first. Consider having the register call `reset_event_registry()` (or assert the singleton is unbuilt) to make the ordering constraint self-enforcing.
- [P3] `aragora/events/registry.py:370` — the contributor call catches `(ImportError, RuntimeError, TypeError, ValueError)`; a contributor raising anything else (e.g., `AttributeError`, `KeyError`) now crashes `EventRegistry.__init__`, a failure mode the old direct-import code didn't have. Given contributors are third-party-composable by design, a broader fail-soft catch here would be safer.
- [P3] `aragora/observability/slo.py:1323` — the `sink is None` branch still logs "Notification dispatcher not configured", but after this inversion `None` means "no sink provider registered / provider declined" — misleading during triage when a dispatcher *is* configured but startup composition didn't run. Same stale wording is baked into `tests/observability/test_slo.py::test_missing_provider_is_fail_soft`.

Non-finding notes verified: both startup paths compose the SLO sink provider via the existing `register_observability_sinks()` (called from `init_prometheus_metrics`, present in sequential and parallel paths) with ImportError handled by the caller, so `slo_alert_sink.py`'s move to top-level imports doesn't regress the old graceful `return False`; the notification payload/priority mapping is byte-identical to the pre-refactor dispatch (tests pin it); the `test_parallel.py` patch target works because `parallel_init` imports the symbol at call time.

dogfood: yes

scarmani:

Head: 11e1654 (11e16545ae6dbafcbcbdd663ca9e96f18c183fb2), committed 2026-07-13T12:29:21Z.
PR: #9265.
Model family: claude

Verdict: PASS

- [P3] `aragora/events/registry.py:365` + `aragora/events/registry.py:508` — Notification events are now only present when startup composition runs, and composition does not backfill: `get_event_registry()` caches the singleton at first call and `register_notification_event_contributor()` doesn't reset it, so any process that builds the registry before (or without) `register_control_plane_event_contributors()` — CLI, gateway, MCP, library use — permanently lacks the 19 `NotificationEventType` entries where they were previously always registered via lazy import. I found no production consumer of the registry outside `aragora/events/` today, so this is latent rather than live, but a `reset_event_registry()` (or generation bump) inside the register call would make composition order-independent.
- [P3] `aragora/server/startup/__init__.py:436` / `aragora/server/startup/parallel.py:841` — `register_control_plane_event_contributors()` is called bare with a function-level import of `aragora.control_plane.event_registry` → `aragora.control_plane.channels`; an ImportError there now aborts server startup where the old in-registry `except ImportError: logger.debug(...)` degraded gracefully. Consistent with the adjacent bare `register_webhook_store()` call, so noting only.
- [P3] `aragora/events/registry.py:371` — the contributor guard catches `(ImportError, RuntimeError, TypeError, ValueError)`; a contributor raising anything else (e.g. `AttributeError`) now fails `EventRegistry()` construction entirely, a failure mode that couldn't exist when the code iterated the enum directly. Fine for the single first-party contributor, but the contract is nominally open.

Verified statically against head 11e1654: the adapter's module-level imports are cycle-free by design (post-inversion, `observability.slo`/`slo_alert_bridge` no longer import control_plane, confirmed by the source-scan tests), `register_observability_sinks()` retains the ImportError fail-soft that the adapter's `register_slo_alert_sink()` dropped, `create_notification_callback`'s existing `except (OSError, ConnectionError, RuntimeError)` covers the new provider-raises path the tests assert, `register_channel_alert_sink(None)` is a valid signature, and per-call resolution (no dispatcher/sink caching) preserves the patchability fixed in 11e16545ae. The channel-bridge sink and the direct notification sink serve distinct pre-existing delivery paths (bridge vs. `create_notification_callback`), so dual registration introduces no double-delivery. Note: I could not execute the test suite in this session (Bash execution denied by permission mode), so test results are from static inspection only.

dogfood: yes

PR #9279, opened 2026-07-14
fix(debate): bootstrap cold arena bridge subscribers

scarmani:
## Source

Structural Excellence epic #8257, P4a scrutiny remediation for `VAL-P4A-015`.

## Behavior

- Makes cold `create_arena_bridge(EventBus())` obtain and preserve the manager returned by `bootstrap_debate_event_subscribers()`.
- Keeps the factory domain-only, without importing workflow or server composition homes.
- Preserves exact identity and registration scope for explicitly injected `CrossSubscriberManager` instances, including falsy manager implementations.
- Adds fresh-process cold-bootstrap and explicit-injection regression coverage.

## Validation

- `python3 -m pytest tests/debate/test_arena_bridge.py -q` -> `14 passed`
- `python3 -m pytest tests/debate/test_arena_bridge.py -q -k 'test_create_arena_bridge_cold_bootstraps_domain_subscribers or test_create_arena_bridge_preserves_explicit_manager'` -> `2 passed`
- `python3 /tmp/p4val/t1.py aragora/debate/arena_bridge.py aragora.workflow` -> `PASS`
- `python3 /tmp/p4val/t1.py aragora/debate/arena_bridge.py aragora.server` -> `PASS`
- `mypy --ignore-missing-imports --follow-imports=skip --show-error-codes aragora/debate/arena_bridge.py` -> no issues
- `bash $MISSION/scripts/typecheck_differential.sh` -> `TYPECHECK PASS (new=103 <= 108)`
- `make lint` and `ruff format --check aragora/ tests/ scripts/` -> passed
- `make test-smoke` -> Core, Server, and Memory imports passed
- `PYTEST_BIN="python3 -m pytest" bash scripts/test_tiers.sh smoke` -> `138 passed`
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` -> passed

## Risk

Tier 1 behavior fix. The only production behavior change is that the public factory now performs the existing debate-domain bootstrap before connecting. Direct explicit injection remains narrow and unchanged apart from correctly accepting falsy manager objects.

## Reviewer context

The Arena bridge remains at its canonical no-shim domain path, `aragora.debate.arena_bridge`. This repair does not restore any `aragora.events` compatibility surface and does not invoke the application/interface superset composition roots.

### DIFF
--- a/aragora/debate/arena_bridge.py
+++ b/aragora/debate/arena_bridge.py
@@ -81,7 +81,9 @@ class ArenaEventBridge:
             cross_manager: Optional CrossSubscriberManager (uses global if not provided)
         """
         self._event_bus = event_bus
-        self._cross_manager = cross_manager or get_cross_subscriber_manager()
+        self._cross_manager = (
+            cross_manager if cross_manager is not None else get_cross_subscriber_manager()
+        )
         self._connected = False
 
     def connect_to_cross_subscribers(self) -> None:
@@ -172,7 +174,10 @@ def create_arena_bridge(event_bus: EventBus) -> ArenaEventBridge:
     Returns:
         Connected ArenaEventBridge
     """
-    bridge = ArenaEventBridge(event_bus)
+    from aragora.debate.event_subscribers import bootstrap_debate_event_subscribers
+
+    cross_manager = bootstrap_debate_event_subscribers()
+    bridge = ArenaEventBridge(event_bus, cross_manager)
     bridge.connect_to_cross_subscribers()
     return bridge
 
--- a/tests/debate/test_arena_bridge.py
+++ b/tests/debate/test_arena_bridge.py
@@ -1,8 +1,13 @@
 """Tests for ArenaEventBridge integration."""
 
-import pytest
 from datetime import datetime, timezone
+from pathlib import Path
+import subprocess
+import sys
+import textwrap
 from unittest.mock import MagicMock, patch
+
+import pytest
 
 from aragora.debate.arena_bridge import (
     ArenaEventBridge,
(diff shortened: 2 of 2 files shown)


scarmani:

Head: e1b586f (e1b586f6cf5cacba03f6da31165a8c2e8d90ce3c), committed 2026-07-14T05:42:52Z.
PR: #9279.
Model family: claude

Verdict: PASS

Verified against the worktree at head e1b586f: `bootstrap_debate_event_subscribers()` returns the same global singleton (`bootstrap()` → `get_cross_subscriber_manager()`, `__init__.py:89`), application is idempotent per manager via `_applied_subscribers` and domain-scoped via `include_names`, the deferred in-function import avoids a module-level cycle, and the change exactly mirrors the already-proven production path in `orchestrator_memory.py:321-323` (`init_cross_subscriber_bridge`). The `or` → `is not None` fix is strictly safer (no real falsy managers exist today, so no regression). The subprocess test's monkeypatch works precisely because the factory imports at call time, and the test would fail loudly if that import were hoisted — it pins the contract.

- [P3] `aragora/debate/arena_bridge.py:179` — `create_arena_bridge` can now propagate the bootstrap completeness `RuntimeError` (`event_subscribers.py:458-462`), whereas the sibling production path `orchestrator_memory.py:329` catches `RuntimeError` and degrades to a warning + `None`. Previously the factory could not raise post-import. Loud-fail is defensible for a factory, but the inconsistency between the two entry points is worth a deliberate choice.
- [P3] `aragora/events/cross_subscribers/manager.py:173-181` — `apply_registered_subscribers` is check-then-act with no lock, and `register()` (manager.py:627) appends unconditionally; two threads cold-calling `create_arena_bridge` concurrently could double-apply a home, permanently duplicating handler dispatch. Pre-existing gap, but this PR adds a new reachable trigger path.
- [P3] `tests/debate/test_arena_bridge.py:267` — `test_create_arena_bridge_preserves_explicit_manager` never calls `create_arena_bridge` (which accepts no manager); it exercises the `ArenaEventBridge` constructor. Misleading name.
- [P3] `tests/debate/test_arena_bridge.py:228-232` — assertions on private internals (`bridge._cross_manager`, `manager._applied_subscribers`) couple the test to implementation details; `get_stats()` / behavior-level checks (already present) would suffice.

Note: I could not execute the new tests (Bash test runs are permission-denied in this session), so the pass/fail claim for the subprocess test rests on static verification — `emit_sync` kwargs signature (`event_bus.py:318`), `ensure_registered` (`security_response.py:210`), `events_processed` stat key (`admin.py:72`), and absence of application-tier home imports in the involved modules all check out.

dogfood: yes

PR #9289, opened 2026-07-14
fix(stream): fail closed on escalated alert brake

scarmani:
## Summary

- route critical `emit_alert_event(event_type=\"escalated\")` calls through the production superset subscriber bootstrap before WebSocket broadcast
- add strict named cross-subscriber delivery that invokes one registered brake handler once, without sampling, retries, or circuit-breaker suppression, and requires explicit success confirmation
- track live base and enhanced workflow engines process-wide so `WorkflowEngine.pause_all()` requests termination exactly once on every live instance without latching future executions
- preserve broadcast-only behavior for noncritical escalations and all non-escalated alert helpers

## Contract

Implements mission feature `p4a-alert-escalated-workflow-brake` and `VAL-P4A-016`.

The production helper now fails closed when the workflow-brake wiring is missing, duplicated, disabled, filtered, sampled, throwing, or unable to confirm delivery. Broadcast occurs only after confirmed brake delivery. Prior circuit state never suppresses the required attempt, while outcomes and metrics remain recorded.

## Verification

- exact `VAL-P4A-016` production-helper regressions: 2 passed
- focused production alert/wiring suites: 168 passed
- workflow engine regression suites: 255 passed
- open-circuit required-delivery regression: passed
- concurrent live base/enhanced engine brake regression: passed
- changed-file mypy: success
- changed-file `ruff check`: pass
- changed-file `ruff format --check`: pass
- code-quality ratchet: pass (900/900 `except Exception`)
- alert brake import and multi-instance runtime canaries: pass
- `make test-smoke`: pass
- pre-commit and pre-push hooks: pass
- `bash scripts/automation_pr_preflight.sh origin/main HEAD`: pass
- exact-head Claude and OpenAI model quorum on `9643f5e`: pass

## Invariants preserved

- alert event payload shape and `StreamEventType.ALERT_ESCALATED` WebSocket delivery
- non-escalated alert behavior
- noncritical escalation broadcast behavior
- ordinary cross-subscriber sampling, retry, filtering, circuit-breaker, and fail-soft semantics
- existing dispatch metrics, latency statistics, circuit-breaker metrics, and SLO recording
- E5 post-debate registration and dispatch behavior
- future workflow executions after a point-in-time emergency brake

## Risk and non-goals

Tier 1 bounded behavior repair. This PR does not change E5 registration, live merge-gate logic, scheduler admission control, or general workflow concurrency semantics.

### DIFF
--- a/aragora/events/cross_subscribers/dispatch.py
+++ b/aragora/events/cross_subscribers/dispatch.py
@@ -306,7 +306,7 @@ class DispatchMixin:
         if filter_func is not None:
             try:
                 accepted = filter_func(event)
-            except Exception as exc:  # noqa: BLE001 - fail closed on any filter failure
+            except (RuntimeError, TypeError, AttributeError, ValueError, KeyError) as exc:
                 stats.events_failed += 1
                 raise RuntimeError(
                     f"Required cross-subscriber '{handler_name}' filter failed"
--- a/aragora/events/cross_subscribers/dispatch.py
+++ b/aragora/events/cross_subscribers/dispatch.py
@@ -86,6 +86,8 @@ except ImportError:
 
 logger = logging.getLogger(__name__)
 
+CrossSubscriberHandler = Callable[[StreamEvent], object]
+
 
 def _default_async_event_types() -> set:
     """Default event types for async dispatch."""
@@ -112,7 +114,7 @@ class DispatchMixin:
     """
 
     # Mixin attribute declarations (provided by the composed class)
-    _subscribers: dict[StreamEventType, list[tuple[str, Callable[[StreamEvent], None]]]]
+    _subscribers: dict[StreamEventType, list[tuple[str, CrossSubscriberHandler]]]
     _stats: dict[str, SubscriberStats]
     _filters: dict[str, Callable[[StreamEvent], bool]]
     _circuit_breaker: CircuitBreaker
@@ -159,11 +161,8 @@ class DispatchMixin:
                 continue
 
             # Get retry config
-            retry_config = (
-                self._stats[name].retry_config
-                if name in self._stats and self._stats[name].retry_config
-                else self._default_retry_config
-            )
+            configured_retry = self._stats[name].retry_config if name in self._stats else None
+            retry_config = configured_retry or self._default_retry_config
 
             # Execute handler with timing, retry, and metrics
             start_time = time.time()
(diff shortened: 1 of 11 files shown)


scarmani:

Head: 62d96bb (62d96bb3c361add925a101ef2f50628ff1a7af27), committed 2026-07-14T23:07:22Z.
PR: #9289.
Model family: claude

Verdict: PASS

Verified before ruling: `StructuredLogger.critical` accepts the `reason=` kwarg (aragora/observability/logging.py:365), `bootstrap_event_subscribers` exists, is idempotent, and hard-fails if the brake handler is missing; registry application is exactly-once per manager (`_applied_subscribers`), so `dispatch_required`'s len==1 check holds; `EnhancedWorkflowEngine` subclasses `WorkflowEngine` so the brake methods apply to v2; the autonomous emitter only broadcasts to WS clients, so there is no double brake dispatch; there are no in-repo callers of `emit_alert_event`, and previously the brake was a silent no-op (`WorkflowEngine` had neither `pause_all` nor `emergency_stop`), so this PR makes the safety valve real. All 201 tests in the four changed test files pass at head.

- [P3] `aragora/workflow/event_subscribers.py:262` — unreachable trailing `return False`: every path in the preceding try/except already returns. Dead code; drop it.
- [P3] `aragora/workflow/event_subscribers.py:236` — `data.get("reason", data.get("message", ""))[:200]` raises `TypeError` if a producer passes `reason=None` (or any non-sliceable), and it runs before the severity gate and outside the handler's try. Under `dispatch_required` that now aborts the entire critical-alert emit (no brake latch, no broadcast) where previously the exception was swallowed by ordinary dispatch. Coerce with `str(... or "")` before slicing.
- [P3] `aragora/workflow/event_subscribers.py:258` — brake-failure root cause is logged at `debug` and not chained; `dispatch_required` then raises a generic "handler did not confirm delivery". On a safety-critical path the original traceback is lost. Log at `error`/`exception` or re-raise so the chained cause survives.
- [P3] `aragora/events/cross_subscribers/dispatch.py:296` — `dispatch_required` refuses delivery while the shared circuit breaker is open. If the brake fails repeatedly during an incident storm (e.g., unwired window), the breaker opens and later critical escalations are rejected without even attempting the brake until cool-down — the safety action is least available when alerts are storming. Consider bypassing (or only reading, never honoring) the breaker on the required path.
- [P3] `aragora/workflow/engine.py:420` — semantics change beyond the brake: user-initiated cancel (`request_termination("User requested")` from `aragora/server/handlers/workflows/execution.py:444`) now yields `success=False, error="User requested"` where completed-then-cancelled runs previously reported `success=True`. Arguably more correct, but it's an observable API change for workflow-cancel consumers; worth a changelog note.
- [P3] `aragora/workflow/engine.py:1142` — the latch has no operator-facing release: `reset_emergency_brake()` is only reachable programmatically and `reset_workflow_engine()` is test-only, so one critical escalation disables all workflow execution until process restart. If that's intended, document it; otherwise expose an admin/HTTP unlatch. Relatedly, `emit_alert_event` is a public `aragora.server.stream` export that previously never raised and now raises `RuntimeError` on critical escalations when wiring/brake fails — intentional fail-closed, but external alert loops should be warned.

dogfood: yes

scarmani:

Head: 62d96bb (62d96bb3c361add925a101ef2f50628ff1a7af27), committed 2026-07-14T23:07:22Z.
PR: #9289.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

Head: 9643f5e (9643f5e09cdc65686c704d4fdbf34e1b5e5de600), committed 2026-07-14T23:44:46Z.
PR: #9289.
Model family: claude

Verdict: PASS

- [P3] `aragora/server/stream/autonomous_stream.py:214-219` — Fail-closed ordering suppresses the broadcast entirely: when `dispatch_required` raises, `emitter.emit_sync(event)` is never reached, so stream/dashboard consumers never see the critical escalation; the only signal is the propagated RuntimeError, which no in-repo caller handles (`emit_alert_event` has zero internal callers but is exported in `aragora.server.stream.__all__`, so this is a public-API behavior change from fire-and-forget to raising). The design is deliberate and tested, but consider broadcasting in a `finally` or emitting a degraded "brake-failed" event so operators retain visibility when both the brake and the alert path go dark simultaneously.
- [P3] `aragora/workflow/engine.py:1140-1148` — The brake is not a latch: `execute()` resets `_should_terminate`/`_termination_reason` at start (engine.py:402-404, pre-existing), so any engine re-executed after `pause_all` — or any engine constructed after the brake, as the new test explicitly asserts — runs unimpeded. There is also a narrow race where an `execute()` entered concurrently with `pause_all` clears the just-set flag. Fine if the contract is "stop in-flight executions only," but the "Emergency brake" naming and the `latch alert brake across engines` commit message suggest stronger semantics than delivered; worth documenting the boundary on `pause_all`.
- [P3] `aragora/workflow/engine.py:1147` — `logger.critical("Workflow emergency brake activated: " + reason)` uses eager string concatenation of caller-supplied text instead of the repo's lazy `logger.critical("...: %s", reason)` pattern used two lines up in `request_termination`.

Verified against head 9643f5e: `SubscriberStats` has all fields `dispatch_required` touches (`events_skipped`, `record_latency`, `min_latency_ms=inf` init); `datetime` is already imported in dispatch.py; `bootstrap_event_subscribers` exists, is idempotent, and asserts the `alert_escalated_to_workflow_brake` handler is wired (registration name matches event_subscribers.py:274); `from __future__ import annotations` is present in engine.py so the self-referential `ClassVar[weakref.WeakSet[WorkflowEngine]]` annotation is safe; the emit-side severity gate (`new_severity` falling back to `severity`) matches the handler's precedence exactly, so no spurious fail-closed mismatch; the `new_severity`-first flip fixes the real pre-PR bug where an escalation to critical with a lower prior `severity` never braked; `pause_all` copies the WeakSet under the lock before iterating; and the required-dispatch path correctly bypasses the open circuit breaker while still recording outcomes (covered by the new test). Note I could not execute the test suite — the sandbox denied `pytest` — so test greenness is asserted from static reading only.

dogfood: yes

scarmani:

Head: 9643f5e (9643f5e09cdc65686c704d4fdbf34e1b5e5de600), committed 2026-07-14T23:44:46Z.
PR: #9289.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9292, opened 2026-07-15
fix(observability): canonicalize active debate gauge

scarmani:
## Source

- Structural Excellence feature `p4a-active-debates-canonical-state`
- Fulfills `VAL-P4A-018`

## Behavior

- Uses `aragora.observability.server_metrics.ACTIVE_DEBATES` as the only stored active-debate count.
- Keeps `aragora.observability.metrics`, `aragora.observability.metrics.debate`, and deprecated `aragora.server.metrics` compatibility shapes on the same gauge object.
- Exposes one read-only Prometheus collector backed by that canonical gauge and reconciles a stale same-name collector without creating duplicate storage.
- Makes `MetricsCollector._collect_debate_metrics` read the canonical gauge through its public `get()` method.
- Preserves one `aragora_active_debates` family and a valid terminal newline in the combined server metrics export.
- Serializes canonical collector reconciliation and resets canonical state in observability test cleanup.
- Intentionally excludes the separate observability provider inversion.

## Validation

- Focused canonical tests, normal and `ARAGORA_OFFLINE=1` -> 2 passed in each mode
- No-op cleanup ordering regression -> 3 passed
- Expanded related suite under seeds 1, 42, 100, 2024, 12345 -> 309 passed per seed
- `make lint` -> passed
- Changed-file Ruff and formatting -> passed
- Changed-file mypy -> no issues in 4 source files
- Differential typecheck -> PASS, new=104 <= 108 ambient threshold
- `make test-smoke` -> Core/Server/Memory imports passed
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` -> ok

## Risk

Tier 2 observability-internal change. The compatibility import paths and metric name remain stable; the Prometheus adapter is read-only, mirrors the canonical lifecycle state, and has exact lifecycle/export/alerting coverage.

### DIFF
--- a/aragora/observability/alerting.py
+++ b/aragora/observability/alerting.py
@@ -629,11 +629,9 @@ class MetricsCollector:
     async def _collect_debate_metrics(self, snapshot: MetricsSnapshot) -> None:
         """Collect debate-related metrics."""
         try:
-            from aragora.observability.metrics.debate import ACTIVE_DEBATES
+            from aragora.observability.server_metrics import ACTIVE_DEBATES
 
-            if ACTIVE_DEBATES is not None:
-                # Prometheus Gauge value
-                snapshot.active_debates = int(ACTIVE_DEBATES._value.get())
+            snapshot.active_debates = int(ACTIVE_DEBATES.get())
         except (ImportError, AttributeError):
             pass
 
--- a/aragora/observability/metrics/debate.py
+++ b/aragora/observability/metrics/debate.py
@@ -8,6 +8,7 @@ performance monitoring, phases, and consensus outcomes.
 from __future__ import annotations
 
 import logging
+from threading import Lock
 import time
 from contextlib import contextmanager
 from typing import Any
@@ -34,6 +35,7 @@ DEBATE_EARLY_TERMINATION_TOTAL: Any = None
 DEBATE_STABILITY_SCORE: Any = None
 
 _initialized = False
+_active_debates_collector_lock = Lock()
 
 
 class _ActiveDebatesCollector:
@@ -54,12 +56,13 @@ class _ActiveDebatesCollector:
 
 def _ensure_active_debates_collector(registry: Any) -> None:
     """Register one read-only Prometheus view of the canonical gauge."""
-    existing = registry._names_to_collectors.get(_CANONICAL_ACTIVE_DEBATES.name)
-    if getattr(existing, "_canonical_gauge", None) is _CANONICAL_ACTIVE_DEBATES:
-        return
-    if existing is not None:
-        registry.unregister(existing)
-    registry.register(_ActiveDebatesCollector())
+    with _active_debates_collector_lock:
+        existing = registry._names_to_collectors.get(_CANONICAL_ACTIVE_DEBATES.name)
+        if getattr(existing, "_canonical_gauge", None) is _CANONICAL_ACTIVE_DEBATES:
+            return
+        if existing is not None:
+            registry.unregister(existing)
+        registry.register(_ActiveDebatesCollector())
 
 
 def init_debate_metrics() -> None:
--- a/aragora/observability/metrics/debate.py
+++ b/aragora/observability/metrics/debate.py
@@ -14,6 +14,9 @@ from typing import Any
 from collections.abc import Generator
 
 from aragora.observability.metrics.base import NoOpMetric, get_metrics_enabled
+from aragora.observability.server_metrics.api import (
+    ACTIVE_DEBATES as _CANONICAL_ACTIVE_DEBATES,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -25,12 +28,38 @@ AGENT_PARTICIPATION: Any = None
 SLOW_DEBATES_TOTAL: Any = None
 SLOW_ROUNDS_TOTAL: Any = None
 DEBATE_ROUND_LATENCY: Any = None
-ACTIVE_DEBATES: Any = None
+ACTIVE_DEBATES: Any = _CANONICAL_ACTIVE_DEBATES
 CONSENSUS_RATE: Any = None
 DEBATE_EARLY_TERMINATION_TOTAL: Any = None
 DEBATE_STABILITY_SCORE: Any = None
 
 _initialized = False
+
+
+class _ActiveDebatesCollector:
+    """Prometheus adapter that reads the canonical active-debates gauge."""
+
+    _canonical_gauge = _CANONICAL_ACTIVE_DEBATES
+
+    def collect(self) -> Generator[Any, None, None]:
+        from prometheus_client.core import GaugeMetricFamily
+
+        metric = GaugeMetricFamily(
+            _CANONICAL_ACTIVE_DEBATES.name,
+            _CANONICAL_ACTIVE_DEBATES.help,
+        )
+        metric.add_metric([], _CANONICAL_ACTIVE_DEBATES.get())
+        yield metric
+
+
+def _ensure_active_debates_collector(registry: Any) -> None:
+    """Register one read-only Prometheus view of the canonical gauge."""
+    existing = registry._names_to_collectors.get(_CANONICAL_ACTIVE_DEBATES.name)
+    if getattr(existing, "_canonical_gauge", None) is _CANONICAL_ACTIVE_DEBATES:
+        return
+    if existing is not None:
+        registry.unregister(existing)
+    registry.register(_ActiveDebatesCollector())
 
 
 def init_debate_metrics() -> None:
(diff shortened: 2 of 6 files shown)


scarmani:

Head: 6f62ba7 (6f62ba74ebc3022131414f797e985844e1b96b21), committed 2026-07-15T03:11:59Z.
PR: #9292.
Model family: claude

Verdict: PASS

Review basis: static analysis of the diff against repo context at PR head 6f62ba7 (checked the custom `Gauge` API, import graph, all `ACTIVE_DEBATES` call sites, the `/metrics` endpoint, and the observability conftest). Test execution was denied by the session's permission mode, so I could not run the new tests locally.

Verified sound:
- The canonical `server_metrics` Gauge exposes `get/set/inc/dec/name/help/collect` (`server_metrics/types.py:64-101`), so both the new alerting read (`alerting.py:634`) and the `_ActiveDebatesCollector` adapter contract hold.
- No import cycle: `server_metrics` submodules import only relatively plus `aragora.exceptions` (stdlib-only), so the new module-level import in `metrics/debate.py` is safe in both directions.
- No double-count: `metrics.debate.track_debate` / `increment_active_debates` have zero production callers; the orchestrator's inc (`orchestrator.py:1251`) / dec (`orchestrator_runner.py:1297`) pair is the only live writer besides `track_debate_execution`, and they never wrap the same debate.
- This is a real bugfix: alerting previously read a prometheus gauge nobody incremented; the export dedupe filter also fixes a pre-existing duplicate-metric-family in the combined `generate_metrics()` exposition. The `tests/observability/conftest.py` autouse cleanup resolves the test-ordering hazard I initially suspected around the `_initialized` guard.

Findings:
- [P3] `aragora/observability/server_metrics/export.py:198-206` — `splitlines()`/`"\n".join` strips the prometheus blob's trailing newline, so when the observability section is last, `generate_metrics()` output no longer terminates with a line feed as the Prometheus text-format v0.0.4 spec requires. Non-blocking only because no in-repo endpoint serves this function (`metrics_endpoint.py:388` uses `generate_latest` directly). Relatedly, the sample filter only matches unlabeled samples (`"aragora_active_debates "` with a space) — correct for the label-less canonical gauge, brittle if a labeled variant ever appears.
- [P3] `aragora/observability/metrics/debate.py:56-62` (`_ensure_active_debates_collector`) — check-then-act on the registry is not atomic; a concurrent first `_ensure_init()` can race past the existing-collector check and `registry.register()` then raises `ValueError`, which escapes since `init_debate_metrics` only catches `ImportError`. Same risk class as the pre-existing `_get_or_create_*` helpers and init is normally single-threaded at startup, hence non-blocking. Also relies on the private `_names_to_collectors` API (an established pattern in this module).
- [P3] `aragora/observability/metrics/debate.py:205` — `_init_noop_metrics` now binds `ACTIVE_DEBATES` to the real shared gauge, so e.g. `tests/observability/test_observability.py:133-134` (`.dec(); .dec(2)`) permanently leaves the process-global canonical gauge at −3.0 (the conftest cleanup unregisters prometheus collectors but never resets the custom gauge's value). Nothing fails today — the new file's autouse fixture guards its own tests and other suites assert deltas — but any future absolute-value assertion elsewhere will flake by test ordering.
- [P3] `tests/observability/test_active_debates_canonical.py:69-97` — `test_observability_initialization_registers_single_collector` silently depends on `get_metrics_enabled()` being true (forced false under `ARAGORA_OFFLINE`, per `config.py:170-171`); in offline shards the init takes the noop path, the stale gauge is never replaced, and the test fails. Also the second `init_debate_metrics()` call early-returns on the `_initialized` guard, so re-init idempotence is only trivially exercised.

dogfood: yes

scarmani:

Head: 9ea79e1 (9ea79e198ffc758474cbe31906b31ae49844cece), committed 2026-07-15T03:57:36Z.
PR: #9292.
Model family: claude

Verdict: PASS

The change fixes a real split-brain bug: the orchestrator's inc/dec went to the canonical `server_metrics` gauge (via the deprecated `aragora.server.metrics` re-export) while alerting and the prometheus exposition read a separate, never-incremented gauge in `metrics/debate.py` that stayed at 0. I verified on head 9ea79e1: no import cycle (`server_metrics` submodules never import back into `observability.metrics`), the custom `Gauge` dataclass supports every operation used (`get/set/inc/dec/name/help/collect`), the canonical gauge is rendered exactly once by `generate_metrics()` (own gauges list; prometheus-copy filtered), no production caller mixes the `metrics.debate` inc/dec helpers with the orchestrator path (no new double-counting), and the observability conftest already unregisters `aragora_*` collectors between tests so the stale-gauge regression test can't collide with a leftover collector.

- [P3] `aragora/observability/server_metrics/export.py:200-208` — the dedup filter hardcodes `aragora_active_debates` in four string literals even though `ACTIVE_DEBATES` is imported in this module; if the gauge is renamed in `api.py`, the filter silently stops matching and the exposition emits duplicate HELP/TYPE/sample blocks for the same metric name, which strict Prometheus parsers reject. Build the prefixes from `ACTIVE_DEBATES.name`.
- [P3] `aragora/observability/metrics/debate.py:58-66` — `_ensure_active_debates_collector` reads private `registry._names_to_collectors` and does unregister→register without the registry's own lock; the module lock only serializes against itself, and `init_debate_metrics` catches only `ImportError`, so an unexpected `ValueError`/`AttributeError` here (concurrent same-name registration, or a future prometheus_client renaming the private attr) would abort metrics init partway. Precedent exists (`_get_or_create_*` uses the same private dict), so noting fragility only.
- [P3] `aragora/observability/metrics/debate.py:208` — `_init_noop_metrics` now binds the live canonical gauge instead of `NoOpMetric()`, so with `METRICS_ENABLED=false` the `increment/decrement/set_active_debates` helpers mutate real shared state rather than no-op'ing. Consistent with the never-gated `server_metrics` path, but it quietly changes what the disabled flag means for this one metric — worth a comment.

One scope note: `git diff main...pr-9292-review` includes 12 additional files (arena_bridge, cross_subscribers, workflow engine, etc.) beyond the 6 listed; per instructions I reviewed only the 6-file diff provided.

dogfood: yes

scarmani:

Head: 18a8838 (18a8838c7e557a4b0acccc3d96a604d1e6160a10), committed 2026-07-15T04:33:25Z.
PR: #9292.
Model family: claude

Verdict: PASS

Review basis: full static verification at head 18a8838 (branch present locally) — import graph, all `ACTIVE_DEBATES` call sites, custom Gauge API, legacy shim, conftest registry hygiene. Could not execute the new tests (worktree creation denied in this session), so findings are from code inspection.

Verified as sound:
- No import cycle: `server_metrics/*` never imports back into `aragora.observability.metrics`; the new `metrics/debate.py` → `server_metrics/api.py` edge is one-way.
- Orchestrator hot path (`orchestrator.py:1251` `.inc()`, `orchestrator_runner.py:1299` `.dec()`) goes through the `aragora.server.metrics` `__getattr__` shim to the canonical Gauge, whose `inc/dec/set/get` API is compatible; no stale `._value` accesses remain anywhere.
- No double-counting: `track_debate()` / `track_debate_execution()` have no production callers; orchestrator inc/dec is the sole mutator.
- The `export.py` filter is precisely anchored (requires ` ` or `{` after the name, so `aragora_active_debates_*` metrics are untouched) and actually fixes a pre-existing exposition-spec violation (duplicate `aragora_active_debates` series in native + prometheus sections).
- Test isolation holds: `tests/observability/conftest.py` already unregisters all `aragora_*` collectors per-test, so the new test's real `Gauge("aragora_active_debates")` cannot collide, and the added `set(0)` reset closes gauge-value leakage.

Findings:
- [P3] `aragora/observability/metrics/debate.py:57-66` — `_ensure_active_debates_collector` is called inside a `try` that only catches `ImportError` (`debate.py:189`). A `ValueError` from `registry.register()`/`unregister()` (e.g., a non-aragora collector racing to claim the name from another thread, outside this module's lock) would propagate out of `init_debate_metrics()` into `_ensure_init()` callers on the debate path, where the old `_get_or_create_gauge` flow never raised past init. Consider catching `ValueError` and logging.
- [P3] `aragora/observability/metrics/debate.py:208` — `_init_noop_metrics` now binds `ACTIVE_DEBATES` to the live canonical gauge, so with metrics disabled, `increment_active_debates()`/`track_debate()` mutate and export real state instead of no-oping. Consistent with the canonicalization intent (the server_metrics gauge was always flag-independent), but it is a semantic change to disabled mode worth a docstring note.
- [P3] `aragora/observability/server_metrics/export.py:198-205` — the filter silently drops any `aragora_active_debates{...}` labeled series a third party might register in the prometheus REGISTRY, with no log; divergent duplicate registrations become invisible rather than detectable. Acceptable dedup tradeoff, flagging for awareness.

dogfood: yes

PR #9295, opened 2026-07-15
fix(observability): register layered metric providers

scarmani:
## Source

- Structural Excellence feature `p4a-observability-provider-inversion`
- Fulfills `VAL-P4A-019`

## Behavior

- Replaces hidden Nomic metric discovery in `aragora.observability.server_metrics.export` with an explicit lower-layer provider contract.
- Replaces hidden server versioning discovery in `aragora.observability.prometheus_recording` with an explicit sunset-days callback.
- Registers the Nomic metric collection from server observability startup and the V1 sunset callback from server deprecation composition.
- Preserves all ten existing Nomic metric exports and both Prometheus and simple-metric sunset gauge paths when providers are registered.
- Omits optional higher-layer data when providers are absent or raise expected runtime errors, without importing upward.
- Moves Nomic metric types off the deprecated server compatibility path onto canonical observability types.

## Validation

- Exact forbidden-reference scan under `aragora/observability` -> no matches, `rg_exit=1`
- `python3 -m pytest tests/observability/test_layered_metrics_providers.py -q` -> 5 passed
- Related observability, startup, Nomic, server metrics, middleware, and versioning suites -> 242 passed
- Changed-file mypy -> no issues in 5 source files
- `make lint` -> passed
- `ruff format --check aragora/ tests/` -> 10,254 files already formatted
- `python3 scripts/report_code_quality.py --check` -> ratchet passed
- `make test-smoke` -> Core/Server/Memory imports passed
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` -> ok

## Risk

Tier 2 bounded layering repair. The change keeps existing registered metric behavior, adds fail-soft optional-provider handling, and is covered at the lower contracts and both higher-layer registration roots.

### DIFF
--- a/aragora/nomic/metrics.py
+++ b/aragora/nomic/metrics.py
@@ -27,9 +27,11 @@ import logging
 import time
 from typing import TYPE_CHECKING, Any
 
-from aragora.server.metrics import Counter, Gauge, Histogram
+from aragora.observability.server_metrics import Counter, Gauge, Histogram
 
 if TYPE_CHECKING:
+    from aragora.observability.server_metrics.export import NomicMetricCollection
+
     from .events import Event
     from .states import NomicState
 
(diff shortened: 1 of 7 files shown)


scarmani:

Head: 3935bf0 (3935bf0447978b5f18c9a690896aa7e00024cd58), committed 2026-07-15T05:51:17Z.
PR: #9295.
Model family: claude

Verdict: PASS

Verified before ruling: `aragora.server.metrics` is a pure deprecated shim over `aragora.observability.server_metrics`, so the nomic import swap references the same metric classes; the only production caller of `update_v1_days_until_sunset()` is `register_default_deprecations()`, which now registers the provider first; `generate_metrics()` has no in-repo production caller outside the package, and `register_observability_sinks()` runs on both the sequential (`startup/__init__.py:449`) and parallel (`startup/parallel.py:537`) startup paths. Both touched test files pass locally (29 passed, 0.85s).

- [P3] `aragora/server/startup/observability.py:29-31` — the new `aragora.nomic.metrics` import sits in the same shared try block as all other sink registrations, and importing it pulls the entire `aragora.nomic` package tree (which transitively imports `aragora.server` and connectors). Any ImportError anywhere in that tree now silently (debug-level) disables PagerDuty/channel/webhook/KM registration too. Marginal in practice since this only runs inside server startup, but per-registration try blocks would remove the shared-fate coupling.
- [P3] `aragora/observability/server_metrics/export.py:167` and `prometheus_recording.py:515` — behavior contract change: out-of-repo/SDK consumers calling the public `generate_metrics()` or `update_v1_days_until_sunset()` without running a composition root now silently lose the nomic series / sunset gauge (previously auto-resolved via importlib). In-repo call sites are covered; worth a release-note line.
- [P3] `aragora/observability/server_metrics/export.py:85` — the fail-soft catch was narrowed from `(ImportError, AttributeError)` to `(OSError, RuntimeError, TypeError, ValueError)`; a buggy provider raising `AttributeError`/`KeyError` would now propagate out of `generate_metrics()` and break the exposition despite the "cannot break metrics export" test claim. Current registered providers can't realistically raise those, and narrowing matches repo convention, so non-blocking.

dogfood: yes

scarmani:

Head: 3935bf0 (3935bf0447978b5f18c9a690896aa7e00024cd58), committed 2026-07-15T05:51:17Z.
PR: #9295.
Model family: openai

Verdict: PASS

- [P3] `aragora/server/startup/observability.py:29` - Adding Nomic provider import to the single all-or-nothing adapter import block makes observability registration more fragile: an unrelated `aragora.nomic` import failure would prevent the already-independent SLO/KM adapters from registering. Non-blocking, but isolating optional adapter imports would reduce regression blast radius.

dogfood: yes

PR #9298, opened 2026-07-15
fix(observability): isolate optional Nomic startup adapter

scarmani:
## Summary

- isolate the optional Nomic metrics adapter import and registration from the four existing startup observability adapters
- preserve fail-soft logging and return `false` when Nomic is unavailable or fails
- add a focused regression proving PagerDuty, channel, webhook, and Knowledge Mound registration still run on Nomic import failure

## Validation

- `python3 -m pytest tests/server/startup/test_observability.py tests/observability/test_layered_metrics_providers.py -q` (`30 passed`)
- `python3 -m mypy --ignore-missing-imports --follow-imports=skip --show-error-codes aragora/server/startup/observability.py` (`Success: no issues found`)
- `make lint` and `ruff format --check aragora/ tests/ scripts/`
- `python3 scripts/ci/check_import_contracts.py --layers foundation,infrastructure`
- `make test-smoke`
- `PYTEST_BIN="python3 -m pytest" bash scripts/test_tiers.sh smoke` (`138 passed`)

## Boundaries

- no changes to `init_prometheus_metrics()`, provider contracts, observability-layer imports, import-contract baselines, or the three ambient higher-layer edges
- 64 additions and 11 deletions across two focused files

### DIFF
--- a/aragora/server/startup/observability.py
+++ b/aragora/server/startup/observability.py
@@ -13,6 +13,8 @@ logger = logging.getLogger(__name__)
 
 def register_observability_sinks() -> bool:
     """Register higher-layer adapters with observability contracts."""
+    results: list[bool] = []
+
     try:
         from aragora.connectors.devops.slo_alert_sink import (
             register_slo_alert_sink as register_pagerduty_sink,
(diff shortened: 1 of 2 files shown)


scarmani:

Head: 3669038 (366903850a0b897a57a04955b018f530aee92247), committed 2026-07-15T06:59:44Z.
PR: #9298.
Model family: claude

Verdict: PASS

- [P3] `aragora/server/startup/observability.py:32-45` — the four non-Nomic adapters remain coupled in a single try block: an `OSError`/`RuntimeError`/`ValueError` raised by, say, `register_channel_sink()` still skips webhook and KM health registration entirely. This is the same skip-siblings failure mode the PR fixes for Nomic, left in place for the other four. Scope-consistent with an isolated Nomic fix, but worth a follow-up (per-adapter isolation or a loop over `(import_path, fn)` pairs).

Verification performed: the mocked module attribute names in the new test match the real imports (both `aragora.connectors.devops.slo_alert_sink` and `aragora.control_plane.slo_alert_sink` export `register_slo_alert_sink`, so the `assert_called_once_with()` checks bind to the correct mocks); `results` can never be empty so `all(results)` never vacuously returns True; the fully-successful path still returns True, preserving `test_registers_all_external_sink_providers`; and the caller `init_prometheus_metrics` ignores the return value, so the unchanged False-on-partial-failure contract regresses nothing. Full test file passes: 25 passed in 0.86s.

dogfood: yes

PR #9314, opened 2026-07-15
fix(security): count key rotation list failures

scarmani:
## Source

Mission feature `misc-security-key-rotation-failure-accounting`, fulfilling `VAL-P4A-017`.

## Behavior

- Count configured record-list failures as at least one failed record, even before the list function can report a record count.
- Keep required sync-store setup failures fail-closed.
- Preserve intentionally absent integration and Gmail stores as best-effort optional absence.
- Preserve sanitized result errors while leaving exception details only in logs.

## Reviewer context

`VAL-P4A-017` explicitly distinguishes required sync setup/list failures from optional integration and Gmail absence. The bounded implementation changes only the outer record-list failure accounting in `rotate_and_reencrypt_store`; it does not make optional `None` store configurations fail.

## Validation

- `python3 -m pytest tests/security/test_migration.py::TestMigrationAuditProvider::test_live_sync_rotation_fails_closed_when_storage_resolution_fails tests/security/test_migration.py::TestMigrationAuditProvider::test_live_sync_rotation_fails_closed_when_connector_list_fails tests/security/test_migration.py::TestMigrationAuditProvider::test_optional_store_absence_remains_best_effort -q` -> 3 passed
- `python3 -m pytest tests/security/test_migration.py -q --timeout=120` -> 47 passed
- `make lint` -> passed
- `ruff format --check aragora/ tests/ scripts/` -> 10712 files formatted
- changed-file mypy on `aragora/security/migration.py` -> no issues
- mission differential typecheck -> PASS, new=103 <= 108
- `make test-smoke` -> passed
- smoke tier -> 138 passed
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` -> passed

## Risk

Tier 3 security surface. Prepare-only after exact-head two-family evidence; human risk settlement is required before merge.

### DIFF
--- a/aragora/security/migration.py
+++ b/aragora/security/migration.py
@@ -635,7 +635,7 @@ def rotate_and_reencrypt_store(
 
     except (TypeError, RuntimeError, OSError, ValueError) as e:
         result.errors.append("Re-encryption failed due to an internal error")
-        result.failed_records = result.total_records
+        result.failed_records = max(1, result.total_records)
         logger.error("Key rotation re-encryption failed for %s: %s", store_name, e)
 
     return result
--- a/tests/security/test_migration.py
+++ b/tests/security/test_migration.py
@@ -549,9 +549,7 @@ class TestMigrationAuditProvider:
             {"api_key": "decrypted-secret"},
         )
 
-    def test_default_rotation_resolves_unregistered_storage_store(
-        self, encryption_service, monkeypatch
-    ):
+    def test_optional_store_absence_remains_best_effort(self, encryption_service, monkeypatch):
         old_key = MagicMock(version=1)
         new_key = MagicMock(key_id="default", version=2)
         encryption_service._keys = {"default": old_key}
@@ -577,6 +575,8 @@ class TestMigrationAuditProvider:
 
         assert result.success is True
         assert result.stores_processed == 1
+        assert result.failed_records == 0
+        assert result.errors == []
         get_sync_store.assert_awaited_once_with()
         store.list_connectors.assert_awaited_once_with()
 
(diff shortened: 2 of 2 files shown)


scarmani:

Head: 4930d22 (4930d226be180214ecfa80edfe39ba8c490b1aab), committed 2026-07-15T18:05:03Z.
PR: #9314.
Model family: claude

Verdict: PASS

- [P3] `aragora/security/migration.py:224` — the plaintext-migration path (`migrate_store`) still has the identical zero-evidence bug this PR fixes: if `list_fn()` raises, `failed_records = total_records` stays 0 and `MigrationResult.success` returns True despite the appended error. Same-class latent bug, outside this diff's scope, but worth a follow-up applying the same `max(1, ...)` treatment.
- [P3] `aragora/security/migration.py:755-759` and `771-773` — in the aggregator, exceptions for non-sync stores (integration/gmail) and the outer catch (e.g. `rotate_key` itself failing) append to `errors` without incrementing `failed_records`, so `KeyRotationResult.success` can still read True with errors present. Pre-existing, not touched by this diff.
- [P3] `aragora/security/migration.py:638` — the fix intentionally allows `failed_records (1) > total_records (0)`; the only consumer (line 752) just sums into the aggregate, so no ratio/invariant breakage today, but `to_dict()` output can now show `total_records: 0, failed_records: 1`, which downstream report renderers should tolerate.

The core change is correct: previously a `list_connectors()` failure left `total_records == 0`, so the except handler set `failed_records = 0` and both `MigrationResult.success` and the aggregated `KeyRotationResult.success` reported True on a completely failed re-encryption — exactly the zero-evidence-PASS pattern. `max(1, total_records)` fails closed, partial-progress semantics are unchanged, and the new test verifies fail-closed behavior plus error-message sanitization (the `credential=secret-value` exception text is not leaked into `result.errors`). The renamed best-effort test correctly pins the opposite path (optional store absent → success, zero failures). All 47 tests in `tests/security/test_migration.py` pass at head 4930d22.

dogfood: yes

scarmani:

Head: 4930d22 (4930d226be180214ecfa80edfe39ba8c490b1aab), committed 2026-07-15T18:05:03Z.
PR: #9314.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9318, opened 2026-07-15
fix(memory): stamp continuum tiers in UTC to match julianday('now')

scarmani:
## Summary

Extends the CritiqueStore UTC fix (#9311 flake root cause, `_utc_now_iso()` in `aragora/memory/store.py`) to the continuum memory subsystem, which had the identical local-vs-UTC clock mix.

**Bug:** every `continuum_memory` writer stamped `created_at` / `updated_at` / `last_promotion_at` with local-time `datetime.now().isoformat()`, while the readers age those columns with UTC clocks — `julianday('now')` decay ranking in all four tier retrievals + `retrieval.py` + `coordinator_search.py`, `datetime('now')`-style TTL cutoffs in `continuum_stats.cleanup_expired_memories`, cooldown gating in `TierManager.should_promote` / `continuum_consolidation`, and `calculate_glacial_decay`. SQLite treats naive strings as UTC, so TTL/decay/retrieval math skewed by the machine's UTC offset. Worst case at negative offsets: a fresh fast-tier row (2h TTL) looks hours old and is archived on sight.

**Fix:** a shared `utc_now_iso_naive()` helper in `aragora/utils/datetime_helpers.py` writes naive-UTC ISO strings — same on-disk format as legacy rows and SQLite's `CURRENT_TIMESTAMP` defaults, so `julianday` and `datetime.fromisoformat` consumers parse unchanged and **no migration is needed** (legacy local-time rows keep a bounded skew that self-corrects on next update). All writers route through it: fast/medium/slow/glacial `store_*` mixins, crud + coordinator add/update/promote/demote, outcome surprise updates, KM/search prewarm, batch promotion, `outcome_bridge`. Python-side readers that compared stamps against local `datetime.now()` now use the same naive-UTC clock and convert aware inputs to UTC before stripping tzinfo.

**Audited as already consistent (unchanged):** `continuum/retrieval.py` (read-only julianday scoring; its writers are fixed here), `aragora/learning/meta.py` (inserts rely on SQLite `CURRENT_TIMESTAMP`, already UTC), `memory/streams.py` (self-contained local-clock system, no julianday reads).

## Tests

- New `tests/memory/test_continuum_timestamp_utc.py` following the `TestTimestampUTCConsistency` pattern: pins `TZ=Pacific/Kiritimati` (UTC+14) and `Etc/GMT+12` (UTC-12) via `time.tzset()`, asserts julianday age of fresh rows < 5 min across write paths (core `add`, coordinator `store_*`, `update_entry`), fresh fast-tier rows survive cleanup, fresh promotions stay inside the cooldown, fresh glacial stamps show ~no decay, and `CURRENT_TIMESTAMP`-format legacy rows stay compatible.
- Existing tests that backdated rows with local `datetime.now()` (`test_tier_ttl_expiration.py`, `continuum/test_coordinator.py`, `test_continuum_consolidation.py`, glacial decay tests) switched to the naive-UTC clock.
- Continuum suites (756 tests) pass under `UTC`, `UTC+14`, `UTC-12`, and `America/Chicago`; full `tests/memory/` passes (2,263) excluding pre-existing env-dep failures (`numpy` / `pytest-benchmark` missing locally).
- `ruff check` / `ruff format` clean; scoped mypy shows only the 3 pre-existing errors also present on pristine `origin/main`.

### DIFF
--- a/aragora/memory/continuum/coordinator.py
+++ b/aragora/memory/continuum/coordinator.py
@@ -19,7 +19,6 @@ import json
 import logging
 import sqlite3
 import threading
-from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
@@ -32,6 +31,7 @@ from aragora.persistence.db_config import DatabaseType, get_db_path
 from aragora.resilience.retry import PROVIDER_RETRY_POLICIES, with_retry
 from aragora.storage.base_store import SQLiteStore
 from aragora.storage.schema import SchemaManager
+from aragora.utils.datetime_helpers import utc_now_iso_naive
 from aragora.utils.json_helpers import safe_json_loads
 
 # Import base types and tier mixins
@@ -305,7 +305,7 @@ class ContinuumMemory(
         Returns:
             The created memory entry
         """
-        now: str = datetime.now().isoformat()
+        now: str = utc_now_iso_naive()
 
         with self.connection() as conn:
             cursor: sqlite3.Cursor = conn.cursor()
@@ -504,7 +504,7 @@ class ContinuumMemory(
                 SET success_count = ?, failure_count = ?, updated_at = ?
                 WHERE id = ?
                 """,
-                (entry.success_count, entry.failure_count, datetime.now().isoformat(), entry.id),
+                (entry.success_count, entry.failure_count, utc_now_iso_naive(), entry.id),
             )
             conn.commit()
             return cursor.rowcount > 0
@@ -559,7 +559,7 @@ class ContinuumMemory(
 
         # Always update timestamp
         updates.append("updated_at = ?")
-        params.append(datetime.now().isoformat())
+        params.append(utc_now_iso_naive())
 
         # Add memory_id as final parameter
         params.append(memory_id)
--- a/aragora/memory/continuum/coordinator_search.py
+++ b/aragora/memory/continuum/coordinator_search.py
@@ -13,11 +13,11 @@ import hashlib
 import json
 import logging
 import sqlite3
-from datetime import datetime
 from typing import TYPE_CHECKING, Any
 
 from aragora.memory.tier_manager import MemoryTier
 from aragora.resilience.retry import PROVIDER_RETRY_POLICIES, with_retry
+from aragora.utils.datetime_helpers import utc_now_iso_naive
 from aragora.utils.json_helpers import safe_json_loads
 
 from aragora.memory.continuum.base import (
@@ -533,8 +533,8 @@ class CoordinatorSearchMixin:
                 return 0
 
             # Batch update all entries in a single transaction using executemany
-            prewarm_time: str = datetime.now().isoformat()
-            current_time: str = datetime.now().isoformat()
+            prewarm_time: str = utc_now_iso_naive()
+            current_time: str = utc_now_iso_naive()
 
             # Prepare batch update data
             update_data: list[tuple[str, str, str]] = []
--- a/aragora/memory/continuum/coordinator_tier_ops.py
+++ b/aragora/memory/continuum/coordinator_tier_ops.py
@@ -13,10 +13,10 @@ from __future__ import annotations
 import logging
 import math
 import sqlite3
-from datetime import datetime
 from typing import TYPE_CHECKING, Any
 
 from aragora.memory.tier_manager import MemoryTier, TierConfig
+from aragora.utils.datetime_helpers import utc_now_iso_naive
 from aragora.utils.json_helpers import safe_json_loads
 
 from aragora.memory.continuum.base import (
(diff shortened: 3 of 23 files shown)


scarmani:
Independent review from the session that authored #9311 (the original julianday-flake finding). Verified at head 23c7184462:

**Verdict: LGTM with one [P2] advisory finding worth fixing before mark-ready.**

Verified clean:
- `utc_now_iso_naive()` helper is correct and well-documented (naive-UTC matches legacy rows, SQLite `CURRENT_TIMESTAMP`, and keeps `fromisoformat` consumers naive).
- All continuum tier/crud/coordinator writers converted; Python-side readers (TTL cutoffs in `continuum_stats`, `tier_manager.should_promote` cooldown at line 278, glacial decay) moved to UTC together — no writer/reader split.
- Ran all 6 PR test files under `TZ=UTC` (301 passed), `TZ=Pacific/Kiritimati` UTC+14 (86 passed), `TZ=America/Chicago` (21 passed).

**[P2] Missed writer in the same bug family: `aragora/memory/continuum_snapshot.py`** — snapshot *restore* stamps `continuum_memory.updated_at` with local-time `datetime.now().isoformat()` (line 229 on the merge-existing UPDATE path; lines 256-257 as INSERT fallback defaults). That's the same julianday-aged column this PR converts everywhere else, so restoring a snapshot on a non-UTC machine re-skews rows by the UTC offset. Three-line fix with the new helper; suggest including it here since it's squarely in scope.

[P3] Benign local-time stamps you may want for consistency (display/bookkeeping only, never julianday-compared — fine to leave): `continuum/outcome.py:313` (feedback-log dict), `tier_manager.py:142/177/549/567` (`last_reset`/`last_run_at` metrics).

scarmani:

Head: 23c7184 (23c718446269b5e0ee6ea707f48f706e83e809b2), committed 2026-07-15T18:56:39Z.
PR: #9318.
Model family: claude

Verdict: PASS

- [P3] `aragora/memory/continuum/glacial_tier.py` (277-284): The new timezone-aware logic for `calculate_glacial_decay` correctly handles naive-UTC stamps, but the docstring should explicitly note that `updated_at` strings are expected to be naive UTC (not just "legacy local-time rows"). Current docstring is ambiguous.
- [P3] `aragora/memory/continuum_glacial.py` (246-253): Same timezone-aware logic as `glacial_tier.py`, same docstring suggestion.
- [P3] `tests/memory/test_continuum_timestamp_utc.py`: Excellent regression coverage. Minor nit: the `MAX_SKEW_SECONDS = 300` comment could clarify that this is a generous allowance for clock drift, not a tolerance for local-time skew (which would be >= 3600s).

dogfood: yes

scarmani:

Head: 23c7184 (23c718446269b5e0ee6ea707f48f706e83e809b2), committed 2026-07-15T18:56:39Z.
PR: #9318.
Model family: claude

Verdict: PASS

Verified at head 23c7184 (fetched branch `worktree-utc-continuum-tz-fix`; base repo untouched, review ran in a throwaway worktree since removed):

- All 294 tests in the changed test files pass, including the 7 new TZ-pinned regression tests (`Pacific/Kiritimati` / `Etc/GMT+12`).
- Checked the mixed-format trap: `cleanup_expired_memories` compares via `datetime(updated_at) < datetime(?)`, which SQLite normalizes, so legacy space-separated `CURRENT_TIMESTAMP` rows and the new `T`-separated isoformat stamps compare correctly. `promote_batch`'s raw string comparison on `last_promotion_at` is internally consistent (both sides are Python isoformat).
- The tz-aware branch in `calculate_glacial_decay` is now genuinely more correct: the old code stripped `tzinfo` from non-UTC offsets (silently wrong); the new `astimezone(utc)` conversion fixes that.
- The leftover `datetime.now()` at `aragora/memory/continuum/outcome.py:313` is an in-memory feedback-log stamp (never DB-aged), so the retained `datetime` import is used — no dead import. `tier_manager.py`'s remaining local stamps (`last_reset`, `last_run_at`) are telemetry-only, never compared against row timestamps.

Findings:

- [P3] No backfill for pre-existing local-time rows — `aragora/memory/continuum_stats.py:164` (cleanup cutoff). On west-of-UTC hosts, legacy rows now look older to the UTC cutoff than they did to the old local-time cutoff, so near-TTL legacy rows can expire on the first cleanup after upgrade (one-time, bounded by UTC offset; archive-by-default means recoverable, not lost). Worth a release note, not a code change.
- [P3] Missed sibling writer in the same bug family — `aragora/memory/continuum_snapshot.py:256-257` (not in this PR): snapshot restore defaults missing `created_at`/`updated_at` to local `datetime.now().isoformat()`, re-introducing local stamps into julianday-aged rows. Suggest a follow-up to route it through `utc_now_iso_naive()`.
- [P3] `prewarm_time` and `current_time` are two back-to-back identical `utc_now_iso_naive()` calls — `aragora/memory/continuum/coordinator_search.py:536-537` and `km_integration.py:130-131` (shape carried over from old code); one variable would do.

dogfood: yes

PR #9319, opened 2026-07-15
fix(rlm): restore #8277 load_dotenv guard that never landed on main

scarmani:
## Summary

Running `tests/storage/` with `-n auto --dist loadscope` on a pristine origin/main worktree produced **112 failures** (all `SecretNotFoundError` in strict-secrets mode) while the same files passed standalone.

Root cause: the #8277 fix (commit bd6fd63dcc, authored Jun 12) **was believed merged but never actually landed on main** — #8277 is an issue (not a PR), and the commit sat stranded on the unmerged branch `worktree-fix-8277-registry-pollution`. (A follow-up audit of 1,674 merged PRs confirmed main was never rewritten or force-pushed; the failure mode is stranded-branch work, not lost history.) Without the guard, the official `rlm` package's import-time `load_dotenv()` runs in every pytest-xdist worker; execnet workers' `__main__` has no `__file__`, so python-dotenv searches from cwd upward and injects the repository `.env` (including `ARAGORA_SECRETS_STRICT=true`) into every worker before its first test.

## Changes

- **Cherry-pick bd6fd63dcc** (clean): adds `aragora.utils.env.preserve_environ` + guards in `aragora/rlm/bridge.py` and `aragora/server/handlers/features/rlm.py` + unit tests.
- **New import-site regression test** (`tests/utils/test_preserve_environ.py::test_bridge_import_does_not_inject_repo_dotenv`): reproduces the execnet condition in a subprocess (`python -c`, cwd containing a `.env`) and asserts importing `aragora.rlm.bridge` does not mutate `os.environ`. The unit tests only cover the utility and cannot catch the guard being dropped from the import site — this one fails if the `with preserve_environ():` wrapper is removed (mutation-tested).

## Verification

- `TZ=UTC pytest tests/storage/ -n auto --dist loadscope`: **112 failed → 3547 passed, 21 skipped** (skips are legitimate Redis-unavailable skips that previously died earlier at the strict-secrets error). 5+ consecutive green runs.
- `make ci-required` green locally; ruff + mypy clean on all touched files.
- New test verified to fail with the guard stripped, pass with it in place.


scarmani:

Head: 196e198 (196e19826f7c23dc2feccbe4a42ea9a47b23291c), committed 2026-07-15T19:48:08Z.
PR: #9319.
Model family: claude

Verdict: PASS

Verified before signing off: the rollback logic handles add/mutate/delete/exception correctly (exercised all four paths directly), E402 is globally ignored in `pyproject.toml:177` so the mid-module import in `bridge.py` won't trip the lint gate, the handler's surrounding `except ImportError` (`rlm.py:780`) safely covers the added `from aragora.utils.env import ...` statement, and no import-time `load_dotenv` exists elsewhere in the `aragora.rlm.bridge` import chain that would invalidate the integration test.

- [P3] `aragora/rlm/bridge.py:83` — rolling back the environment also strips vars the `rlm` package legitimately autoloaded for its own runtime (e.g. provider API keys a user kept in a cwd `.env` for rlm's clients), before `OfficialRLM` is ever instantiated. This is the intended tradeoff of #8277 and Aragora has its own secrets path, but it's a silent behavior change for anyone who relied on rlm's implicit dotenv loading — worth a line in the PR description/changelog.
- [P3] `aragora/utils/env.py:35` — `preserve_environ` snapshot/rollback is not concurrency-safe: in the server handler (`_get_rlm_status`), the first request's import window will roll back env mutations made concurrently by other threads. Env mutation is already thread-unsafe process-wide, so this is acceptable, but the docstring could warn against use outside import-time/startup contexts.
- [P3] `tests/utils/test_preserve_environ.py:76` — the end-to-end test spawns two subprocesses each doing a full `aragora.rlm.bridge` import with a 300s timeout and inherits the entire parent environment; a hostile parent env var could fail it spuriously. The control-probe-then-skip design is good; consider stripping `ARAGORA_*` from the probe env for isolation.

dogfood: yes

scarmani:

Head: 196e198 (196e19826f7c23dc2feccbe4a42ea9a47b23291c), committed 2026-07-15T19:48:08Z.
PR: #9319.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

PR: #9319
Exact head: 196e19826f7c23dc2feccbe4a42ea9a47b23291c
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #9319 is non-draft and live exact-head checks/merge-packet remain otherwise green.

scarmani:

Head: 7fb8145 (7fb8145858a8a37daf6f46095c1b31c39eaf5f64), committed 2026-07-16T18:32:24Z.
PR: #9319.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

Head: df8ca39 (df8ca3980c7fb3351e4520e0ac079e6b414037fc), committed 2026-07-16T20:02:32Z.
PR: #9319.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

Head: 57a153e (57a153e44282aa379042f5df30ced8058157311b), committed 2026-07-16T20:20:38Z.
PR: #9319.
Model family: claude

Verdict: PASS

- [P3] `aragora/rlm/bridge.py:83-85` — the bridge's rlm import uses bare `preserve_environ()` and does not participate in `_RLM_IMPORT_LOCK`. In the B-first interleaving (thread B enters `import_rlm_guarded()` and rlm's `load_dotenv()` pollutes the env; thread A begins importing `aragora.rlm.bridge`, snapshots the *polluted* environ, then blocks on rlm's per-module import lock; B's restore cleans the env; A's restore re-writes the polluted keys from its snapshot), the pollution the PR exists to prevent persists. Window is very narrow (both must be first-time imports, typically startup-only), so non-blocking — but the one-line fix is to route bridge through `import_rlm_guarded()` (`rlm_mod = import_rlm_guarded(); OfficialRLM = rlm_mod.RLM`), which also removes the second unserialized `preserve_environ` site entirely.
- [P3] `aragora/utils/env.py:76-78` — the lock-free fast path `sys.modules.get("rlm")` can return a *partially initialized* module while another thread is mid-import (Python puts the module in `sys.modules` before executing its body; a plain `import rlm` blocks on the per-module init lock via `__spec__._initializing`, a raw `sys.modules` probe does not). Current callers tolerate it (`getattr(rlm, "__version__", "unknown")`), but the helper's docstring implies a fully-initialized module. Safer fast path: just `import rlm; return rlm` (the import system's own cached-lookup is already lock-cheap and init-safe), keeping `_RLM_IMPORT_LOCK` only around the first-import/`preserve_environ` section.
- [P3] `tests/utils/test_preserve_environ.py:151` — `monkeypatch.delitem(sys.modules, "rlm")` then re-importing (when rlm is installed) re-executes `rlm/__init__` while stale `rlm.clients` submodule entries remain in `sys.modules`, producing a second parent module object for the rest of the session (monkeypatch restores the entry, but cross-module identity oddities can surface in unrelated tests within the same worker). Consider also popping `rlm.*` submodules, or running this test in a subprocess like the bridge test.

The core mechanism is sound: `preserve_environ` snapshot/restore logic is correct (handles added, mutated, and deleted keys, exception-safe), the double-checked lock closes the concurrent-first-call race for handler threads, and the subprocess-based regression test guards the import *site* with a control probe — which directly addresses how the original fix sat stranded unnoticed.

dogfood: yes

scarmani:

Head: ca66b85 (ca66b851e23f4195cdc872512e1b76f227d8c75b), committed 2026-07-16T20:35:58Z.
PR: #9319.
Model family: claude

Verdict: PASS

- [P3] `aragora/utils/env.py` (`preserve_environ` restore loops) — the snapshot/restore cannot attribute changes: any env mutation made by an *unrelated* thread during the guarded import window is silently reverted (added keys deleted, changed keys reset). It's a one-time, short window (first `rlm` import only, serialized by the lock) so it's acceptable, but the docstring warns only about concurrent first-calls interleaving, not about clobbering unrelated concurrent writers. Worth a sentence in the docstring.
- [P3] `tests/utils/test_preserve_environ.py:151-175` (`test_import_rlm_guarded_serializes_concurrent_first_calls`) — the closing comment claims "entries never interleaved (lock serializes)" but the test only asserts `len(guard_entries) == len(results)`; it never checks non-interleaving (e.g. an enter/exit event log). The comment overstates the coverage; either assert enter/exit pairing or soften the comment.
- [P3] `aragora/utils/env.py` (`import_rlm_guarded`) — pollution is still *transiently* visible process-wide to other threads while the guarded import runs (restore happens only after). Inherent to the approach and fine for the xdist bug (single-threaded workers), just noting it doesn't protect concurrent server-thread readers during that window.

Checks that came back clean: guard coverage is complete — the only two direct `rlm` import sites in the tree are exactly the two this PR wraps (no other static or `import_module("rlm")` sites); the lock correctly resolves the prior interleaved-restore P2, and the "no lock-free fast path" rationale is sound (relying on the import statement as the memoization primitive avoids the partially-initialized `sys.modules` hazard); no deadlock cycle is possible since `rlm` (third-party) can never re-enter `_RLM_IMPORT_LOCK`; `except (ImportError, AttributeError)` in `bridge.py` preserves the original `from rlm import RLM` failure semantics; the end-to-end subprocess test guards its own `importorskip`, pins `PYTHONPATH` to the tree under test, and includes a control probe so it can't rot into a vacuous pass; monkeypatching `env_mod.preserve_environ` in the concurrency test genuinely intercepts the guard (module-global lookup at call time).

dogfood: yes

scarmani:

Head: ca66b85 (ca66b851e23f4195cdc872512e1b76f227d8c75b), committed 2026-07-16T20:35:58Z.
PR: #9319.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

PR: #9319
Exact head: ca66b851e23f4195cdc872512e1b76f227d8c75b
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #9319 is non-draft and live exact-head checks/merge-packet remain otherwise green.

PR #8882, opened 2026-07-05
fix(tests): repair main-red debate shard failures

an0mium:
## Summary
- Repairs the main-red `test-fast (debate-nz)` failures observed on the main Tests run.
- Preserves the review package callable export after package imports for the golden API path.

## Validation
- `PYTHONPATH=.:aragora-debate/src /Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest -q tests/debate/test_scenarios_root.py::TestRunScenarioDebate::test_handles_exception tests/debate/test_scenarios_root.py::TestRunScenarioDebate::test_error_result_structure tests/debate/test_wisdom_injector_root.py::TestWisdomInjectorInit::test_init_default_storage_path tests/debate/test_unanimous_consensus.py::TestUnanimousThreshold::test_one_dissent_breaks_unanimity tests/debate/test_phases_synthesis.py::TestSynthesisPromptBuilding::test_truncates_long_proposals tests/debate/test_unanimous_consensus.py::TestVotingErrors::test_voting_error_breaks_unanimity tests/debate/test_pr_review_protocol_smoke.py::test_pr_review_protocol_packet_smoke --tb=short --timeout=120` -> `7 passed`
- `PYTHONPATH=.:aragora-debate/src /Users/armand/.pyenv/versions/3.11.11/bin/python -m pytest -q tests/server/test_golden_api.py --tb=short --timeout=120` -> `17 passed`
- `git diff --check origin/main...HEAD` -> passed
- `make ci-required` on this branch: ruff passed; repository-wide mypy failed with `2645 errors in 647 files`.
- Baseline check: `mypy aragora/ --ignore-missing-imports` on current `origin/main` `0ceb5d6898aecae855086d38b98fcc23ed64554d` also failed with `2645 errors in 647 files`, so the mypy failure is inherited from main rather than branch-local.

## Main-Red Context
- Repair branch head: `058a16a75f6c4748c6e3dd6dcf9883856b7a68ee`.
- Current validated base: `origin/main` at `0ceb5d6898aecae855086d38b98fcc23ed64554d`.
- Covered failed nodes from the main `Tests` run `28730456490`: the seven `test-fast (debate-nz)` failures listed in the validation command above.

No admin path, force-push, labels, or branch protection mutations were used.

PR #9324, opened 2026-07-15
fix(tests): recut #8277 self-healing agent registry + probed-branch smoke coverage

### LINKED ISSUE #8277: test pollution: pr_review_protocol smoke fails in debate-nz shard on main (AgentRegistry loses registrations under xdist)

scarmani:
## Summary

Recut of stranded commit `73e9b4b37d` from branch `worktree-fix-8277-registry-pollution` (authored Jun 12, never merged — surfaced by the Jul 15 believed-merged stranded-work audit alongside the rlm dotenv guard, which is being recovered separately in #9319).

**Supersession check performed:** the original commit had two halves.
- The metadata-only smoke-test fix was independently superseded on main by #8882 (Jul 5) — **not** re-applied; main's assertion style kept.
- The pieces below never reached main and are carried here.

## Changes

- **Self-healing agent registry** (defense-in-depth for issue #8277): `@AgentRegistry.register` records each spec on the agent class, and `register_all_agents()` re-applies specs from already-imported `aragora.agents.*` modules, so a cleared or rebound `AgentFactory._registry` recovers even though cached modules never re-run their decorators. Existing entries are never overwritten; test-defined agents (in `tests.*`) are never resurrected.
- **`test_pr_review_protocol_packet_probes_providers_for_live_results`**: passes `execution_failures` so the probed branch keeps the original provider-selection assertions (#8185 removed this coverage from the metadata-only path). Provider expectations verified unchanged on current main.
- 3 registry self-healing tests in `tests/agents/test_registry.py`.

## Conflict resolution

1-file cherry-pick conflict in `tests/debate/test_pr_review_protocol_smoke.py`: kept main's (#8882) assertions for the metadata-only test, grafted the probed-branch test on top, factored shared `_PACKET_KWARGS` / `_mock_provider_availability` helpers.

## Validation

- `pytest tests/debate/test_pr_review_protocol_smoke.py tests/agents/test_registry.py tests/agents/test_agent_registry.py` — 110 passed
- `ruff check` + `ruff format --check` clean on all three files

Closes #8277

### DIFF
--- a/aragora/agents/registry.py
+++ b/aragora/agents/registry.py
@@ -156,6 +156,15 @@ class AgentFactory:
                 accepts_api_key=accepts_api_key,
             )
             cls._registry[type_name] = spec
+            # Record the spec on the class itself so register_all_agents()
+            # can self-heal a cleared/rebound registry without re-importing
+            # (decorators never re-run while the module is cached in
+            # sys.modules). vars() lookup avoids inheriting a parent's specs.
+            own_specs = vars(agent_cls).get("__registry_specs__")
+            if own_specs is None:
+                own_specs = []
+                setattr(agent_cls, "__registry_specs__", own_specs)
+            own_specs.append(spec)
             return agent_cls
 
         return decorator
@@ -463,6 +472,11 @@ def register_all_agents() -> None:
 
     This function should be called once at startup to ensure
     all agents are registered before create() is used.
+
+    Safe to call repeatedly: if the registry was cleared (or rebound) after
+    the agent modules were first imported, the registrations are re-applied
+    from specs recorded on the agent classes, since the cached modules'
+    @AgentRegistry.register decorators never re-run.
     """
     # Import modules to trigger @register decorators
     # These imports are intentionally side-effect only
@@ -480,3 +494,24 @@ def register_all_agents() -> None:
         from aragora.agents import api_agents  # noqa: F401
     except ImportError:
         logger.debug("api_agents module not available for registration")
+
+    _heal_registry()
+
+
+def _heal_registry() -> None:
+    """Re-apply registrations recorded on already-imported agent classes.
+
+    Only production modules (``aragora.agents.*``) are scanned, so agent
+    classes registered ad hoc by tests are never resurrected. Existing
+    entries are never overwritten, so deliberate overrides survive.
+    """
+    import sys
+
+    for module_name, module in list(sys.modules.items()):
+        if module is None or not module_name.startswith("aragora.agents"):
+            continue
+        for obj in list(vars(module).values()):
+            if not isinstance(obj, type):
+                continue
+            for spec in vars(obj).get("__registry_specs__", ()):
+                AgentFactory._registry.setdefault(spec.name, spec)
--- a/tests/agents/test_registry.py
+++ b/tests/agents/test_registry.py
@@ -771,6 +771,55 @@ class TestRegisterAllAgents:
             AgentRegistry._registry.clear()
             AgentRegistry._registry.update(original)
 
+    def test_register_all_agents_self_heals_cleared_registry(self):
+        """A cleared registry is repopulated even when agent modules are cached.
+
+        Regression test for #8277: under xdist a test that clears (or rebinds)
+        AgentFactory._registry used to permanently break agent lookup for the
+        rest of the worker, because aragora.agents.cli_agents stays cached in
+        sys.modules so its @AgentRegistry.register decorators never re-run.
+        """
+        from aragora.agents.registry import AgentRegistry, register_all_agents
+
+        register_all_agents()
+        assert AgentRegistry.is_registered("claude")
+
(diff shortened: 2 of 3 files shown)


scarmani:

Head: f20741e (f20741ebd614d804212208cb7208cf299b78dcc6), committed 2026-07-15T21:01:34Z.
PR: #9324.
Model family: claude

Verdict: PASS

- [P3] `aragora/agents/registry.py:509` — `module_name.startswith("aragora.agents")` would also match a hypothetical sibling package like `aragora.agents_x`. None exists today; the conventional guard is `name == "aragora.agents" or name.startswith("aragora.agents.")`.
- [P3] `aragora/agents/registry.py:502` — docstring overclaims: "agent classes registered ad hoc by tests are never resurrected" holds only for test-*defined* classes. A test that registers a *production* class (defined in `aragora.agents.*`) under an ad-hoc name stamps `__registry_specs__` permanently, and the next `register_all_agents()` resurrects that name process-wide. No current test does this, but the docstring invites reliance on a guarantee the code doesn't fully provide.
- [P3] `aragora/agents/registry.py:497` — heal is add-only and asymmetric: deliberate *overwrites* survive (`setdefault`) but deliberate *removals* (`pop`/`del`) are resurrected on the next `register_all_agents()`. No production code removes agent entries today (the `del cls._registry` hits are in unrelated registries), so this is latent, not live.
- [P3] `aragora/agents/registry.py:165` — `own_specs.append(spec)` is unconditional, so re-registering the same class under the same name accumulates duplicate specs, and heal's `setdefault` then restores the *first* recorded spec while live registration is last-write-wins. Unreachable divergence in production (each of the 46 names registers exactly once, verified empirically), but a dedupe-by-name on append would close it.
- [P3] `tests/debate/test_pr_review_protocol_smoke.py:81` — the expected status is composed as `f"{PROTOCOL_STATUS}_fallback_execution_failed"` while production composes `f"{FALLBACK_PROTOCOL_STATUS}_execution_failed"`; they coincide only because `FALLBACK_PROTOCOL_STATUS == PROTOCOL_STATUS + "_fallback"` today. Importing `FALLBACK_PROTOCOL_STATUS` directly would be less brittle.

Verified explicitly: `AgentRegistry = AgentFactory` alias makes the rebound-registry test coherent; `spec.name == type_name` for all 46 entries so heal keys match registry keys; every registered `agent_class` is reachable via `vars()` of an `aragora.agents*` module so heal has no silent coverage gaps; the conditional kimi-cli registration replays consistently (spec recorded only if the env gate was set at import); asserted provider slots in the new live-results smoke test are deterministic under the mocks (`shutil.which` fully patched, codex/grok CLIs absent, slot3 deliberately unasserted), and both smoke tests pass against unchanged protocol code.

dogfood: yes

scarmani:

Head: f20741e (f20741ebd614d804212208cb7208cf299b78dcc6), committed 2026-07-15T21:01:34Z.
PR: #9324.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9326, opened 2026-07-15
fix(routing): drop undeclared numpy dependency from lara_router

scarmani:
## Summary
- `aragora/routing/__init__.py` eagerly imports `lara_router.py`, which had a top-level `import numpy as np` — but numpy is **not declared in pyproject.toml**. Any environment without numpy could not import `aragora.routing` at all, and every `tests/routing/` module failed collection with `ModuleNotFoundError`.
- numpy was only used for two `np.mean()` calls over lists of floats in `LaRARouter.get_metrics()`. This PR removes the numpy import entirely and uses `statistics.fmean` from the stdlib — same math, no dependency.

## Why this approach
Repo convention treats numpy as optional: `aragora/debate/similarity/backends.py` guards it with `try/except` + `HAS_NUMPY`, and `convergence/detector.py` / `evidence/claim_check.py` import it lazily inside functions. Since lara_router's usage is trivial, dropping the dependency outright is cleaner than adding a guard, and avoids declaring numpy as a core dependency for two mean() calls.

## Verification
- `aragora.routing` imports cleanly with numpy blocked via a `sys.meta_path` hook (simulating a numpy-less env)
- `tests/routing/`: 577 passed; 3 pre-existing failures in `test_auto_routing.py` (`KeyError: 'kimi'`, unrelated to this change)
- LaRA suites (`tests/research/test_lara_router.py`, `tests/scripts/test_bench_lara_routing.py`): 27 passed
- `get_metrics()` exercised end-to-end — output identical (plain `float` values as before)
- mypy + ruff clean on the changed file

### DIFF
--- a/aragora/routing/lara_router.py
+++ b/aragora/routing/lara_router.py
@@ -15,13 +15,12 @@ Factual lookups -> RAG, Complex reasoning -> RLM, Connected concepts -> Graph.
 """
 
 from dataclasses import dataclass, field
+from statistics import fmean
 from typing import Any
 from enum import Enum
 import logging
 import re
 import time
-
-import numpy as np
 
 logger = logging.getLogger(__name__)
 
@@ -551,8 +550,8 @@ class LaRARouter:
             "total_decisions": total,
             "mode_distribution": {mode: count / total for mode, count in mode_counts.items()},
             "mode_counts": mode_counts,
-            "avg_confidence": float(np.mean([d.confidence for d in self._decision_history])),
-            "avg_latency_ms": float(np.mean([d.duration_ms for d in self._decision_history])),
+            "avg_confidence": fmean(d.confidence for d in self._decision_history),
+            "avg_latency_ms": fmean(d.duration_ms for d in self._decision_history),
         }
 
 


scarmani:

Head: 63f233a (63f233a143a0633c37f2b772ae269081afc14622), committed 2026-07-15T21:03:46Z.
PR: #9326.
Model family: claude

Verdict: PASS

- [P3] `aragora/routing/lara_router.py:554-555` — Failure mode under concurrent mutation changes: `fmean` on a lazy generator raises `StatisticsError` if the history is drained mid-iteration (e.g. `reset()` from another thread clears the list between the line-535 empty-check and consumption), whereas the old eagerly-built list + `np.mean` degraded to `nan`. The class has no locking and isn't documented thread-safe, so this is informational only.

Checked and clean: the empty-history case is unreachable (early return at `get_metrics` line 535 guards it, so `fmean`'s empty-data `StatisticsError` can't fire in single-threaded use); these two calls were the file's only numpy usages, so dropping the import is safe; `fmean` accepts generators on all supported Python versions (repo uses 3.10+ syntax); return type stays a builtin `float`, matching the previous `float(np.mean(...))`, so serialization and type expectations are unchanged; `fmean` is also marginally more accurate and drops a heavyweight import from the routing path.

dogfood: yes

scarmani:

Head: 63f233a (63f233a143a0633c37f2b772ae269081afc14622), committed 2026-07-15T21:03:46Z.
PR: #9326.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9328, opened 2026-07-15
[preflight] worker check

scarmani:
Preflight validation of worker read/write/commit/push.


scarmani:
Preflight complete - closing.

PR #9346, opened 2026-07-16
fix(governance): provenance-classified drift inventory + per-PR delta ratchet [Tier-3]

scarmani:
## Summary

Compliant replacement for the Contract Drift Governance repair, superseding closed PR #9325 per the ruling in its closing comment:

> Resetting the baseline to 518 with 0% weekly reduction conflicts with the operator-approved non-weakening constraint, so this draft will not be settled or merged. The replacement repair will preserve the 655-item 2026-04-17 start and 10% weekly reduction, use a canonical provenance-classified live inventory, and fail closed without threshold inflation or unexplained debt absorption.

This PR follows those constraints exactly.

### The program schedule is untouched — and still red

`scripts/baselines/contract_drift_program.json` is byte-identical to main: **655 items @ 2026-04-17, −10%/week compounding**. Program mode today evaluates to target 185 vs 517 open items — **delta +332, FAIL**. That redness is correct and preserved; nothing in this PR absorbs the debt or moves the threshold.

(Measured note: the three baseline files sum to 518 raw entries, but one — `GET /api/podcast/episodes/{param}` in `python_sdk_drift` — is a literal duplicate string, so there are 517 unique open items. Both copies are 2026-04-17 cohort members; there are zero post-cohort additions on main.)

### What changes

1. **Canonical provenance-classified inventory** — `scripts/baselines/contract_drift_inventory.json`, built by the new `scripts/generate_contract_drift_inventory.py`. Every baseline entry gets a stable id, a class, a discovery date, and provenance:
   - All 517 currently-open items are verified original-cohort survivors (present in the baselines at cohort commit `5cc7a81b77`) → class `start_cohort`, provenance "2026-04-17 program baseline cohort".
   - Any future post-cohort item **requires** `--provenance` with a PR/issue link; without it the generator fails closed and lists the unexplained items. No silent debt absorption.
   - Items that leave the baselines are marked `resolved` with a date and are **never deleted** (audit trail). Resolution is verifiable because the sibling no-regression gates keep the baselines a superset of live violations, so baseline pruning only happens when items are actually fixed.
   - Deterministic sorted output; `--check` fails when the inventory is out of sync with the baselines.

2. **`scripts/check_contract_drift_ratchet.py` gains `--mode {program,pr}`** (default `program`; JSON output is a superset of the previous shape):
[...]

PR #9322, opened 2026-07-15
feat(routing): decision-stakes router — route by tier, record rationale (recut of #8393, #8233)

scarmani:
## Summary

Recut of stranded work for #8233: recovers the two content commits from `origin/claude/odr-decision-stakes-routing` (authored Jun 14, believed-merged but never merged — Jul 15 stranded-work audit) onto current `main`.

- `aragora/routing/decision_stakes_router.py` — `DecisionStakesRouter`: routes by decision tier (Tier 0-2 → cost-efficient via `COST_OPTIMIZED`/`BALANCED`, Tier 3-4 → frontier via `QUALITY_OPTIMIZED`, `min_quality` rising with stakes), delegates within-tier selection to the existing Pareto `CostQualityOptimizer`, and emits an auditable `RoutingRationale` in the canonical `aragora.routing_rationale/v1` schema (`selector=decision_stakes_pareto`, `pareto_optimizer_consulted=true`, cost honestly `recorded: false`).
- `tests/routing/test_decision_stakes_router.py` — 10 tests: tier→class policy, clamping, rising min_quality, low-tier→cost / high-tier→frontier selection, min_quality floor, budget constraint, frontier recorded as `models_considered`, receipt-shaped `to_dict`.

## Why recover instead of drop

- Issue #8233 is still **open** and `docs/FEATURE_GAP_LIST.md` still lists decision-stakes routing as a gap — nothing on `main` superseded this.
- The **consumer half already landed on `main`**: `scripts/measure_cost_per_settled_pr.py` reads `aragora.routing_rationale/v1` records. This PR is the missing producer half.
- Verified against current `main`: `CostQualityOptimizer.select_provider` / `get_pareto_frontier` signatures and `ProviderMetrics` fields all match; schema constant matches `scripts/auto_evidence_cycle.py`.

## Provenance and prior dissent (disclosure)

Prior PR #8393 (same two-file surface) was closed 2026-06-30 by queue-drain: exact-head settlement at `56b8040a` recorded CHANGES_REQUESTED from both Claude and Grok, and the drain contract forbids post-dissent repair without a new owner. The close message says "Reopen or recut only with a new owner prepared to address the model dissent." This recut takes new ownership; the dissent evidence itself has rotated out of the local evidence store, so the concrete objections are not recoverable — fresh quorum review on this PR should surface them again if they still hold. Cherry-picked commits: `5049a6cf2f` (router) + `56b8040ad5` (schema alignment to `aragora.routing_rationale/v1`).

## Testing

[...]

PR #9170, opened 2026-07-10
docs: defer #8858 outsider run behind a measured ≥8/10 quality bar

an0mium:
## Summary
Records the founder decision of 2026-07-10 as a formal replan (plan §4): **no external human is recruited for the #8858 outsider verification run until the project measures ≥8/10 on every dimension of the new [`docs/status/QUALITY_BAR.md`](../blob/docs/defer-outsider-quality-bar/docs/status/QUALITY_BAR.md)** — founder holistic assessment at decision time: ~4/10.

- **QUALITY_BAR.md** — eight dimensions, each tied to a *repeatable instrument* (not self-assessment): simulated stranger runs (clean VM, public docs only, adversarial persona, failures recorded before fixes), clean-env install round trips, docs-consistency scans, signed-receipt offline verification, settlement signal-integrity telemetry, the existing architecture/test ratchets, and single-key core-workflow checks. Baselines anchor to the 2026-06-12 codebase-health audit and this week's measured settlement telemetry; rubric awaits founder calibration at settlement.
- **Plan amendments (minimal):** W1 external outcome/exit/founder rows swap the outsider recruit for the stranger-sim gate + rubric calibration; the Jul 16 replan trigger is superseded by a *stranger-sim-by-Jul-16* trigger; the W3 EU AI Act bundle keeps Jul 30 (it never depended on the outsider).
- Conductor operator-context updated in parallel so no lane keeps escalating outsider recruitment.

Docs-only, Tier 0. The counterweight is recorded in the doc itself: internal scores drift optimistic, so instruments — not impressions — move the numbers.

#### Test plan
- [x] docs-only diff; `automation_pr_preflight.sh origin/main HEAD` ok
- [x] All amended cross-references resolve (QUALITY_BAR ↔ plan §4 ↔ W1)

Generated with [Devin](https://devin.ai)

PR #9351, opened 2026-07-16
fix(ci): RCP cancels queued advisory runs only — stop killing in-flight checks [Tier-4]

scarmani:
## Diagnosis (verified)

`.github/workflows/required-check-priority.yml` (RCP) is the "external cancellation actor" from `docs/governance/PR_RUN_CANCELLATION_DIAGNOSIS.md` — it was in-repo all along. On every `pull_request` event RCP runs 4 sweeps (10 s apart) that cancel every **queued AND in-progress** advisory `pull_request` run at the PR head that isn't in its keep-list.

**Smoking gun:** RCP run [29520862671](https://github.com/synaptent/aragora/actions/runs/29520862671) cancelled runs 29520863012 / 29520863349 / 29520863087 / 29520862378 on PR #9346 seconds after creation, with **no newer push**. Sweep log: `Canceled run 29520863012 (Module Tier Drift)` … `Sweep 1/4: considered 14, canceled 4`.

The other half of the churn is by design: the cancelled-run guardian (`pr-cancelled-run-guardian.yml` + `scripts/ci/required_workflow_manifest.json`) deliberately reruns **only** keep-listed workflows, so advisory reds persist as `cancelled` conclusions until a manual rerun. The conservative merge executor treats those reds as blocking, so **every long-lived PR** pays a manual-rerun tax before settlement.

**Settlement-latency impact (2026-07-15/16):** repeated manual reruns were required on PRs #9170, #9322, and #9346 solely to clear RCP-inflicted cancelled advisory checks.

## Fix (minimal)

Restrict RCP's cancellation to `status === 'queued'` runs only — never cancel `in_progress` runs.

- An in-progress run **already holds a runner**: cancelling it frees zero queue capacity, and these advisory runs finish in 1-5 minutes anyway. Cancelling only converts them into red `cancelled` conclusions the guardian will never rerun.
- **Queue-pressure relief is fully preserved:** queued advisory runs still yield to required checks, which is the entire point of RCP.
- Keep-list, sweep structure, fail-closed empty-keep-list guard, and required-context defensive check are all unchanged. No new `${{ }}` interpolation; injection-safety patterns preserved.

## Alternatives considered

1. **Keep-list expansion** (add the advisory workflows to `alwaysKeepWorkflowPaths`/manifest): would stop the churn but also exempts them from queue-yielding entirely, defeating RCP's purpose, and requires perpetual list maintenance for every new advisory workflow. Queued-only fixes the class, not the instances.
2. **Guardian-rerun expansion** (rerun cancelled advisory runs too): treats the symptom and *doubles* runner spend — RCP cancels an in-flight run, the guardian re-queues it from scratch, and RCP's next sweep may cancel it again (rerun loop risk). Queued-only removes the wasteful cancellation instead of paying to undo it.

Queued-only dominates both: smaller diff, no new spend, no list maintenance, and the guardian's conservative no-advisory-rerun policy can stay exactly as designed.

## Changes

- `.github/workflows/required-check-priority.yml` — status filter narrowed to `queued` with rationale comment (~line 211)
- `scripts/check_required_check_priority_policy.py` — the existing keep-list policy guard now also enforces the queued-only invariant (regression protection against reverting to the two-status filter)
- `tests/scripts/test_check_required_check_priority_policy.py` — 3 new tests: rejects the pre-fix `queued || in_progress` filter, rejects a missing filter, accepts double-quoted variant; fixture updated. 16/16 pass.
- `docs/governance/PR_RUN_CANCELLATION_DIAGNOSIS.md` — dated correction section retracting the "external actor" conclusion, naming RCP as the canceller with the run-29520862671 evidence, and noting the guardian's by-design no-rerun policy (M2 resolved)

## Test state

The cancellation loop itself is inline `actions/github-script` JS with no JS test harness; it is guarded indirectly via the Python policy checker above (the repo's established pattern for this workflow). `pytest tests/scripts/test_check_required_check_priority_policy.py` — 16 passed. Workflow YAML parses; extracted inline JS passes `node --check`.

## Tier-4

This PR modifies a workflow file. **Operator settlement required — do not auto-merge.**

## Round-2 completion: the guardian now clears queued-phase displacements (commit 4fdf8948c6)

Round-1 review (openai P2) correctly noted a **residual settlement tax**: even with RCP restricted to queued-only cancellation, a displaced queued advisory run still ends as a terminal red `cancelled` conclusion, because the guardian (`scripts/retrigger_cancelled_pr_runs.py`, cron `*/20`) reruns only manifest-protected workflows. This round completes the two-part design:

1. **RCP cancels only QUEUED advisory runs** (round 1) — in-flight work is never killed, so no run is ever cancelled mid-execution by priority sweeps.
2. **The guardian reruns exactly those queued-phase displacements, once** (this commit) — a cancelled run outside the protected manifest becomes rerun-eligible **iff it never executed**: a jobs probe (`GET /actions/runs/{id}/jobs`) must show no job/step with a non-null `started_at`. A run cancelled at e.g. the Checkout step has started steps and stays ineligible (that class no longer occurs, per part 1).

Together: **no advisory run ever ends terminally `cancelled` due to priority displacement, and no in-flight run is ever killed.** The rerun cannot loop with the canceller: once the rerun starts executing, RCP's queued-only rule can no longer touch it, and `run_attempt == 1` bounds the guardian to one rerun per run regardless.

Guard parity and budget:

- The new class passes **all** existing guards identically: `run_attempt == 1`, active (branch, sha) head pair, open non-draft PR association, TTL on cancellation time, supersession-by-newest-same-group, path-based matching.
- The jobs probe is **lazy** — fetched only for advisory candidates that survived every cheap guard, so API spend scales with surviving candidates (typically 0-4 per pass), not with `--max-runs 300`. Probe failures fail **closed** (run stays cancelled) and loud (stderr).
- Output classifies the new class as `queued_phase_advisory_displacement` vs `unexpected_required_cancellation`, so guardian reports distinguish the two.
- `tests/scripts/test_retrigger_cancelled_pr_runs.py`: 7 new tests (queued-phase selected + classified; checkout-phase never selected; protected behavior unchanged and never probed; attempt-2 bounded; laziness asserted via fetcher call counts; failed probe fails closed; never-executed signal matrix). **26/26 pass**; ruff clean.

This supersedes the round-1 "Alternatives considered #2" objection: rerunning *in-flight* cancellations would have doubled runner spend and risked rerun loops, but rerunning *never-executed* displacements re-queues work that was never paid for, and part 1 removes the loop risk.

### DIFF
--- a/.github/workflows/required-check-priority.yml
+++ b/.github/workflows/required-check-priority.yml
@@ -208,7 +208,15 @@ jobs:
               for (const run of runs) {
                 if (run.head_sha !== headSha) continue;
                 if (run.id === selfRunId) continue;
-                if (run.status !== 'queued' && run.status !== 'in_progress') continue;
+                // Only cancel QUEUED runs. An in-progress run already holds a
+                // runner, so cancelling it frees no queue capacity; it only
+                // converts a 1-5 minute advisory check into a red "cancelled"
+                // conclusion that the cancelled-run guardian deliberately does
+                // not rerun (its manifest covers required workflows only),
+                // creating a manual-rerun tax that stalls merge settlement on
+                // long-lived PRs. Queue-pressure relief is fully preserved:
+                // queued advisory runs still yield to required checks.
+                if (run.status !== 'queued') continue;
                 // Fail closed on a broken (empty) keep-list: cancel nothing.
                 if (keepListEmpty) continue;
                 passConsidered += 1;
--- a/docs/governance/PR_RUN_CANCELLATION_DIAGNOSIS.md
+++ b/docs/governance/PR_RUN_CANCELLATION_DIAGNOSIS.md
@@ -1,5 +1,47 @@
 # PR Workflow-Run Cancellation — Diagnosis & Mitigation Plan
 
+## Correction (2026-07-16): the canceller is in-repo — Required Check Priority
+
+The "external cancellation actor" conclusion below is **retracted**. The canceller
+is `.github/workflows/required-check-priority.yml` (RCP) itself. RCP runs on every
+`pull_request` event and performs 4 sweeps (10 s apart) that cancelled every queued
+**and in-progress** advisory `pull_request` run at the PR head not in its
+keep-list.
+
+Smoking gun: RCP run **29520862671** cancelled runs 29520863012 / 29520863349 /
+29520863087 / 29520862378 on PR #9346 seconds after creation, with no newer push.
+Its sweep log reads:
+
+```
+Canceled run 29520863012 (Module Tier Drift)
+…
+Sweep 1/4: considered 14, canceled 4
+```
+
+This also explains every "external actor" signature observed below: the raciness
+(a fast advisory run survives if it finishes before a sweep reaches it — hence
+Portability Lint's ~50/50 split), the selectivity (only non-keep-listed advisory
+`pull_request` runs), and why `push`-to-main runs are always green (RCP only
+triggers on `pull_request`). M2 ("identify & scope the external canceller") is
+resolved: no org-level App or account automation is involved.
+
+The rerun side of the loop is by design, not a gap: the cancelled-run guardian
+(`.github/workflows/pr-cancelled-run-guardian.yml`, driven by
+`scripts/ci/required_workflow_manifest.json`) deliberately reruns **only**
+manifest/keep-listed workflows — advisory cancellations "are that system working
+and stay cancelled" (see below). So every RCP-cancelled advisory run persisted as
+a red `cancelled` conclusion until a human rerun, which blocked the conservative
+merge executor on every long-lived PR (repeated manual reruns observed on
+PRs #9170, #9322, #9346 on 2026-07-15/16).
+
+**Fix applied:** RCP's sweep now cancels **queued runs only**
+(`if (run.status !== 'queued') continue;`). In-progress runs already hold a
+runner, so cancelling them saved nothing; queued advisory runs still yield to
+required checks, preserving all queue-pressure relief. The queued-only invariant
+is enforced by `scripts/check_required_check_priority_policy.py`.
+
+---
+
 Status: **diagnosis + plan (no code/CI changes in this doc).** Companion to
 `docs/governance/MODULE_TIER_DRIFT_GUARDIAN.md`, which established that the recurring
 `check`/`portability` reds on PRs are **cancellations, not drift**. This doc pins
--- a/scripts/check_required_check_priority_policy.py
+++ b/scripts/check_required_check_priority_policy.py
@@ -80,6 +80,16 @@ REQUIRED_CONTEXT_TO_WORKFLOW_PATH = {
     "TypeScript SDK Type Check": ".github/workflows/sdk-test.yml",
 }
 
+# The cancellation sweep must only ever cancel QUEUED runs. Cancelling an
+# in_progress run frees no queue capacity (the run already holds a runner)
+# and leaves a red "cancelled" conclusion on advisory checks that the
+# cancelled-run guardian deliberately does not rerun, forcing manual reruns
+# before merge settlement. See PR run 29520862671 (2026-07-15), which
+# cancelled in-flight advisory runs on PR #9346 seconds after creation.
+_QUEUED_ONLY_STATUS_FILTER_RE = re.compile(
+    r"if\s*\(\s*run\.status\s*!==\s*(['\"])queued\1\s*\)\s*continue;"
+)
+
 
 def _extract_js_set_items(workflow_text: str, set_name: str) -> list[str] | None:
     pattern = r"const\s+" + re.escape(set_name) + r"\s*=\s*new Set\(\[(?P<body>.*?)\]\);"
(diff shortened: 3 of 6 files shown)


scarmani:

Head: 1dc893b (1dc893ba493396def43ec3178248d14d48bc7cac), committed 2026-07-16T18:28:06Z.
PR: #9351.
Model family: claude

Verdict: PASS

- [P3] `.github/workflows/required-check-priority.yml:211` (new filter) — The queued-only invariant is best-effort, not airtight: `run.status` is snapshotted at list time, then `runHasRequiredContext(run)` may await several paginated API calls (check-suites, jobs fallback) before `cancelWorkflowRun` fires. A queued run that starts in that window is still cancelled while in progress (GitHub's cancel API doesn't discriminate), reproducing the red-cancelled failure mode at low frequency. Worth a re-check of `run.status` immediately before cancel (re-fetch via `getWorkflowRun`) or at least softening the "must never be cancelled" wording in the policy error.
- [P3] `scripts/check_required_check_priority_policy.py` `_QUEUED_ONLY_STATUS_FILTER_RE` — Presence-only text check: it's satisfied by the literal line appearing anywhere in the workflow, including inside a JS comment or an unrelated loop, and it doesn't reject a future edit that re-adds in-progress cancellation in a second sweep alongside the kept line. Cheap hardening: additionally fail on any `run.status !== '…' && run.status !== '…'` / `'in_progress'`-tolerant filter pattern in the file.
- [P3] Workflow comment + `docs/governance/PR_RUN_CANCELLATION_DIAGNOSIS.md` — The rationale "cancelling an in-progress run frees no queue capacity" / "queue-pressure relief is fully preserved" is overstated: cancelling an in-progress advisory run does free its runner immediately, which is exactly the relief that matters on a saturated or degraded fleet (currently 4/12 runners down per #9361). The change is still the right tradeoff (red-cancelled + manual-rerun tax outweighs a ≤1–5-min runner hold), but the docs should state it as a tradeoff, not zero-cost — otherwise the next queue-pressure incident will look like a regression with no recorded justification.

Non-finding notes: all four `violations == []` fixtures route through the updated `_valid_workflow_text()`, so test consistency holds; `_write_policy_repo`'s path extraction is unaffected by the added snippet; the fail-closed empty-keep-list guard and required-context defensive keep are preserved unchanged.

dogfood: yes

scarmani:

Head: 4fdf894 (4fdf8948c68eb94129e113ea4abd85aad4033b78), committed 2026-07-17T00:07:41Z.
PR: #9351.
Model family: claude

Verdict: PASS

- [P3] `.github/workflows/required-check-priority.yml:208` (comment) / `docs/governance/PR_RUN_CANCELLATION_DIAGNOSIS.md` — the stated rationale "an in-progress run already holds a runner, so cancelling it frees no queue capacity" is factually wrong: cancelling an in-flight run releases its runner immediately, which *is* queue relief on a 12-runner fleet. The real trade is bounded (advisory checks are 1–5 min) and the settlement-tax cost clearly dominates, so the change itself is sound — but "queue-pressure relief is fully preserved" overstates it; required checks can now wait up to one advisory-run duration for a runner. Worth correcting the comment/docs so a future reader doesn't re-litigate from a false premise.
- [P3] `scripts/check_required_check_priority_policy.py:417` — the queued-only invariant is presence-only: a workflow that keeps the literal `if (run.status !== 'queued') continue;` line while adding a second cancellation path (e.g. another sweep block with `!== 'in_progress'` logic, or an `||` variant) passes the policy. Consider also flagging any `'in_progress'` occurrence within the sweep region to make the invariant two-sided.
- [P3] `scripts/retrigger_cancelled_pr_runs.py` (`run_never_executed`) — residual race: RCP lists a run as `queued`, the run starts in the seconds before the cancel POST lands, cancellation hits mid-execution → started step present → guardian correctly declines → that run is still a terminal red requiring manual rerun. Rare and bounded, but the docs' claim that the mid-execution class "is eliminated upstream" is only almost-true; worth a one-line note.
- [P3] `scripts/retrigger_cancelled_pr_runs.py:199` (`list_run_jobs`, `max_pages=3`) — a >300-job matrix run gets a truncated jobs view; if only jobs beyond page 3 had started, `run_never_executed` misclassifies and issues one benign extra rerun. Negligible in practice (bounded once by the `run_attempt==1` marker), but an explicit truncation comment or a `total_count` cross-check would close it.

Checked hard for blocking issues and found none: the queued-only workflow change, policy regex (old `&& !== 'in_progress'` form correctly fails the match), lazy probe ordering (probe is last guard, protected runs never probed), fail-closed paths (`fetch_jobs=None`, probe `GitHubApiError` → None → skip), rerun bounding (`run_attempt != 1` rejects reruns even before probe spend), and the RCP↔guardian interplay (no livelock: worst case is cancel → one rerun → cancel-at-attempt-2 → permanently skipped) are all consistent with the tests, and the test matrix covers the empty-jobs, job-stamped-no-step, and checkout-started cases.

dogfood: yes

scarmani:

Head: 4fdf894 (4fdf8948c68eb94129e113ea4abd85aad4033b78), committed 2026-07-17T00:07:41Z.
PR: #9351.
Model family: openai

Verdict: PASS

- [P3] `scripts/retrigger_cancelled_pr_runs.py:196` `list_run_jobs()` caps the jobs probe at 3 pages. A large advisory matrix with started jobs beyond the first 300 jobs could be misclassified as `queued_phase_advisory_displacement` if the fetched pages show no started steps. Bounded to one rerun, but the “never executed” proof is only as complete as the pagination cap.

dogfood: yes

scarmani:

PR: #9351
Exact head: 4fdf8948c68eb94129e113ea4abd85aad4033b78
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #9351 is non-draft and live exact-head checks/merge-packet remain otherwise green.

PR #9352, opened 2026-07-16
fix(governance): explained discovered-intake path for pr-mode drift ratchet [Tier-3]

scarmani:
## Summary

Amends the contract-drift governance gate merged in #9346 (`scripts/check_contract_drift_ratchet.py`, `pr` mode). Design gap found on first contact: pr mode failed ANY increase in the five baseline counts, with no intake path for legitimately discovered debt. Concrete victim: #9332 adds 11 canary-probe-exposed orphan routes to `scripts/baselines/validate_openapi_routes.json` — newly VISIBLE debt with clear provenance — and could never pass pr-mode governance.

## The amendment

In `pr` mode, a count increase is now allowed **iff EVERY baseline entry that is new vs the base ref** exists in the head inventory as:

- born in this PR (no base-inventory history), `class=discovered`,
- provenance containing a PR/issue reference (`#123` or URL),
- `discovered_on` = a valid date.

The existing invariants — birth-state open, cohort-derivability, append-only, unique ids, provenance-reference format — already police forgery. The discovered batch then immediately starts its own -10%/week clock in **program** mode (already implemented in #9346).

Still FAILS:
- any increase with a new entry missing from the inventory or failing the checks above,
- **reopens**: an entry whose id has base-inventory history is a regression, not intake — it may not be re-absorbed as fresh debt (its batch clock has already been burning),
- **vacuous increases**: a count that rises with zero new ids (duplicate list entries) can never satisfy the EVERY-quantifier,
- any integrity violation (fails closed regardless of deltas, unchanged).

## Rationale

This implements the #9325 ruling's "no UNEXPLAINED debt absorption": the ruling bans *unexplained* absorption, and explained intake with auditable provenance + a fresh burn-down clock is the designed mechanism, not a loophole. Without it, the gate structurally punishes making debt visible — the opposite of what a drift inventory is for.

## Tests

New (`tests/scripts/test_check_contract_drift_ratchet.py`, "pr mode: discovered intake" section):
- increase with fully-inventoried discovered items **passes** (the #9332 shape; asserts the batch gets its own program-mode clock),
- increase with one uninventoried entry **fails**,
- increase with free-text (no PR/issue ref) provenance **fails**,
- increase with zero new entries (duplicate list entry) **fails**,
- honest reopen (original clock, legal transitions) **fails** as regression-not-intake.

[...]

PR #9332, opened 2026-07-15
fix(openapi): restore handler-route spec coverage and re-arm route gate [Tier-3]

scarmani:
## Problem

The `Validate handler route coverage` gate in `openapi.yml` reports 69 new `missing_in_spec` routes (live 70 vs 11 baselined) and 36 new `orphaned_in_spec` (39 vs 17) — but **CI shows green**: the step pipes the validator through `| tee` and GitHub's default bash has no `pipefail`, so the exit 1 has been masked since **Jan 26** (`62191e806c`). Drift grew unpoliced while the contract-drift governance check was also permanently red (fixed separately in #9325).

## Investigation findings

1. **Regenerating the spec does NOT close the gap.** A fresh generation produces a path set identical to the committed one — still 69/36. The spec is not stale.
2. **Root cause of the 69 missing:** the canonical-registry generator never covered endpoints declared via handler `ROUTES` / `*_ROUTES`. The pre-Mar-27 `openapi.json` (8.5MB) carried legacy autogenerated placeholders for them; the Mar 27 sync (#1449) overwrote it with generator output (7.1MB), silently dropping ~260 documented path families. The route gate exists to catch exactly this — it was disabled by the tee mask.
3. **All 69 missing routes are genuinely served**: they trace to `ROUTES` on handlers registered in `HANDLER_REGISTRY` (43 wildcard-derived detail routes, 26 exact paths). Zero come from decorators.
4. **The 36 orphans split**: 12 are **served** via `can_handle` logic without `ROUTES` declarations (validator blind spot: chat/knowledge/channel, codebase/quick-scan(s), debates/public, devices/apple/shortcuts, knowledge/mound/graph, memory/tier, quotas/request-increase, review-queue ×2, marketplace/templates ×2); 24 are **dark** — documented in the canonical registry but no registered handler routes them (verified: e.g. gusto handlers exist but were never registered; `/api/v1/patterns*` documents what handlers serve at `/api/v1/workflow/patterns`).

## Changes

- **`scripts/generate_openapi.py` — Tier 4 handler-ROUTES supplement**: emits explicit `x-autogenerated` / `x-method-inferred` / `x-aragora-stability: experimental` placeholder operations for every handler-served route the spec doesn't cover (wildcards → named path params, internal route prefixes excluded, validator-consistent normalization). `missing_in_spec` 70 → **0**, coverage **100%**. Structurally prevents recurrence: any future undeclared handler route is auto-documented on regeneration.
[...]

PR #9360, opened 2026-07-16
fix(openapi): route-gate re-arm recut — internal-route policy + derived_verbs fixes (recut of #9332)

scarmani:
Recut of #9332 (closed) onto fresh main, fixing both P2 findings from the dual-model CHANGES-REQUESTED dissent (Tier-4 collect run grounded on #9332 head 47e5131; verdicts were never posted to the PR — the on-PR comments show an earlier PASS round).

> **⚠️ Contains workflow changes — approval-required surface, Tier-4, needs human preapproval before merge.** `.github/workflows/openapi.yml` gains `set -o pipefail` on the "Validate handler route coverage" step (the `| tee` pipe has masked the validator's exit 1 since Jan 26, `62191e806c`). No branch-protection or runner-label changes.

> **Governance:** pr-mode contract-drift ratchet currently fails on `routes_orphaned_in_spec` 17 → 38 (+21 explained discovered entries) — this is the exact shape #9352 (discovered-intake ratchet amendment) legalizes, and #9332 is the named victim in its description. **Governance rerun deferred until #9352 merges.** Ratchet *integrity* already passes: all new baseline entries are registered in `scripts/baselines/contract_drift_inventory.json` as `class=discovered` with PR-referenced provenance and `discovered_on=2026-07-16`.

### 1. Claude P2 — internal routes published into the public spec, invisibly to the gate

> **[P2] Internal route families newly published into the public spec, and the re-armed gate can never flag them** — `scripts/generate_openapi.py:620` applies the internal-route policy only to the Tier-4 supplement (`_collect_handler_routes_supplement`, line ~911); Tier-1/2/3 outputs bypass it entirely. The regenerated `docs/api/openapi.json` committed in this PR adds 12 unversioned `/api/control-plane/*` paths […] plus 7 `/api/v1/sme/*` paths, growing internal-family paths in the published contract from 106 to 122 vs the merge base. Because `validate_openapi_routes.py` normalizes […] and then excludes internal prefixes from *both* sides of the comparison […] this leak is invisible to the gate forever. […] Fix: apply the internal-prefix filter (on normalized form) to the final merged path set before writing the spec, then regenerate all four spec artifacts and the TS types.

**Fix (as prescribed):** new `_filter_internal_paths()` applies the fail-closed internal-route policy (`scripts/baselines/internal_route_prefixes.json`, loader extracted to `_load_internal_route_prefixes()`) to the **final merged path set** — every tier, including the `--ast-only` branch — immediately before the spec is built. Result: **122 internal paths excluded** (the 106 pre-existing + 16 this PR would have added); matching semantics deliberately mirror the validator so the generator excludes exactly the set the gate excludes. All four spec artifacts + `openapi-types.ts` regenerated.

Knock-on, handled honestly rather than by baseline growth: `verify_sdk_contracts.py` and `check_sdk_parity.py` now load the **same policy file** and exclude internal-family SDK surface from public-contract drift/staleness (served-but-internal endpoints can no longer be vouched for by the public spec). Verify baseline **shrinks** 420 → 406 (14 internal entries pruned); sdk-parity stale budget improves to 39/54.

### 2. OpenAI P2 — class-level derived_verbs applied to uncovered generic routes

> **[P2]** `scripts/generate_openapi.py:923` / `docs/api/openapi_generated.yaml:54196` — The supplement applies class-level `derived_verbs` to the single *uncovered* generic route, not the handler's full route set. This publishes bogus methods when other generic routes are already covered. Concrete example: `EmailWebhookHandler` only serves GET for `/api/v1/bots/email/status` and POST for webhook paths, but the generated spec now advertises `GET /api/v1/bots/email/webhook/mailgun` […]; runtime `handle()` returns `None` for that path while only `handle_post()` handles Mailgun. This corrupts the public OpenAPI contract and generated SDK types.

**Fix:** verb derivation is now bounded by the handler's **total declared generic routes** (counted before the covered-skip), not by how many happen to be uncovered. A multi-generic-route handler with one uncovered leftover falls back to the per-path segment heuristic. Verified in the regenerated spec: `/api/v1/bots/email/webhook/mailgun` publishes **POST only**.

## Carried forward from #9332 (unchanged in intent)

- **Tier-4 handler-ROUTES supplement**: `missing_in_spec` 70 → **0**, route coverage **100%**; placeholders marked `x-autogenerated` / `x-method-inferred` / `x-aragora-stability: experimental`.
- **`validate_openapi_routes.py`**: `can_handle` orphan probe with canary check (`served_undeclared` report field).
- **`memory.store()`** SDK methods (Python + TypeScript) backed by the real `POST /api/v1/memory/store` handler.
- **Baseline**: `missing_in_spec` 11 → 0 (shrink); `orphaned_in_spec` 17 → 38 — the 21 dark wire-or-remove entries are conscious, inventoried debt (see governance note above), each family still needs an operator wire-or-remove decision.
- Regenerated: spec artifacts (CI-order), TS types, capability matrix (+ docs-site mirror), METRICS + generated doc blocks.

## Tier classification

**Tier 4** — includes a `.github/workflows/` change (pipefail) and touches governance baselines. Requires operator human settlement; do not auto-merge.

## Local verification

| Check | Command | Result |
|---|---|---|
| Route gate | `validate_openapi_routes --fail-on-missing --json` (with pipefail) | exit 0 — coverage 100%, 0 new missing, 0 new orphaned |
| Docs coverage | `audit_openapi_docs --min-coverage 90` | 99.8% |
| SDK contracts | `verify_sdk_contracts --strict` | PASS, 0 regressions (baseline 420→406 shrink) |
| SDK parity | `check_sdk_parity --strict --budget` | PASS — missing_from_both 0/0, stale_python 39/54 |
| Namespace/cross parity | `check_sdk_namespace_parity --strict` / `check_cross_sdk_parity --strict` | PASS / PASS |
| Drift ratchet | `check_contract_drift_ratchet --mode pr --base-ref origin/main --strict` | integrity PASS; delta fails only on orphaned +21 → **#9352-gated** |
| TS types | `sdk/typescript: tsc --noEmit` (pinned 6.0.3) | PASS |
| Targeted tests | `pytest tests/scripts/test_{generate_openapi,verify_sdk_contracts,check_sdk_parity,validate_openapi_routes,check_contract_drift_ratchet}.py` | **78 passed** |
| OpenAPI suite | `pytest tests/server/openapi` | **479 passed** |
| SDK suite | `pytest sdk/python/tests` | **4161 passed, 6 failed** — all 6 fail identically on pristine origin/main (pre-existing, unrelated: boss_cycle, capability-matrix CLI count, charter compliance, elves refs, work-loss, docs-site links) |
| Lint | `pre-commit run --files <changed>` | all hooks Passed |

New tests (never weakening existing ones): internal-family filtering across tiers + fail-closed policy load (`test_generate_openapi.py`), EmailWebhookHandler-style verb-union regression (`test_generate_openapi.py`), internal-family exclusion in contracts/parity checkers (`test_verify_sdk_contracts.py`, `test_check_sdk_parity.py`).

Relation to the #9332 close rationale: the supersession asked for "no guessed methods" — fix 2 removes the bogus-verb class; the internal-route leak fix removes the other unsound published surface. The Tier-4 pipefail workflow change is kept in this PR (flagged above) rather than split; the operator can request a split at settlement.


scarmani:
## Maintainer response to round-2 review (openai P2: Python extractor misses `self._client.request(`)

**Verified and measured — the finding is real and much larger than the flagged instance.** `PY_REQUEST_RE` matches only `self._client._request(`; 6,782 of ~7k call sites in `aragora_sdk/namespaces/` use `.request(`. The extractor currently sees 128 of 2,642 Python SDK endpoints, and the corrected regex exposes **488 NEW `python_sdk_drift` entries** (63 visible → 551).

**Deliberately split out of this PR.** The blindness is pre-existing (not introduced here — `memory.py`'s new methods follow the dominant `.request(` form used by 96% of the SDK), and fixing it forces a 488-item discovered-intake that nearly doubles the open drift inventory (513 → ~1,001) and starts its own -10%/week burn-down clock (~49 items/week). That is a program-reshaping governance commitment requiring its own operator-settled Tier-3 PR, per the same integrity-first sequencing that produced this recut. A tracked follow-up task has been created with the full measurement and fix plan; this PR's round-1 fixes (method-aware supplement, exact internal-prefix matching) remain in scope because they were introduced-or-touched here.

Head at time of measurement: 4b30ecac2793522f28cf988bad57941cb0d5e179.

scarmani:
## Maintainer response to round-3 review (openai P2: pipefail makes the OpenAPI workflow fail at head)

**Does not reproduce — refuted by both local reproduction and real CI at this exact head (4b30ecac27).**

Local, running the workflow's exact command (`set -o pipefail; python scripts/validate_openapi_routes.py --spec docs/api/openapi.json --fail-on-missing --baseline scripts/baselines/validate_openapi_routes.json --internal-prefixes scripts/baselines/internal_route_prefixes.json --json`):

```
exit=0
coverage_percentage: 100.0, missing_in_spec_count: 0,
new_missing_in_spec_count: 0, new_orphaned_in_spec_count: 0
```

CI: the **OpenAPI Spec** workflow — which executes that exact step with pipefail armed — completed **success** at this head: https://github.com/synaptent/aragora/actions/runs/29537177719

The cited routes (`/api/v1/backups/cleanup`, `/api/v1/workflows`, `/api/v2/backups`) are baselined in `scripts/baselines/validate_openapi_routes.json` on this branch; a 15-orphan result is consistent with running the validator against a checkout missing this branch's regenerated spec/baseline, not against head.

scarmani:
## Maintainer response to round-4 review

**P1 (route gate fails at head with 15 orphans) — repeat of the round-3 claim, refuted again with three independent proofs at head 4b30ecac27:**

1. The exact workflow command (with `set -o pipefail`) exits 0 locally: `new_orphaned_in_spec_count: 0`, coverage 100%.
2. Real CI ran that exact step at this head: **OpenAPI Spec workflow = success** (https://github.com/synaptent/aragora/actions/runs/29537177719).
3. The three cited routes (`/api/v1/backups/cleanup`, `/api/v1/workflows`, `/api/v2/backups`) are present in `docs/api/openapi.json` at head and are neither flagged nor baselined — the validator resolves their handlers. A 15-orphan result is consistent with a review sandbox where some handler imports fail (shrinking the served-route set), not with this head.

**P2 (`DELETE /api/v1/memory/store` is a phantom operation) — verified REAL, root-caused, and split to a tracked follow-up:**

The phantom op is minted by `aragora/server/openapi_impl.py::_collect_autogenerated_paths` (Tier-1 canonical registry), which applies each handler's class-wide verb union to every one of its routes — `MemoryHandler.handle_delete()` exists for `/memory/continuum/*`, so `/memory/store` inherits a DELETE it never serves. It is **pre-existing**: `DELETE /api/v1/memory/store` is already in `origin/main:docs/api/openapi_generated.json`; this PR only promotes the artifact that main's own workflow auto-promotes. Measured scale: 217 published placeholder delete/put/patch operations are suspects — a server-side fix with spec-wide churn that needs its own Tier-3 cycle. A tracked follow-up task has been created with the root cause, measurement, and the fix pattern (port the Tier-4 supplement's `total_generic_declared == 1` conservatism, added IN THIS PR, to the canonical path).

This PR's own supplement code already refuses exactly this bug class ("the class verb set is the union across all of them and would advertise nonexistent operations" — `scripts/generate_openapi.py`). Head at time of measurement: 4b30ecac2793522f28cf988bad57941cb0d5e179.

scarmani:

Head: 4b30eca (4b30ecac2793522f28cf988bad57941cb0d5e179), committed 2026-07-16T21:44:48Z.
PR: #9360.
Model family: claude

Verdict: PASS

Verification performed at head 4b30eca (local worktree): route-coverage gate exits green (100% coverage, 0 new missing, 0 new orphaned vs baseline), `check_sdk_parity --strict` passes (missing_from_both 0/0, stale_python 39/54 within budget), `verify_sdk_contracts --strict` passes (0 baseline regressions), all 73 tests in the changed test files pass, doc-claimed counts (2,896 paths / 3,348 ops) exactly match the regenerated spec, `openapi.json` and `openapi_generated.json` are byte-identical, and no internal-prefix path leaks into the published spec. The round-1 fixes hold up: `is_internal_route` exact-root matching (no more `/api/v1/smear` overmatch), method-wise `_merge_supplement` (no more path-level setdefault swallowing POST_ROUTES), the canary guard on `can_handle` probing prevents prefix-matchers from hiding real orphans, fail-closed policy loading is enforced and tested in all four consumers, and the `pipefail` re-arm is correct (the only other `| tee` in the workflow already had it).

- [P3] `scripts/generate_openapi.py:1040` + regenerated spec — 848 operations carry `x-method-inferred`, and some advertise verbs the server won't serve (e.g. spec publishes `delete` on `/api/v1/memory/store`, which only `handle_post` serves, and bare `/api/v1/memory` / `/api/v1/memory/tier` paths that no handler routes — only the `/*` wildcard forms exist in `MemoryHandler.ROUTES`). These are flagged `x-autogenerated` and baselined as the 38 tracked orphans, but public-spec consumers and codegen will see nonexistent operations until the wire-or-remove queue drains.
- [P3] `sdk/typescript/src/namespaces/memory.ts:530` — the new `storeEntry` duplicates the pre-existing client-level `storeMemory` (`sdk/typescript/src/client.ts:3032`) on the same endpoint; the old sibling sends `{key, value}`, which the handler rejects with 400 ("Missing required field: content"). The new method matches the handler contract (`{content, tier, importance}` → `{id, tier}`); the broken older method should be fixed or deprecated in a follow-up to avoid two divergent surfaces for one route.
- [P3] `docs/security/PENTEST_RFP.md:20` — the API-operation counts were refreshed to 3,348 but "216,016 automated tests" two lines above was left stale (other docs in this PR updated the same metric to 223,954).

dogfood: yes

scarmani:

Head: 4b30eca (4b30ecac2793522f28cf988bad57941cb0d5e179), committed 2026-07-16T21:44:48Z.
PR: #9360.
Model family: openai

Verdict: CHANGES-REQUESTED

- [P1] `.github/workflows/openapi.yml:375` / `scripts/baselines/validate_openapi_routes.json:3`: The re-armed route gate now fails against this PR’s own baseline. `validate_openapi_routes.py --fail-on-missing --baseline ...` reports `new_orphaned_in_spec_count: 15` including `/api/v1/backups/cleanup`, `/api/v1/workflows`, and `/api/v2/backups`; with `set -o pipefail`, this CI step exits nonzero.

- [P2] `docs/api/openapi.json:185453`: The generated public spec advertises `DELETE /api/v1/memory/store`, but `MemoryHandler.handle_delete()` only handles `/api/v1/memory/continuum/*` and returns `None` for `/api/v1/memory/store`. This publishes a non-existent destructive operation and can generate broken client methods/contracts.

dogfood: yes

scarmani:

Head: 4b30eca (4b30ecac2793522f28cf988bad57941cb0d5e179), committed 2026-07-16T21:44:48Z.
PR: #9360.
Model family: claude

Verdict: PASS

Verified at head 4b30eca in this worktree: all 73 tests in the five changed test files pass locally; the published spec contains zero internal-family paths (checked all 2,896 paths against the 5 policy prefixes with the gate's exact matching semantics); `openapi_generated.json` path set equals `openapi.json`; the 2,896/3,348 counts match every doc update; `bearerAuth` is defined in `components.securitySchemes` for the new placeholder ops; `set -o pipefail` is correctly placed in the same shell block ahead of the `| tee` pipe; the round-1 P2 fixes (method-aware supplement merge, exact internal-prefix matching, EmailWebhook-style class-verb attribution guard) are present and covered by tests; and the new SDK `store_entry`/`storeEntry` methods match a genuinely served handler contract (`POST /api/v1/memory/store` in `aragora/server/handlers/memory/memory.py:278` returns `{"id", "tier"}`, tier default `fast`, importance clamped). No TS code references the operation types removed from `openapi-types.ts`.

- [P3] `docs/api/openapi.json` — the newly published supplement/AST paths still carry phantom inferred operations, e.g. `DELETE /api/v1/memory/store` (the server would dispatch that as a delete of a memory entry literally named "store" via the `/api/v1/memory/*` wildcard, not a real operation). Marked `x-autogenerated`/`x-method-inferred`/experimental, and the class-verb fan-out shrink is already split into follow-up PR #9365 — but until it lands, the pentest-facing public contract (PENTEST_RFP.md cites 3,348 operations) overstates the served surface.
- [P3] `tests/sdk/test_openapi_sync.py:371` — `test_no_new_py_drift` did not get the internal-route-policy carve-out that the TS drift and stability-manifest tests received, despite the Python SDK's `control_plane.py`/`sme.py`/`sso.py` namespaces calling policy-excluded paths. It passes today only because the Python endpoint extractor is largely blind (the `.request(` vs `_request(` pattern gap tracked in PR #9362); when that extractor fix lands, this test will re-flag exactly the endpoints the policy excludes unless the same `_is_internal` filter is added.
- [P3] `scripts/check_sdk_parity.py:265` — `_INTERNAL_PREFIX_CACHE` is a process-lifetime module global with no invalidation; tests or callers that exercise `build_parity_report` with a monkeypatched policy file will silently reuse the first-loaded prefixes. Harmless in CI's one-shot invocation, but a test-isolation trap.

dogfood: yes

scarmani:

Head: 4b30eca (4b30ecac2793522f28cf988bad57941cb0d5e179), committed 2026-07-16T21:44:48Z.
PR: #9360.
Model family: openai

Verdict: PASS

- [P3] `scripts/verify_sdk_contracts.py:21` / `tests/sdk/test_openapi_sync.py:83` only extract `self._client._request(...)`, so the new `sdk/python/aragora_sdk/namespaces/memory.py:218` `self._client.request(...)` endpoint is invisible to those Python SDK/OpenAPI sync checks. This is a contract-test blind spot, not an immediate runtime break because `/api/v1/memory/store` is present in the spec.

dogfood: yes

scarmani:

Head: 638a489 (638a4894428f7eb54a9fdea78ea5e8af5677917e), committed 2026-07-16T23:58:51Z.
PR: #9360.
Model family: claude

Verdict: PASS

Verified against the worktree at head 638a489 (not just the diff): the committed spec contains zero internal-family paths (grep across all five policy prefixes, versioned and unversioned forms), `openapi.json` and `openapi_generated.json` are byte-identical, `is_internal_route` matching is boundary-safe (`/api/v1/smear` doesn't match `/api/v1/sme/`), the Tier-4 supplement's verb attribution is bounded (single-generic-route handlers only, bare-`get` rejected, internal filter applied to the final merged set in both normal and `--ast-only` paths), the new SDK `store_entry`/`storeEntry` methods match the real pre-existing handler contract (`MemoryHandler._store_memory`: POST `/api/v1/memory/store` under `handle_post` with `require_permission("memory:manage")` + mutation rate limit; body `{content, tier, importance?}` → `{id, tier}`, tier names match `MemoryTier`), the `--json` gate output stays parseable (suppression logs go to stderr), and the pipefail fix is correct for the `| tee` masking it targets.

- [P3] `docs/api/openapi.json:185453` — newly published paths carry phantom inferred operations, e.g. `DELETE /api/v1/memory/store` which the handler does not serve (store is POST-only; DELETE falls through to 404). These come from the *unchanged* canonical-tier class-verb fan-out in `aragora/server/openapi_impl.py:_autogenerate_missing_paths` (class verb set applied to every route), are flagged in-band (`x-autogenerated`, `x-method-inferred: true`, `experimental`), and the same defect class already exists in main's committed artifacts (this PR removes several such placeholders in `sdk/typescript/src/openapi-types.ts`). Known round-4 finding with remediation split to PR #9365; that follow-up should land promptly since these ops are now part of the enforced public contract.
- [P3] `scripts/validate_openapi_routes.py:63` — `load_internal_prefixes` is documented as fail-closed but silently drops individual invalid entries; it only exits when *no* valid prefix remains. A typo'd prefix (e.g. missing leading slash) would silently un-exclude that family: the generator would publish it and the coverage gate would stop excluding it symmetrically, so the leak stays gate-green. Consider erroring on any non-conforming entry.
- [P3] `scripts/validate_openapi_routes.py:34` — importing this module mutates `os.environ` (`setdefault("ARAGORA_USE_SECRETS_MANAGER", "false")`), and this PR widens the import surface: `check_sdk_parity.py`, `verify_sdk_contracts.py`, and `tests/sdk/test_openapi_sync.py` now import it, so any pytest-xdist worker touching those leaks the env var to subsequent tests in the worker. `setdefault` limits the blast radius, but given this repo's prior xdist env-pollution incidents, moving the mutation into `main()`/CLI entry would be safer.

dogfood: yes

scarmani:

Head: 638a489 (638a4894428f7eb54a9fdea78ea5e8af5677917e), committed 2026-07-16T23:58:51Z.
PR: #9360.
Model family: openai

Verdict: CHANGES-REQUESTED

- [P2] `.github/workflows/openapi.yml:371` / `scripts/baselines/validate_openapi_routes.json:3`: the committed spec and baseline fail the re-armed route gate. Running `python scripts/validate_openapi_routes.py --spec docs/api/openapi.json --fail-on-missing --baseline scripts/baselines/validate_openapi_routes.json` exits `1` with 15 `new_orphaned_in_spec` entries, including `/api/v1/backups/cleanup`, `/api/v1/workflows`, `/api/v1/workflows/templates`, and `/api/v2/backups`. With the new `set -o pipefail`, this PR will fail CI until those routes are either backed by handlers/filtered or added to the baseline intentionally.

dogfood: yes

scarmani:
## Maintainer note on the round-6 review (head 638a489442)

The openai reviewer's recurring claim that the re-armed route gate fails with 15 new orphans is again contradicted at THIS head by both proofs:

- **Real CI**: OpenAPI Spec workflow (the exact pipefail'd step) = **success** at 638a489442 — https://github.com/synaptent/aragora/actions/runs/29543478104
- **Local exact command**: exit 0, `coverage 100.0%, new_missing 0, new_orphaned 0`.

The claim recurred in rounds 3, 4, and 6 (dropped in round 5 after head-stamped refutation), and is consistent with a review sandbox where ~15 handler families fail to import/register, shrinking the served-route set. Recorded here as unresolved sandbox-environment dissent; proceeding to Tier-4 operator settlement per the gate's `human_risk_settlement_required` verdict, with this dissent preserved in the decision record.

scarmani:

PR: #9360
Exact head: 638a4894428f7eb54a9fdea78ea5e8af5677917e
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #9360 is non-draft and live exact-head checks/merge-packet remain otherwise green.

scarmani:

Head: dc75a75 (dc75a75102f4334b2c6b78638ef031fc53fb7c9f), committed 2026-07-17T00:34:12Z.
PR: #9360.
Model family: claude

Verdict: PASS

- [P3] `scripts/generate_openapi.py:744-760` — `_is_internal_route`/`_load_internal_route_prefixes` re-implement the matcher that `scripts/validate_openapi_routes.py:73` declares canonical ("every checker that applies the internal-route policy must delegate here"). The one consumer where a semantic drift would be a security leak (the spec publish filter) is the only one holding a copy instead of delegating; today the semantics are identical, but future edits to one file won't propagate to the other.
- [P3] `scripts/generate_openapi.py` (Tier-4 verb inference) — for handlers declaring more than one generic ROUTES path, class-verb derivation is disabled and the last-segment heuristic can still publish a GET placeholder for a POST-only route ("store", "login", etc. are not in `_ACTION_SEGMENTS`). Mitigated by `x-method-inferred: true` + `experimental` stability marking, and I found no wrong instance in the committed `docs/api/openapi.json` (scanned all GET-inferred supplements against action-like tails: zero hits), so this is residual-risk only.
- [P3] `scripts/baselines/internal_route_prefixes.json` — the policy holds just 5 families, and it is now the sole boundary between "served" and "published": every other handler family (GatewayCredentials, BackupOffsite, EmailWebhook, OAuthWizard, MFACompliance, DevOps, ...) is mass-published as bearer-auth placeholders by the Tier-4 supplement. That is this PR's stated intent per the explained intake, but any family that should have been private needed to be in this file *before* this merge; worth a one-time audit of the newly published tag list.

What I checked and found sound: the `is_internal_route` matcher is exact-root-or-slash-bounded (the round-1 `/api/v1/smear` overmatch is fixed and regression-tested); policy loading fails closed with `SystemExit` in all four consumers; `_merge_supplement` is method-wise `setdefault` so documented operations are never overwritten; the class-verb fan-out is bounded to single-generic-route handlers with the bare-`["get"]` dispatcher guard; `filter_served_orphans`' canary check correctly refuses prefix-matching `can_handle` claims (verified against `MemoryHandler`, which stays in the orphan baseline as expected); the shrunken `verify_sdk_contracts.json` baseline fails in the safe direction if a prefix is ever removed from the policy; and both SDK `store_entry` additions match the handler's actual contract (`{content, tier, importance}` → `{id, tier}`) and surrounding client conventions.

dogfood: yes

scarmani:

Head: dc75a75 (dc75a75102f4334b2c6b78638ef031fc53fb7c9f), committed 2026-07-17T00:34:12Z.
PR: #9360.
Model family: openai

Verdict: CHANGES-REQUESTED

- [P2] `scripts/generate_openapi.py:556` `--legacy-handlers` bypasses the new `_filter_internal_paths()` policy and returns `_legacy_generate_openapi_schema()` directly, so a user-facing generation mode can still publish internal route families the PR explicitly says must never enter the public spec.

- [P3] `scripts/check_sdk_parity.py:432` internal-family filtering is still exact-route based for SDK coverage and handler coverage, while only stale SDK paths use the new prefix policy. Running with undocumented routes still treats `/api/v1/control-plane/...` as public SDK gaps.

dogfood: yes

scarmani:
## Maintainer response to round-9 review (head 8926a2af3b)

- **claude [P1] duplicate `storeEntry`** — verified real (a stash-pop conflict silently restored the duplicate after the #9366 merge resolution) and **fixed at head 8926a2af3b**: one definition remains (main's contract-correct #9366 version), `tsc --noEmit` clean on the namespace.
- **claude [P3] stability-manifest phantom rows** — agreed; root fix is manifest pruning, tracked in the phantom class-verb-union follow-up chain (#9365 + its manifest regeneration). The 8 baselined rows keep the gate honest meanwhile and new regressions still flag.
- **claude [P3] fallback verb inference** — known, deliberately mitigated design (`x-method-inferred`/`x-autogenerated`/experimental markers); tightening further is in the same follow-up chain.
- **openai [P2] "15 new orphans"** — same environmental sandbox artifact as rounds 3/4/6/7: the exact gated command at every head, including CI's own OpenAPI Spec workflow runs, reports `new_orphaned_in_spec_count: 0` / exit 0. The cited routes are handler-backed at head.

scarmani:

Head: 8926a2a (8926a2af3b4582e05cf60f2370273e2f9ecc577e), committed 2026-07-17T02:11:42Z.
PR: #9360.
Model family: claude

Verdict: PASS

Review basis: read the full changed scripts at head 8926a2a (the truncated hunks in the prompt), ran all five changed test files (76 passed), and independently verified the artifacts: the emitted `docs/api/openapi.json` contains zero paths matching the internal-route policy (checked with the generator's own normalization), the 2,896-path/3,348-operation counts in the docs match the actual spec, the removed `DEFAULT_EXCLUDED_PREFIXES` constant has no remaining references, and `openapi-types.ts` has no dangling references to the removed control-plane/emergency operations.

Findings:

- [P3] `scripts/baselines/verify_sdk_contracts.json` (new `missing_stable` block): the 8 baselined entries (PATCH/POST on `/api/v1/debates/{param}/forks`, `/followups`, `/export/batch/{param}/results|status`) were present in main's spec union and are dropped by this PR. I verified they were phantom — `fork.py` declares these endpoints GET-only via `@api_endpoint`, and `debates/handler.py` serves only GET on export-batch results/status — so removing them from the spec is correct. But the fiction still lives in `aragora/server/openapi/stability_manifest.json` (lines ~1267–1274, 1463–1486); baselining the gap keeps 8 never-served operations advertised as "stable". Prefer deleting them from the manifest (or confirm the #9367/#9371 follow-up chain covers them).
- [P3] `scripts/baselines/internal_route_prefixes.json` + `generate_openapi.py:786` (`_normalize_for_comparison`): the policy can only ever match v1 families — unversioned routes normalize to `/api/v1/`, but `/api/v2/sso/*` and `/auth/sso/*` remain published (10 such paths, identical on main, so not a regression). If SSO is internal as a family rather than only its v1 surface, the "single enforcement point" filter has a version-shaped blind spot; if v2 SSO login/callback is deliberately public, consider a comment in the policy file saying so.
- [P3] `generate_openapi.py:1062` — every Tier-4 placeholder gets `security: [{bearerAuth: []}]`, including webhook/callback routes (e.g. mailgun, gusto callbacks) that authenticate via signatures, not bearer tokens. The `x-autogenerated`/`x-method-inferred`/`experimental` markers mitigate, but the published contract asserts an auth scheme the server doesn't use on those routes.
- [P3] `scripts/check_sdk_parity.py:39` vs `:268` — the script now carries two distinct "internal" notions: the pre-existing hardcoded `INTERNAL_ROUTES` exact-path set (used for coverage via `public_handler_paths`) and the new policy-file prefix families (used for staleness). They have different semantics and neither references the other; a one-line comment distinguishing them would prevent the next editor from "unifying" them incorrectly.

Not flagged as findings: the 574-item explained intake in `contract_drift_inventory.json` and the grown orphan baseline are the documented accepted-debt mechanism (every entry carries provenance/explanation, and the gates still fail on anything new); the fail-closed rewrite of `load_internal_prefixes` is propagated consistently to all four consumers with matching normalization, confirmed by the passing cross-checker tests.

dogfood: yes

scarmani:

Head: 8926a2a (8926a2af3b4582e05cf60f2370273e2f9ecc577e), committed 2026-07-17T02:11:42Z.
PR: #9360.
Model family: openai

Verdict: CHANGES-REQUESTED

- [P2] `scripts/baselines/validate_openapi_routes.json:3` The committed spec/baseline pair fails the route gate: `python3 scripts/validate_openapi_routes.py --fail-on-missing --baseline scripts/baselines/validate_openapi_routes.json --json` exits `1` with 15 `new_orphaned_in_spec` entries, including `/api/v1/backups/cleanup`, `/api/v1/workflows`, `/api/v1/workflows/generate`, and `/api/v2/backups`. This will break the OpenAPI CI path that runs the same validator with `--fail-on-missing`; either remove those orphaned spec paths, make the validator recognize their handlers, or ratchet the baseline intentionally.

dogfood: yes

scarmani:

Head: 623de66 (623de668a1b945b03121a4bdce4e815c658efd83), committed 2026-07-17T02:14:30Z.
PR: #9360.
Model family: claude

Verdict: PASS

I verified this beyond the diff text: fetched head 623de66 locally, read the full (truncated-in-prompt) hunks of `generate_openapi.py`, `validate_openapi_routes.py`, `verify_sdk_contracts.py`, and `check_sdk_parity.py`, and cross-checked the generated artifacts. Confirmations: the regenerated spec contains zero internal-family paths (control-plane/SSO/SME/emergency-admin grep = 0); `openapi.json` and `openapi_generated.json` are byte-identical (same blob); the documented counts (2,896 paths / 3,348 operations) reproduce exactly from the spec; the drift-inventory diff only adds entries and flips `status`/`resolved_on` — no `discovered_on`/`provenance` mutations; and the round-1/round-2 review P2s (prefix overmatch in `is_internal_route`, path-level setdefault swallowing method supplements, class-verb fan-out) are genuinely fixed with regression tests. The new `missing_stable` baseline entries are phantoms, not lost API: the server dispatches `POST .../followup` (singular, `handler.py:371`) while the plural path is GET-only, so the dropped placeholder ops never corresponded to served routes.

- [P3] `scripts/validate_openapi_routes.py` (`filter_served_orphans`): a `can_handle`-accepted path is treated as "served," but in this codebase `can_handle`=True with no matching `handle()` branch yields 500 `handler_no_result` — such dead spec paths are now structurally invisible to the orphan gate. The canary probe only defeats prefix-matchers, not list-but-don't-implement mismatches. Suppressions are logged and the 500-vs-404 remainder is tracked (#9371), but this is a permanent blind spot worth keeping on the books.
- [P3] `scripts/generate_openapi.py` (`_ACTION_SEGMENTS` fallback in Tier-4 supplement): when no verb evidence exists, the published operation's method is guessed from the last path segment, so a POST-only route whose tail isn't in the action list publishes as GET (and vice versa). Mitigated by `x-method-inferred`/`x-autogenerated`/`experimental` markers, but the public spec now contains guessed verbs that downstream SDK generation will faithfully propagate.
- [P3] `scripts/baselines/verify_sdk_contracts.json` (`missing_stable`): the 8 phantom PATCH/POST plural entries are grandfathered as baseline debt instead of being pruned from `aragora/server/openapi/stability_manifest.json`, which still advertises never-served "stable" endpoints. Follow-up should delete them from the manifest rather than carry the baseline forever.
- [P3] `scripts/check_sdk_parity.py:249`: importing this module now transitively executes `validate_openapi_routes`' module-level `os.environ.setdefault("ARAGORA_USE_SECRETS_MANAGER", "false")` — a new import side effect in any process that imports the parity checker (e.g. pytest). It's setdefault-only so low risk, but given this repo's test-pollution history, consider moving that env write under `if __name__ == "__main__"` in a follow-up.

dogfood: yes

scarmani:

Head: 623de66 (623de668a1b945b03121a4bdce4e815c658efd83), committed 2026-07-17T02:14:30Z.
PR: #9360.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

Head: 60d2eaa (60d2eaa27dbe70e3a608a0f730e3af9eb55c4a9f), committed 2026-07-17T04:39:32Z.
PR: #9360.
Model family: claude

Verdict: PASS

Verified at head 60d2eaa in a clean worktree: all 79 tests across the five changed test suites pass locally; the published spec contains zero paths from any internal policy family (leak filter works); README/METRICS counts (2,896 paths / 3,354 ops) match the actual `docs/api/openapi.json` exactly; and the manifest verbs removed in this round are confirmed phantom — `aragora/server/handlers/debates/routing.py:56-57,130-131` serves `/followups` and `/forks` as GET-only, so the deleted POST/PATCH entries never existed on the server.

Findings:

- [P3] `scripts/verify_sdk_contracts.py:52` (`_normalized_internal_prefixes`), `scripts/check_sdk_parity.py:279` (`_internal_prefix_families`), and `tests/sdk/test_openapi_sync.py:111` (`_internal_prefixes`) all normalize policy prefixes with `normalize_sdk_path`, which strips **every** version prefix (`scripts/sdk_path_normalize.py:24`). The v1-scoped internal prefix `/api/v1/sso/` therefore collapses to `/api/sso` and also swallows the deliberately-public `/api/v2/sso/*` family — 5 paths present in the published spec and called by `sdk/python/aragora_sdk/namespaces/sso.py:36-56`. This contradicts the version-scoping the policy itself documents (`scripts/generate_openapi.py:725-731`: "NOT other versions"). No wrong result today (v2 SSO is in the spec so no drift fires), but if the spec ever loses a v2 SSO endpoint, the SDK-side drift checkers will silently classify it internal and mask the regression — the exact masking class this PR series is eliminating. Consider matching internal families on the pre-version-stripped path in these three call sites.
- [P3] `scripts/baselines/verify_sdk_contracts.json` (`python_sdk_drift`): the phantom verbs removed from the manifest are still exposed as live SDK methods — `POST/PATCH /api/debates/{param}/followups`, `.../forks`, and `.../export/batch/{param}/results|status` were baselined as known drift rather than deprecated. Those SDK calls hit GET-only routes and will fail at runtime; fine as tracked debt given the ongoing SDK-fiction deprecation lane, but the baseline entries should not outlive that lane.
- [P3] `scripts/baselines/validate_openapi_routes.json`: `missing_in_spec` correctly drops to empty (gate re-armed), but `orphaned_in_spec` grew substantially (accounting callbacks, calibration, patterns, ralph, review-queue, slos, marketplace v2, …) — newly surfaced spec-without-handler debt accepted into the baseline rather than resolved. Acceptable only because burn-down tracking exists; the orphan side of the ratchet loosened in the same PR that tightened the missing side.
- [P3] `sdk/typescript/src/openapi-types.ts`: removal of the control-plane / admin-emergency / agent-dashboard path types is a compile-time break for any TS consumer importing those `paths[...]` keys. They were `@deprecated` autogenerated placeholders and internal-by-policy, so this is defensible, but it's a published-package surface change worth a changelog note.

dogfood: yes

scarmani:

Head: 60d2eaa (60d2eaa27dbe70e3a608a0f730e3af9eb55c4a9f), committed 2026-07-17T04:39:32Z.
PR: #9360.
Model family: openai

Verdict: PASS

- [P3] `scripts/validate_openapi_routes.py:437` — `coverage_percentage` still divides by raw `handler_routes` after internal routes are filtered out, so reports can understate public OpenAPI coverage.
- [P3] `docs/LANDING_PAGE.md:152` — SDK namespace counts remain stale (`184/183`) while the updated generated matrix reports `191/190`.
- [P3] `docs/security/PENTEST_RFP.md:20` — automated test count remains stale (`216,016`) while canonical `docs/METRICS.md` now reports `224,049`.

dogfood: yes

PR #9355, opened 2026-07-16
feat(models): canonical model catalog — phase 1, enforcement without rewiring

scarmani:
## Summary
The catalog design adjudicated across the #9073/#9075 review record, now real. Those reviews found model identity/pricing duplicated across **eleven runtime tables**, drifting one adversarial review round at a time — including three live provider reprices caught mid-review. This PR makes that detection mechanical.

**The three rules** (full doc: `docs/architecture/MODEL_CATALOG.md`):
1. **One typed source** — `aragora/models/catalog.py`: `ModelSpec` with canonical/direct/OpenRouter ids + aliases, USD-per-MTok pricing, context/output limits, release + 14-day soak dates; `by_any_id()` resolves any spelling. Eight live-verified frontier models to start.
2. **Offline validation, advisory liveness** — required CI validates against the committed `catalog_snapshot.json` (no network in required checks); `scripts/model_catalog_drift.py` is the advisory live differ with `--refresh` (drift today: 0/8).
3. **Governance stays out** — quorum-family *eligibility* remains in `quorum_evidence.py` under Tier-4 control; the catalog records soak dates but never decides authority.

**Phase 1 = enforcement without rewiring**: no runtime table changes behavior; `tests/models/test_catalog.py` (46 tests) asserts every existing row for an enforced model matches the catalog. Proof it works: on its first run it caught the grok-4.5/4.3 pricing rows missing from this lineage (they exist only on unmerged #9064) — fixed here with catalog-verified values. Known-stale legacy rows (deepseek-v4-pro, qwen3-max) deliberately await adjudication before enforcement rather than being silently overwritten.

**Stacked on `frontier-upgrade` (#9075)** — enforcement requires its mirror fixes. Phase 2 (consumers → generated projections) follows as bounded PRs.

#### Test plan
- [x] 46 catalog/enforcement tests green; 2760 across catalog+pdb+billing suites
- [x] `scripts/model_catalog_drift.py` → drift_count 0 against today's live catalog
- [x] ruff + mypy + gitleaks hooks green; preflight ok; METRICS/doc_stats/docs-site regenerated in-lineage

Generated with [Devin](https://devin.ai)

PR #9364, opened 2026-07-16
feat(routing): project catalog pricing into the router — first Phase 2 consumer

scarmani:
## Summary
Founder-directed first Phase 2 consumer of the canonical catalog (#9355): fixes the real routing bug where the Pareto/decision-stakes router estimated **$0 for every frontier pin** because `provider_config`'s hand rows went stale.

- `PROVIDER_PRICING` projects a row for every catalog model under every accepted spelling; projection **overrides** hand rows for cataloged ids (single source). Legacy non-catalog rows untouched.
- `get_estimated_cost` falls back to `by_any_id()` resolution; unknown models still return 0.0.
- Regression suite pins the bug class: every catalog spelling must price >0 through the router, at exactly catalog rates.

Stacked on #9355; merges after it.

#### Test plan
- [x] 623 routing+models tests green incl. 6 new projection regressions
- [x] Frontier pins verified pricing correctly: fable $60/2MTok, sol $35, grok-4.5 $8 (were all $0)

Generated with [Devin](https://devin.ai)


scarmani:
Current-main supersession assessment: #9374 merged the canonical catalog-backed routing pricing seam and its first consumers. This PR remains conflicting and is mostly superseded, but its narrow legacy `provider_config.py` compatibility residual is not. That residual is now preserved as #9390, with a preferred migration of legacy `get_estimated_cost` behavior to `resolve_model_pricing` plus focused ID/alias/unknown-fallback tests.

Do not restack the full #9364 branch; extract only #9390 against current main.

scarmani:

Head: e86b21b (e86b21b74d7e63dd442db8f412760e85063a7ac3), committed 2026-07-17T16:58:42Z.
PR: #9364.
Model family: claude

Verdict: PASS

Verification notes: reviewed the diff statically and cross-checked surrounding code at head e86b21b (optimizer exclude/cache semantics, `_round_robin_selection` gating, `DEFAULT_PROVIDER_ORDER` vs catalog spellings, other `PROVIDER_PRICING` consumers, ruff config, metrics-store defaults). Sandbox permission mode blocked executing the PR's test files; all claims below are from code inspection.

Confirmed sound (checked adversarially, no finding):
- No import cycle: `aragora.models.catalog` is stdlib-only; the mid-module import in `provider_config.py` is legal and E402 is in ruff's ignore list.
- The billing `PROVIDER_PRICING` (`aragora.billing.usage`) is a distinct dict; no cross-contamination. The only routing-side from-importers are the two documented compat sites.
- All-candidates-under-soak edge in `select_providers_for_debate`: optimizer returns `None` → break → round-robin padding, which is table-gated, so under-soak models cannot leak back in via the fallback.
- Optimizer routing cache is keyed on the frozen exclude set, so a soak-expiry date rollover changes the key and correctly misses stale cached selections.
- Stale-snapshot direction is provably conservative: `soak_until` is immutable and windows only expire, so a previous-day snapshot can only under-enumerate, never offer an under-soak model.
- `DEFAULT_PROVIDER_ORDER` entries don't collide with any catalog spelling at head, so the sweep can't shrink round-robin today; the new membership test pins this against catalog growth.
- New tests are hermetic: `ProviderMetricsStore()` is in-memory by default; real-table tests bind soak evaluation to the snapshot's own build stamp, avoiding UTC-midnight flakes; fake-clock fixtures patch both `utc_today` sites and register rebound globals with monkeypatch for restoration.

Findings:
- [P3] `aragora/routing/__init__.py:64` — `PROVIDER_ROUTING_PRICING` pins the import-day snapshot for process lifetime; any external consumer that enumerates it as a candidate surface gets frozen soak gating until restart (a model whose soak expired never appears). No in-repo consumer does this today and the comment warns, but it's a public re-export whose name doesn't signal the staleness contract — consider deprecating it toward `current_pricing_table()`.
- [P3] `aragora/routing/provider_config.py` (`current_pricing_table`) — returns the live published dict, "immutable-by-convention"; a single caller mutation corrupts the shared snapshot for every reader including from-importers, silently defeating the stale-but-never-corrupt guarantee. `types.MappingProxyType` over the snapshot would enforce the convention at near-zero cost.
- [P3] `aragora/routing/provider_config.py` (`_refresh_projection_if_stale`) — stamp-before-publish means a concurrent thread can observe `current_pricing_as_of() == today` paired with yesterday's table, so the "coherent as-of" guarantee is one-directional under concurrency. Benign in effect (strictly more-conservative gating for a microsecond window) and documented in the code; noting only because downstream comments describe the date as fully coherent.

dogfood: yes

scarmani:

Head: e86b21b (e86b21b74d7e63dd442db8f412760e85063a7ac3), committed 2026-07-17T16:58:42Z.
PR: #9364.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9365, opened 2026-07-16
fix(openapi): stop advertising phantom verbs on multi-route handlers

scarmani:
## Summary

`_collect_autogenerated_paths` in `aragora/server/openapi_impl.py` assigned each handler's class-wide implemented-verb union (`_infer_methods`) to **every** route in that handler's `ROUTES`/`DYNAMIC_ROUTES`/`ROUTE_PREFIXES` lacking explicit per-route methods. Multi-route handlers with mixed verbs therefore advertised phantom operations — e.g. the canonical spec published `DELETE /api/v1/memory/store` although `MemoryHandler.handle_delete()` only serves `/api/v1/memory/continuum/*` (returns `None` if called). Flagged by the openai quorum reviewer in round 4 on #9360.

This ports the conservatism `scripts/generate_openapi.py`'s Tier-4 supplement already applies to its own entries (the `total_generic_declared == 1` guard) to the canonical Tier-1 path:

- **Exactly one generic route, or a single-verb class set** → the class verb set is unambiguously attributable; keep it.
- **Ambiguous multi-verb union** → emit one conservative verb chosen **inside** the union (`get` if present, else `post`, else the first), since a verb outside the union is a guaranteed phantom. Measured shapes across `ALL_HANDLERS`: 1,554 routes single-verb (unchanged semantics), 59 multi-verb with get, 257 multi-verb without get (previously got GET placeholders under a naive fallback — guaranteed phantoms).

## Spec effect (pure shrink)

Against a spec regenerated from unmodified HEAD code: **253 phantom operations removed (3,303 → 3,050), zero operations added, zero paths added/removed.** "Autogenerated placeholder" delete/put/patch operations drop 232 → 16. Contract drift ratchet **pr-mode passes with +0 in every drift class** (count decrease needs no intake).

## SDK follow-through

Regenerating the spec newly documents the real `POST /api/v1/memory/store` (previously shadowed by the phantom union), pulling it into sdk-parity enforcement (budget 0). Added real SDK methods matching the server contract (`{content, tier?, importance?}` → `{id, tier}`):
- Python: `client.memory.store(content, tier=..., importance=...)`
- TypeScript: `client.memory.storeEntry(content, { tier, importance })`

⚠️ Pre-existing, out of scope: TS `client.storeMemory()` sends `{key, value, ...}`, which the server rejects (`Missing required field: content`) — needs its own fix.

## Regenerated artifacts

[...]

PR #9362, opened 2026-07-16
fix(sdk-gate): un-blind Python SDK contract extraction (self._client.request) + explained drift intake

scarmani:
## Summary

- `PY_REQUEST_RE` in `scripts/verify_sdk_contracts.py` (and the duplicated extractor in `tests/sdk/test_openapi_sync.py`) only matched the private `self._client._request(` form. ~6,780 of ~7k Python SDK call sites in `sdk/python/aragora_sdk/namespaces/` use the public `self._client.request(` form, so the extractor saw only 128 of ~2,6xx endpoints — Python SDK contract verification was ~96% blind. Flagged by openai quorum review round 2 on #9360 (measured 2026-07-16).
- Fix: extend both regexes to `self\._client\._?request\(`, with regression tests (synthetic dual-form content + extraction over the real `memory.py` namespace).
- Measured after fix: Python SDK 3,478 endpoint references extracted (was 128); exposes 640 total / 529 NEW `python_sdk_drift` entries vs `scripts/baselines/verify_sdk_contracts.json`.
- The new entries are intaken via the drift ratchet's explained-intake path (#9352): baseline update + `generate_contract_drift_inventory.py --provenance` referencing this PR. **Operator note (Tier-3 settlement):** this batch starts its own -10%/week burn-down clock in program mode (~53 items/week) — accepting this commitment is a knowing operator decision.

## Test plan
- [x] `pytest tests/scripts/test_verify_sdk_contracts.py tests/sdk/test_openapi_sync.py tests/scripts/test_check_contract_drift_ratchet.py`
- [x] `python scripts/verify_sdk_contracts.py --strict` passes against updated baseline
- [x] `python scripts/check_contract_drift_ratchet.py` in both `pr` and `program` modes
- [x] `python scripts/generate_contract_drift_inventory.py --check` in sync

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9366, opened 2026-07-16
fix(sdk): deprecate key/value memory methods, add contract-correct entry APIs

scarmani:
## Summary

The TS SDK's `storeMemory` / `retrieveMemory` / `deleteMemory` (and the `client.memory.store/retrieve/delete` namespace wrappers) target a key/value memory API the server never implemented:

| SDK method | Wire call | Server reality |
|---|---|---|
| `storeMemory(key, value, opts)` | POST `/api/v1/memory/store` with `{key, value, ...}` | Handler requires `{content, tier?, importance?}` → **400 "Missing required field: content"**; returns `{id, tier}`, not the declared `{stored, tier}` |
| `retrieveMemory(key)` | GET `/api/v1/memory/retrieve` | **Route does not exist** (only `/api/v1/memory/continuum/retrieve` does); the method swallows the 404 and **silently resolves to `null` on every call** |
| `deleteMemory(key, tier)` | DELETE `/api/v1/memory/delete` | **Route does not exist** (only `/api/v1/memory/continuum/{id}`); always rejects 404 |

None of these can ever have worked against this server, so there are no functioning callers to preserve.

## Decision: deprecate in place, add contract-correct parallels

Chosen over an in-place signature rewrite because the operating contract treats SDK signature changes/removals as an operator-review, never-auto-merge class — and because open PR #9365 already started down the parallel-method path with `storeEntry`. This PR is **additive**: no existing signature changes.

- `@deprecated` JSDoc on the three client methods + three namespace wrappers, each explaining the actual server contract and pointing at the replacement
- **New** `client.storeMemoryEntry(content, {tier?, importance?}) → {id, tier}` and namespace `memory.storeEntry(...)` — same public signature as #9365's parallel method, but delegating through the client for testability; the Python `MemoryAPI.store()` hunk matches #9365 verbatim, so the rebase either way is trivial
- **New** `client.deleteMemoryEntry(id) → {success, message}` / `memory.deleteEntry(id)` → DELETE `/api/v1/memory/continuum/{id}`
- **Bug fix** on the retrieve replacement: `retrieveFromContinuum` sent `q`/`tier` where the handler reads `query`/`tiers` (filters were silently ignored), and declared `{entries}` where the server returns `{memories, count}`. Params fixed; declared type now truthful (`retrieveContinuum` namespace type likewise). Note: this return-*type* correction changes no runtime behavior — the old type was unconditionally false — but flagging it for review since it can surface compile errors in downstream code that was already broken at runtime.
- Python SDK: matching `store()` + `delete_entry()` on sync and async `MemoryAPI`

## Relationship to PR #9365

#9365 adds namespace-level `storeEntry` (TS) and sync-only `store()` (Python) as a side item of its OpenAPI phantom-verb fix. This PR supersedes those two hunks with identical public signatures plus the retrieve/delete halves; whichever merges second has a trivial conflict.

## Verification

- `tsc` 6.0.3 `--noEmit`: clean
- `eslint src`: 0 errors (pre-existing `no-explicit-any` warnings untouched)
- `vitest`: 1717/1717 across 83 files (new tests for `storeMemoryEntry`/`deleteMemoryEntry`/`storeEntry`/`deleteEntry` wire contracts)
- `sdk/python` pytest: 26/26 (new sync+async `store`/`delete_entry` tests); mypy 2.1.0 clean
- `check_sdk_parity.py --strict --allow-missing`, `check_cross_sdk_parity.py`, `check_sdk_namespace_parity.py`: all PASS, 0 regressions

### DIFF
--- a/scripts/baselines/contract_drift_inventory.json
+++ b/scripts/baselines/contract_drift_inventory.json
@@ -947,6 +947,14 @@
       "status": "open"
     },
     {
+      "class": "discovered",
+      "discovered_on": "2026-07-16",
+      "id": "typescript_sdk_drift:DELETE /api/memory/continuum/{param}",
+      "provenance": "PR #9366: contract-correct deleteEntry calls DELETE /api/v1/memory/continuum/{id} (served by MemoryHandler.handle_delete) but the generated spec lacks the route (generator ROUTES gap, see #9360)",
+      "source": "typescript_sdk_drift",
+      "status": "open"
+    },
+    {
       "class": "start_cohort",
       "discovered_on": "2026-04-17",
       "id": "typescript_sdk_drift:DELETE /api/memory/snapshots/{param}",
--- a/scripts/baselines/verify_sdk_contracts.json
+++ b/scripts/baselines/verify_sdk_contracts.json
@@ -490,7 +490,8 @@
     "POST /api/webhooks/{param}/deliveries/{param}/retry",
     "POST /api/webhooks/{param}/events",
     "POST /api/webhooks/{param}/rotate-secret",
-    "PUT /api/webhooks/{param}/retry-policy"
+    "PUT /api/webhooks/{param}/retry-policy",
+    "DELETE /api/memory/continuum/{param}"
   ],
   "missing_stable": [
     "PATCH /api/v1/debates/export/batch/{param}/results",
--- a/sdk/python/aragora_sdk/namespaces/memory.py
+++ b/sdk/python/aragora_sdk/namespaces/memory.py
@@ -194,6 +194,41 @@ class MemoryAPI:
         """
         return self.list_critiques(agent=agent, limit=limit, offset=offset)
 
+    def store(
+        self,
+        content: str,
+        *,
+        tier: str = "fast",
+        importance: float | None = None,
+    ) -> dict[str, Any]:
+        """
+        Store a new memory entry in continuum memory.
+
+        Args:
+            content: The memory content to store
+            tier: Target memory tier (fast, medium, slow, glacial)
+            importance: Importance score between 0.0 and 1.0
+
+        Returns:
+            Dict with the stored entry ID and tier
+        """
+        body: dict[str, Any] = {"content": content, "tier": tier}
+        if importance is not None:
+            body["importance"] = importance
+        return self._client.request("POST", "/api/v1/memory/store", json=body)
+
+    def delete_entry(self, memory_id: str) -> dict[str, Any]:
+        """
+        Delete a continuum memory entry by ID.
+
+        Args:
+            memory_id: ID of the entry to delete (as returned by store())
+
+        Returns:
+            Dict with success flag and message (404 if the entry does not exist)
+        """
+        return self._client.request("DELETE", f"/api/v1/memory/continuum/{memory_id}")
+
     def store_critique(
         self,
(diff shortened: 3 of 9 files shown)


scarmani:

Head: 937e41a (937e41a54fb24c92206b62ff0b3c26c615a98278), committed 2026-07-17T00:34:03Z.
PR: #9366.
Model family: claude

Verdict: PASS

Verification performed against head 937e41a: server contracts match the SDK exactly (`_store_memory` expects `{content, tier?, importance?}` → `{id, tier}` at `aragora/server/handlers/memory/memory.py:307`; `handle_delete` serves `DELETE /api/v1/memory/continuum/{id}` → `{success, message}` at `memory.py:404` / `memory_continuum.py:331`; retrieve reads `query`/`tiers`/`limit` and returns `{memories, count}` at `memory_continuum.py:59-126`). `verify_sdk_contracts.py --strict` passes (py=0 ts=0 stable=0 new), `check_contract_drift_ratchet.py --mode pr` passes via explained intake (+1 TS drift with matching class=discovered inventory entry), Python SDK memory tests pass (26/26), `tsc --noEmit` is clean, `POST /api/v1/memory/store` is present in the generated spec while the DELETE route is genuinely absent (matching the single drift entry), and all in-repo consumers of the changed `{memories, count}` shape were updated.

- [P3] `sdk/python/aragora_sdk/namespaces/memory.py:229` and `:721` — `delete_entry` interpolates `memory_id` into the path unquoted, while the TS counterpart in this same PR uses `encodeURIComponent`. A caller-supplied ID containing `/` or `?` silently changes the request path. Other Python namespaces (receipts.py, costs.py, dashboard.py) do `urllib.parse.quote` path params; match that.
- [P3] `sdk/typescript/src/namespaces/memory.ts:319-324` — `MemoryClientInterface` gains `storeMemoryEntry`/`deleteMemoryEntry` declarations, but `MemoryAPI.storeEntry`/`deleteEntry` bypass them and call `this.client.request` directly (unlike the deprecated `store`→`storeMemory` delegation pattern). The interface members — and the `vi.fn()` mocks added for them in `memory.test.ts:45-46` — are dead. Either delegate or drop the declarations.
- [P3] `sdk/typescript/src/client.ts:3342` — `retrieveFromContinuum`'s return type change (`{entries}` → `{memories, count}`) and `q`→`query` param fix are compile-breaking for external TS consumers. Justified (the old `.entries` was always `undefined` at runtime and `q` was ignored by the server, so results were silently unfiltered), but this deserves an explicit changelog/semver note since previously-"working" code will now both compile-break and start actually filtering.
- [P3] `sdk/python/aragora_sdk/namespaces/memory.py:217`/`:229` — the new Python methods use `self._client.request(`, which the contract verifier's `PY_REQUEST_RE` (`self._client._request(`) cannot see (known py-extractor blindness, #9362). When #9362 lands, `python_sdk_drift` will need its own `DELETE /api/memory/continuum/{param}` baseline entry — worth noting in that PR so it doesn't surface as a surprise regression.

dogfood: yes

scarmani:

Head: 937e41a (937e41a54fb24c92206b62ff0b3c26c615a98278), committed 2026-07-17T00:34:03Z.
PR: #9366.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9092, opened 2026-07-10
fix(ci): propagate OpenAPI route validator exit status (#9082)

an0mium:
## Summary

Tier-4 workflow-policy repair for #9082. The `Validate handler route coverage` step now enables Bash `pipefail` before running `validate_openapi_routes.py | tee`, preserving `/tmp/route-coverage.json` while propagating a nonzero validator status.

The focused policy test parses the live workflow, pins `pipefail` before the validator pipeline, verifies the artifact path remains present, and executes a failing synthetic producer to prove the report is written while the pipeline returns the producer's nonzero status.

## Tier-4 authorization

Operator authorization: https://github.com/synaptent/aragora/issues/9082#issuecomment-4931712425

> Tier-4 Human Implementation Authorization
>
> Issue: #9082
> Exact affected main head: 43fea520f53b47c58b2d0085705aa6880d9ef375
>
> I authorize one isolated draft repair PR that preserves the route-coverage JSON artifact while propagating any nonzero validate_openapi_routes.py exit status, with focused workflow-policy regression coverage. Keep route-baseline changes separate.
>
> Do not remove/re-arm .aragora/merge_executor.halt, manually rerun CI, merge, mark ready, or modify branch protection.

The affected pipeline remained unchanged on implementation base `3c0c63a10999653b89715f05286346fb384288fc`.

## Verification

- `python3 -m pytest -q tests/workflows/test_openapi_route_coverage_policy.py tests/workflows/test_aragora_merge_quorum_workflow.py` -> `3 passed`
- `python3 -m ruff check tests/workflows/test_openapi_route_coverage_policy.py` -> `All checks passed!`
- `python3 -m ruff format --check tests/workflows/test_openapi_route_coverage_policy.py` -> `1 file already formatted`
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/openapi.yml', encoding='utf-8'))"` -> exit `0`
- `actionlint .github/workflows/openapi.yml` -> exit `0`
- `python3 scripts/check_workflow_checkout_integrity_policy.py` -> passed
- `python3 scripts/check_required_check_priority_policy.py` -> passed
- `bash -c 'set -o pipefail; false | tee /tmp/9082-pf-check >/dev/null; printf "exit=%s\\n" "$?"'` -> `exit=1`
- `git diff --check origin/main...HEAD` -> passed
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` -> `preflight: ok`
[...]

PR #9368, opened 2026-07-17
fix(ci): re-arm route-coverage gate with pipefail [Tier-4]

scarmani:
## Summary

3-line change: `set -o pipefail` before the piped route-coverage validator in `.github/workflows/openapi.yml`. Without it, `| tee` masks the validator's exit code — the gate has been silently non-enforcing since Jan 26 2026 while route drift accumulated (69 new missing / 36 new orphaned by Jul 15).

Split from #9360 per the #9332 close-out charter ("a method-aware truthful route-core PR plus separate Tier-4 workflow/pipefail changes") so the route-core PR is not classified Tier-4 by a workflow-file touch.

## Sequencing

Merge AFTER #9360: the re-armed gate requires the restored spec coverage + baselines that #9360 carries. At #9360's head the exact gated command passes (CI OpenAPI Spec workflow success, `coverage 100%, new_missing 0, new_orphaned 0`); on current main without #9360 the re-armed gate would fail honestly.

## Settlement

Tier-4 (workflow change): operator human-settlement required.

Refs: #9360, #9332, #9092 (prior propagate-exit-status issue).


scarmani:

Head: 35f854a (35f854ab57b7ecf64333b80765d93a66e7257daf), committed 2026-07-17T12:38:26Z.
PR: #9368.
Model family: claude

Verdict: PASS

Verified against the full file at head 35f854a, not just the diff hunk:

- The step (`generate-run` job, "Validate handler route coverage", `.github/workflows/openapi.yml:371`) has no step-level `shell:` key, so GitHub Actions runs it as `bash -e {0}` — pipefail OFF by default. The `| tee /tmp/route-coverage.json` on line 381 therefore masked the validator's exit code exactly as the comment claims, and `set -o pipefail` is the correct, necessary fix (the implicit `-e` then aborts the step when the pipeline fails).
- Failure mode is sound: on validator failure the step now dies at the pipeline; the follow-up stat-extraction lines are skipped, and the JSON is still in the log via tee.
- Matches the established pattern already used at line 331 and the `set -euo pipefail` steps at lines 132/179. No behavior change on the green path.

- [P3] `.github/workflows/openapi.yml:394` — the adjacent "Check for missing operationIds" step has the same masked-pipe shape (`audit_openapi_docs.py --json | python -c ...`): a non-zero left-side exit that still emits valid JSON passes silently. It's labeled warning-only so this appears intentional, but if it's ever meant to gate, it needs the same pipefail treatment. Out of this diff's scope; noting so it isn't rediscovered as "drift" later.

dogfood: yes

scarmani:

Head: 35f854a (35f854ab57b7ecf64333b80765d93a66e7257daf), committed 2026-07-17T12:38:26Z.
PR: #9368.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

PR: #9368
Exact head: 35f854ab57b7ecf64333b80765d93a66e7257daf
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #9368 is non-draft and live exact-head checks/merge-packet remain otherwise green.

scarmani:

Head: 35f854a (35f854ab57b7ecf64333b80765d93a66e7257daf), committed 2026-07-17T12:38:26Z.
PR: #9368.
Model family: claude

Verdict: PASS

Verified against the full file at head 35f854a: the step's run block (`.github/workflows/openapi.yml:371-390`) has no explicit `shell:` key and the workflow sets no `defaults.run.shell`, so GitHub runs it as `bash -e {0}` — errexit on, pipefail **off**. The `python … --json | tee /tmp/route-coverage.json` pipeline at line 381 therefore returned tee's exit 0 regardless of the validator's status, exactly as the comment claims. Adding `set -o pipefail` makes the validator's nonzero status the pipeline status, which the implicit `-e` then turns into a step failure — the fix is correct and functional, not a no-op. The only pipe in the block feeds `tee`, which consumes all input, so no SIGPIPE false-failure risk, and errexit now also stops the block before the `COVERAGE=$(…)` extractions can run against a bad/partial JSON file. Because the workflow runs on `pull_request` from the PR's merge ref, this PR's own "Generate & Validate" check exercises the armed gate before merge — an undetected currently-failing validator would surface on the PR, not on main.

- [P3] `.github/workflows/openapi.yml:394` — the "Check for missing operationIds" step has the same masked-pipe shape (`audit_openapi_docs.py --json | python -c …`); it's warning-only and a truncated stream would make the downstream `json.load` fail anyway, but if you ever make it enforcing it will need the same treatment. A workflow-level `defaults: { run: { shell: bash } }` would enable pipefail everywhere (GitHub's explicit-`bash` invocation includes `-o pipefail`), at the cost of auditing every other pipe in the file first — the narrow scoping here is a reasonable choice, just noting the alternative.
- [P3] `.github/workflows/openapi.yml:376-377` — the newly armed gate validates the committed `docs/api/openapi.json` while every neighboring step validates the freshly generated `openapi_generated.json`; pre-existing and intentional per the baseline design, but worth keeping in mind that pipefail now enforces drift against the *checked-in* spec only.

dogfood: yes

scarmani:

Head: 35f854a (35f854ab57b7ecf64333b80765d93a66e7257daf), committed 2026-07-17T12:38:26Z.
PR: #9368.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9374, opened 2026-07-17
feat(routing): price routing decisions from the canonical model catalog (Phase 2 consumer)

scarmani:
## The $0-cost bug

The smart-routing Pareto optimizer (`aragora/routing/cost_quality_optimizer.py`) has been seeing **$0 cost for every frontier model pin** (noted Jul 9), which corrupts every cost/quality/latency tradeoff it computes. Two stale inputs combine to produce it:

1. **The legacy pricing table is stale.** `aragora/routing/provider_config.py::PROVIDER_PRICING` tops out at claude-opus-4 / gpt-4o — it predates every July-2026 frontier model. `get_estimated_cost()` returns `0.0` for any id it does not know, so claude-fable-5, gpt-5.6-sol, grok-4.5, etc. all price at $0.
2. **The optimizer never actually reads the table** (a surprise found while wiring this): it ranks on `ProviderMetrics.avg_cost_per_debate`, i.e. *recorded* spend. Several outcome recorders — notably `aragora/pipeline/unified_orchestrator.py`, which calls `record_outcome(name, consensus_reached=...)` with the default `cost=0.0` — record no spend at all. Any provider recorded that way enters every cost comparison as **free**, dominates the Pareto frontier, wins `COST_OPTIMIZED` selection, and passes every budget filter.

The newly merged decision-stakes router (#9322) records rationale (`models_considered`, `frontier_before_constraints`) that inherits those $0 costs into audit records.

## The fix: a pricing provider with a fallback ladder

New `aragora/routing/pricing.py` resolves any model id / provider pin to real USD-per-MTok prices:

1. **Canonical model catalog** (`aragora.models`, PR #9355) via `by_any_id()` — the drift-enforced source of truth.
2. **Legacy table** (`PROVIDER_PRICING`) for rows the catalog does not (yet) cover.
3. **Conservative nonzero default** with a once-per-model logged warning — frontier families (claude/gpt/grok/gemini/...) default to $5/$25 per MTok, unknown models to $1/$4. **A known-frontier family never resolves to $0.**

Wiring:
- `CostQualityOptimizer` normalizes store reads through `with_effective_costs()`: entries with no observed spend (`avg_cost_per_debate <= 0`) get a catalog-priced per-debate expectation (2K input / 1K output tokens, matching the existing `_details_from_pricing` convention) and are marked `cost_estimated=True`. Observed spend passes through untouched. Selection-from-recorded-candidates consistency (#9322's audit invariant) is preserved: normalization happens only at store-read boundaries, so the rationale's recorded candidate set is exactly the set selected from.
- `DecisionStakesRouter` rationale entries now carry `cost_source: "observed" | "estimated_from_pricing"`, and the receipt payload keeps `cost.recorded: false` — estimates feed the tradeoff, never claimed as spend.
- `aragora.routing` exports `ResolvedPricing`, `resolve_model_pricing`, `estimate_model_cost_usd`. The legacy `get_estimated_cost` is left as-is (documented 0.0-for-unknown behavior, other consumers unchanged).

## Dependency note: PR #9355

The catalog import is **guarded** (`aragora/routing/pricing.py::_catalog_resolver`): this PR is based on `origin/main` and cherry-picks nothing from #9355. It is mergeable whether #9355 lands **before or after** this PR:
- pre-#9355: the ladder skips to the legacy table / conservative defaults (still never $0 for frontier pins);
- post-#9355: catalog prices take precedence automatically, and `tests/routing/test_pricing.py::TestCatalogRung::test_real_catalog_when_available` un-skips and validates against the real `CATALOG`.

## Tests

`tests/routing/test_pricing.py` (22 passing + 1 importorskip-gated):
- known catalog model resolves to nonzero real price (patchable resolver seam; verified against the real catalog by temporarily materializing #9355's `aragora/models` — 85/85 pass with it present);
- unknown model falls back with a logged warning (once per model);
- frontier pins (claude-fable-5, gpt-5.6-sol, grok-4.5, ...) never score $0;
- optimizer no longer picks an unpriced frontier pin as "cheapest" and budget filters now bind on estimated frontier cost;
- decision-stakes rationale carries real costs + `cost_source`.

Full `tests/routing/` suite: 638 passed / 5 failed — the 5 failures are pre-existing on clean `origin/main` (LLM-mock `test_domain_detector.py` tests, unrelated). `tests/debate/test_budget_aware_debate.py` green. mypy clean on `aragora/routing/`. Docs metrics regenerated in-commit (`regenerate_metrics.py` + `doc_stats.py --write`).

## Follow-ups (out of scope)

- `ProviderRouter._details_from_pricing` and `DEFAULT_PROVIDER_ORDER` still enumerate only the legacy table; migrating them to catalog projections is a natural next Phase-2 consumer step.
- Outcome recorders that pass `cost=0.0` (e.g. `unified_orchestrator`) could estimate real spend at record time via `estimate_model_cost_usd`.


scarmani:

Head: b55179f (b55179fc5cbb16e93e0d19693327b7321cac414b), committed 2026-07-17T02:14:22Z.
PR: #9374.
Model family: claude

Verdict: PASS

- [P3] `aragora/routing/pricing.py:70` — the "never $0" design leaves no way to represent a genuinely-free provider. Subscription-backed CLI agents / local proxy backends that legitimately record `cost=0.0` will be re-priced as paid API traffic (frontier default $0.035/debate), which can exclude the actually-free option under tight `budget_remaining`. The zero-means-unrecorded assumption is documented and deliberate, but an explicit "known-free" sentinel/flag would remove the ambiguity.
- [P3] `aragora/routing/pricing.py:70` (`FRONTIER_FAMILY_PREFIXES`) — the list covers only Anthropic/OpenAI/Google/xAI families. Frontier pins from other vendors (e.g. Meta Muse Spark, Mistral Large, DeepSeek) fall to the unknown default ($1/$4 per MTok), ~5x cheaper than the frontier default, so an unpriced non-Western frontier pin still under-prices relative to intent. Worth extending the tuple or noting the limitation.
- [P3] `tests/routing/test_pricing.py:104` (`test_real_catalog_when_available`) — `importorskip` only guards module presence; the test then assumes `models.CATALOG` and `spec.input_per_mtok`/`output_per_mtok` from unmerged PR #9355. If #9355 lands with a different surface (rename, dict shape), this test errors rather than skips, coupling this PR's green-ness to another PR's API.
- [P3] `aragora/routing/pricing.py:141` (`_catalog_resolver`) — a one-time `ImportError` pins `_by_any_id = None` for the process lifetime with only a debug-level log, so a transient import failure (stale .pyc, partial deploy) silently disables the catalog rung until restart while routing continues on legacy/default prices. Consider warn-level on first fallback.

The core ladder math checks out (per-1K↔per-MTok conversions, 2K/1K debate estimate = $0.035 frontier default), `with_effective_costs` copies rather than mutates and never writes estimates back to the store, `from_dict` defaults keep old snapshots compatible, and the audit trail (`cost_estimated` → `cost_source` in rationale, receipt `cost.recorded` stays false) correctly avoids presenting estimates as observed spend.

dogfood: yes

scarmani:

Head: da03fc7 (da03fc7d919e99f599e909d380f590c19caa0217), committed 2026-07-17T08:00:56Z.
PR: #9374.
Model family: claude

Verdict: PASS

- [P3] `aragora/routing/pricing.py:169-177` (`resolve_model_pricing`, catalog rung) — catalog prices are returned verbatim with no `<= 0` guard, so a future zero-priced catalog row (e.g. a `:free` OpenRouter variant) would flow through as $0/$0, violating the module's documented "never returns $0/$0" guarantee and silently resurrecting the free-domination bug for that model. Today's catalog has no zero rows, so this is latent; a one-line fall-through-to-default guard would make the guarantee structural.
- [P3] `aragora/routing/pricing.py:106-118` (`ESTIMATED_*_TOKENS_PER_DEBATE` / `estimate_debate_cost_usd`) — the substituted estimate prices one 2K-in/1K-out turn (~$0.035 at the frontier default), while observed `avg_cost_per_debate` is full-debate spend, typically much higher. An unpriced frontier pin therefore still enters COST_OPTIMIZED/BALANCED comparisons well below providers with real recorded spend; the fix removes the $0 pathology but estimated and observed values are not on the same scale. The `cost_estimated`/`cost_source` markers make this auditable, but worth a follow-up note (e.g. scale by expected turns per debate).

dogfood: yes

PR #9377, opened 2026-07-17
fix(cli): demo --receipt writes one receipt file and the banner names it

scarmani:
## Problem

`aragora demo --offline --receipt r.json` dual-wrote receipts confusingly:

- `_run_demo_debate()` always saved `aragora-demo-receipt.json` and the DECISION RECEIPT banner printed "Full receipt saved to: ./aragora-demo-receipt.json"
- `_run_mock_demo()` (and `run_demo()`) then additionally wrote the user's `--receipt` path silently, producing two receipt files with divergent timestamps for one run
- Result: a user passing `--receipt r.json` saw a banner pointing at a different file

## Fix

Pass the requested receipt path down into `_run_demo_debate()`:

- exactly one receipt file is written per run — the `--receipt` path when given, else the default `aragora-demo-receipt.json`
- the banner names the file actually written
- absolute `--receipt` paths are displayed as-is instead of with a bogus `./` prefix

## Tests

New `TestReceiptSingleWrite` class in `tests/cli/test_demo_command.py` covering: builtin-fallback branch (`HAS_ARAGORA_DEBATE=False`), custom-topic branch, named-demo branch, absolute-path display, default-file behavior with no flag, and `aragora receipt verify` round-trip of the requested file. Verified end-to-end via the real CLI (single file written, banner correct, receipt verifies VALID 3/3).

Pre-existing, unrelated: `tests/cli/test_cli_main.py::TestRunDebate` fails when run after `test_offline_golden_path.py` on a clean tree too (cross-file pollution).

### DIFF
--- a/aragora/cli/demo.py
+++ b/aragora/cli/demo.py
@@ -382,7 +382,8 @@ def _print_receipt_summary(result: DebateResult, elapsed: float, receipt_file: s
         short_hash = artifact_hash[:12] + "..." if len(artifact_hash) > 12 else artifact_hash
         print(f"  Hash:         sha256:{short_hash}")
     print()
-    print(f"  Full receipt saved to: ./{receipt_file}")
+    display_path = receipt_file if Path(receipt_file).is_absolute() else f"./{receipt_file}"
+    print(f"  Full receipt saved to: {display_path}")
     print("=" * 64)
     print()
 
@@ -598,15 +599,24 @@ def _build_builtin_demo_result(topic: str) -> Any:
     )
 
 
-async def _run_demo_debate(topic: str) -> tuple[DebateResult, float]:
-    """Run a demo debate and return (result, elapsed_seconds)."""
+async def _run_demo_debate(
+    topic: str,
+    receipt_path: str | None = None,
+) -> tuple[DebateResult, float]:
+    """Run a demo debate and return (result, elapsed_seconds).
+
+    The receipt is written once, to ``receipt_path`` when provided (e.g. via
+    ``--receipt``), otherwise to ``aragora-demo-receipt.json`` in the CWD. The
+    receipt summary banner names whichever file was written.
+    """
+    receipt_target = receipt_path or "aragora-demo-receipt.json"
     if not HAS_ARAGORA_DEBATE or StyledMockAgent is None or Arena is None or DebateConfig is None:
         agent_names = [name for name, _ in _AGENT_CONFIGS]
         _print_banner(topic, agent_names)
         result = _build_builtin_demo_result(topic)
         elapsed = 0.0
         _print_result(result, elapsed)
-        receipt_file = _save_demo_receipt(result, elapsed, "aragora-demo-receipt.json")
+        receipt_file = _save_demo_receipt(result, elapsed, receipt_target)
         _print_receipt_summary(result, elapsed, receipt_file)
         return cast(Any, result), elapsed
 
@@ -692,8 +702,8 @@ async def _run_demo_debate(topic: str) -> tuple[DebateResult, float]:
 
     _print_result(result, elapsed)
 
-    # Auto-save receipt JSON to CWD
-    receipt_file = _save_demo_receipt(result, elapsed, "aragora-demo-receipt.json")
+    # Save the receipt once: --receipt path if given, else default in CWD
+    receipt_file = _save_demo_receipt(result, elapsed, receipt_target)
     _print_receipt_summary(result, elapsed, receipt_file)
 
     return result, elapsed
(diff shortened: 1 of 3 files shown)


scarmani:

Head: 477de0c (477de0ca0e3efde0a573f57c80ab2b9f2c290d0f), committed 2026-07-17T04:27:12Z.
PR: #9377.
Model family: claude

Verdict: PASS

- [P3] `aragora/cli/demo.py:619,706` (via `_save_demo_receipt` at `demo.py:400-411`) — no parent-directory creation: `--receipt out/r.json` with a missing `out/` dir raises an unhandled `FileNotFoundError` after the entire demo has run. Pre-existing, but the write now happens before the receipt banner in every path, so the failure also swallows the banner where the old dual-write flow at least printed the default receipt summary first. A `path.parent.mkdir(parents=True, exist_ok=True)` (or a clean error) would close it.
- [P3] `tests/cli/test_demo_command.py:535,549` — `test_custom_topic_...` and `test_named_demo_...` don't pin `HAS_ARAGORA_DEBATE`, so which branch of `_run_mock_demo` they exercise depends on whether `aragora-debate` is importable in the test environment; in a fallback env both collapse onto the builtin branch and the `run_demo`/named-demo path (the one whose dual-write `demo.py:919` removed) goes untested. Assertions pass either way, but patch `HAS_ARAGORA_DEBATE` to `True` in the named-demo test to make the regression coverage deterministic.

Verified during review: all three offline branches in `_run_mock_demo` thread `receipt_path` through; the only external caller of `_run_demo_debate` (`aragora/cli/commands/starter.py:216`) passes no receipt path and keeps its prior default-file behavior; the removed `"Receipt saved to:"` print from `run_demo` has no test/doc dependents (the surviving occurrence at `demo.py:804` is the untouched live-demo path); the absolute-vs-relative display logic at `demo.py:385` is sound since `_save_demo_receipt` returns the path verbatim. The empty-string `--receipt` edge falls back to the default filename, identical to the old falsy check — not a regression.

dogfood: yes

scarmani:

Head: 477de0c (477de0ca0e3efde0a573f57c80ab2b9f2c290d0f), committed 2026-07-17T04:27:12Z.
PR: #9377.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

Head: ee76549 (ee76549810c64186b1552446afecaf99092ddc43), committed 2026-07-17T13:14:39Z.
PR: #9377.
Model family: claude

Verdict: PASS

Review basis: read the full changed code plus all callers/consumers (`starter.py`, `run_demo` importers, `cmd_receipt_verify` exit semantics, docs/tests referencing the default receipt filename). Test execution was denied by the session's permission mode, so this is a static review; the fix logic, fixture ordering, and every dependency I traced check out.

- Single-write fix is correct in all three mock paths (builtin fallback, custom topic, named demo via `run_demo`): `receipt_target` threads through `_run_demo_debate`, and no path calls `_save_demo_receipt` twice anymore. The previously silent named-demo dual-write (`run_demo` writing both the default file and the `--receipt` file while the banner named the wrong one) is gone.
- Compatibility verified: `aragora/cli/commands/starter.py:216` calls `_run_demo_debate(topic)` positionally — the new keyword-default keeps it working, and `tests/cli/test_starter_command.py:25` still gets its default file. No test or doc depends on the removed `run_demo` "Receipt saved to:" print (the only remaining prints of that string are the untouched live path `demo.py:804` and unrelated commands). All README/quickstart/onboarding flows pass `--receipt aragora-demo-receipt.json`, so nothing relies on the old side-write of the default file.
- The `_restore_aragora_offline_env` fixture is sound: defined before the other autouse fixtures so its teardown runs last; snapshot/restore is monkeypatch-safe; and its docstring's pytest claim is accurate (`delenv` on an absent var records nothing to undo). The leak source (`cmd_ask` demo tests) is confined to this module — `test_cli_main.py` only parses args, never executes cmd_ask — so the fixture placement covers the actual source.
- [P3] aragora/cli/commands/debate.py:1821 — the root cause of the env leak (`os.environ.setdefault("ARAGORA_OFFLINE", "1")` as a permanent process-wide mutation in cmd_ask demo mode) is left in place; the fixture is a test-side shield, and any future test module that drives cmd_ask demo mode will silently re-leak until it copies the same fixture. Consider a product-side scoped set/restore in a follow-up.
- [P3] tests/cli/test_demo_command.py:536-559 — `test_custom_topic_...` and `test_named_demo_...` run the real `aragora_debate` Arena end-to-end (deterministic offline, but the heaviest way to assert "one file was written"); acceptable as regression coverage, just noting the cost.

dogfood: yes

scarmani:

Head: ee76549 (ee76549810c64186b1552446afecaf99092ddc43), committed 2026-07-17T13:14:39Z.
PR: #9377.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

Issue #9397, opened 2026-07-17
Contract drift CD-098: typescript SDK drift — debates (27 entries)


scarmani:
Wave-1 worker ticket CD-098 of the contract-drift burn-down lane (#9378). Source: `typescript_sdk_drift`, domain: `debates`, 27 open entries.

Each entry is a route the typescript SDK calls that the OpenAPI spec does not document. The route-coverage gate is armed (#9368), so the spec now documents every handler-**served** route — these survivors are therefore likely fictional or mis-shaped SDK calls.

## Recipe (per entry)
1. Probe the server: does any handler actually dispatch this VERB+path? Check `can_handle` AND a real dispatch branch (ROUTES membership alone is not enough — see #9366/#9367/#9371).
2. **Served but undocumented** (rare now): fix the generator coverage (`scripts/generate_openapi.py`), never hand-edit the spec; entry resolves on spec regen.
3. **Not served**: the SDK method is fiction. Deprecate it in place with a `@deprecated` docstring/JSDoc naming the real alternative (pattern: #9366/#9367), and record it on the wire-or-remove list in this issue's comments. NOTE: deprecation alone does NOT resolve the baseline entry — removal (batched, operator-review) or wiring does. Only prune an entry here if the SDK call itself is removed or the route becomes documented.
4. Mis-shaped call (wrong path/verb for a real endpoint): fix the SDK call to the documented contract (breaking-change rules apply — additive parallel + deprecation if the signature must change).

## Entries
- [ ] `DELETE /api/debates/{param}/notes/{param}`
- [ ] `DELETE /api/debates/{param}/permanent`
- [ ] `DELETE /api/debates/{param}/tags`
- [ ] `GET /api/debates/analytics/consensus`
- [ ] `GET /api/debates/analytics/trends`
- [ ] `GET /api/debates/archived`
- [ ] `GET /api/debates/batch/{param}/results`
- [ ] `GET /api/debates/statistics/agents`
- [ ] `GET /api/debates/{param}/agent-statistics`
- [ ] `GET /api/debates/{param}/audience/suggestions`
- [ ] `GET /api/debates/{param}/checkpoints`
- [ ] `GET /api/debates/{param}/health`
- [ ] `GET /api/debates/{param}/quality`
- [ ] `GET /api/debates/{param}/reasoning`
- [ ] `GET /api/debates/{param}/replay`
- [ ] `GET /api/debates/{param}/similar`
[...]

Issue #9395, opened 2026-07-17
Contract drift CD-001: python SDK drift — debates (39 entries)


scarmani:
Wave-1 worker ticket CD-001 of the contract-drift burn-down lane (#9378). Source: `python_sdk_drift`, domain: `debates`, 39 open entries.

Each entry is a route the python SDK calls that the OpenAPI spec does not document. The route-coverage gate is armed (#9368), so the spec now documents every handler-**served** route — these survivors are therefore likely fictional or mis-shaped SDK calls.

## Recipe (per entry)
1. Probe the server: does any handler actually dispatch this VERB+path? Check `can_handle` AND a real dispatch branch (ROUTES membership alone is not enough — see #9366/#9367/#9371).
2. **Served but undocumented** (rare now): fix the generator coverage (`scripts/generate_openapi.py`), never hand-edit the spec; entry resolves on spec regen.
3. **Not served**: the SDK method is fiction. Deprecate it in place with a `@deprecated` docstring/JSDoc naming the real alternative (pattern: #9366/#9367), and record it on the wire-or-remove list in this issue's comments. NOTE: deprecation alone does NOT resolve the baseline entry — removal (batched, operator-review) or wiring does. Only prune an entry here if the SDK call itself is removed or the route becomes documented.
4. Mis-shaped call (wrong path/verb for a real endpoint): fix the SDK call to the documented contract (breaking-change rules apply — additive parallel + deprecation if the signature must change).

## Entries
- [ ] `DELETE /api/debates/{param}/notes/{param}`
- [ ] `GET /api/debates/analytics/consensus`
- [ ] `GET /api/debates/analytics/trends`
- [ ] `GET /api/debates/archived`
- [ ] `GET /api/debates/batch`
- [ ] `GET /api/debates/batch/{param}/results`
- [ ] `GET /api/debates/{param}/agent-statistics`
- [ ] `GET /api/debates/{param}/agents`
- [ ] `GET /api/debates/{param}/audience/suggestions`
- [ ] `GET /api/debates/{param}/checkpoints`
- [ ] `GET /api/debates/{param}/context/{param}`
- [ ] `GET /api/debates/{param}/decision-integrity`
- [ ] `GET /api/debates/{param}/export`
- [ ] `GET /api/debates/{param}/health`
- [ ] `GET /api/debates/{param}/metadata`
- [ ] `GET /api/debates/{param}/notes`
- [ ] `GET /api/debates/{param}/publish/youtube/status`
[...]

Issue #9378, opened 2026-07-17
Contract-drift burn-down lane: work the accepted batch clocks (1,049 open)


scarmani:
Anchor issue for the contract-drift burn-down lane. The #9352 explained-intake amendment made drift intake honest; this lane is the other half of the bargain: actually burning the accepted batches down on their clocks.

## Current state (generated 2026-07-16 from main @ 6d6ccb7e59)

| Source | Open items |
|---|---:|
| `verify_python_sdk_drift` | 599 |
| `verify_typescript_sdk_drift` | 422 |
| `routes_orphaned_in_spec` | 17 |
| `routes_missing_in_spec` | 11 |
| **Total** | **1049** |

Active decay clocks (10%/wk per batch from its own discovery date): the 2026-04-17 start cohort (target already decayed to ~184) plus the 2026-07-16 discovered batches (#9362's 529-entry py-extractor exposure, #9366/#9369 singletons). The program-mode ratchet (weekly `governance` run on main) is red against these targets today; the PR-mode gate only enforces non-worsening, so this lane is how the program line comes back under target.

## The mechanical recipe (per entry)

1. **Is the route actually served?** Probe the handler dispatch (`can_handle` + a real dispatch branch — ROUTES membership alone is NOT enough; see the memory-surface fiction PRs #9366/#9367/#9371 where ~20 ROUTES-listed paths had no dispatch branch).
2. **Served** → the entry resolves by documenting the route in the spec. **Do not hand-edit `docs/api/openapi.json`** — regeneration overwrites it. This class is blocked on #9360 (generator handler-ROUTES supplement); once it merges, regenerate the spec and most served-route entries resolve in bulk.
3. **Not served (fictional SDK method)** → wire-or-remove decision. Default mechanical action: deprecate in place + add the contract-correct parallel (pattern proven in #9366/#9367). NOTE: deprecation alone does NOT resolve the drift entry — the SDK still calls the route. Resolution requires either wiring the route server-side or removing the method (public-surface change: operator-review class, batch these for a coordinated vN+1 removal PR).
4. Baseline/inventory bookkeeping: resolved entries drop out on the next `generate_contract_drift_inventory.py` run; never delete inventory history, and never reopen a resolved entry as fresh intake (#9352 rules).

## Sequencing

[...]

PR #9402, opened 2026-07-17
fix(sdk): CD-098 typescript debates drift wave-1 — classify, deprecate fiction, fix mis-shaped calls (#9397)

scarmani:
Wave-1 burn-down worker for #9397 (CD-098): typescript_sdk_drift, domain `debates`, 27 entries. Part of the burn-down lane #9378.

## Method

Dispatch truth was probed with the real `RouteIndex.build()` over `HANDLER_REGISTRY` (same mechanics as `aragora/server/handler_registry/core.py`), using the exact paths the TS SDK sends. Result: **every one of the 27 paths resolves to `DebatesHandler`** — whose `can_handle` claims the whole `/api/debates/*` subtree and whose `handle()` has no dispatch branch for any of them, so requests fall through to the `_get_debate_by_slug` lookup and return 404 — **except** `POST .../evidence`, which resolves to `ExplainabilityHandler` (GET-default `can_handle`) and silently returns the evidence *explanation* with 200 while storing nothing.

## Per-entry classification (27/27)

| Entry | Classification | Action |
|---|---|---|
| `DELETE /api/debates/{param}/notes/{param}` | fictional (no notes feature) | deprecated `debates.deleteNote` |
| `DELETE /api/debates/{param}/permanent` | fictional | deprecated `debates.deletePermanently` → `delete()` (documented `DELETE /api/v1/debates/{id}`) |
| `DELETE /api/debates/{param}/tags` | fictional | deprecated `debates.removeTags` → `update()` (`PATCH /api/v1/debates/{id}` `tags`) |
| `GET /api/debates/analytics/consensus` | fictional (ROUTES-listed, never dispatched) | deprecated `debates.getConsensusAnalytics` → `GET /api/analytics/consensus-quality` |
| `GET /api/debates/analytics/trends` | fictional (ROUTES-listed, never dispatched) | deprecated `debates.getTopicTrends` → `GET /api/v1/analytics/debates/trends` |
| `GET /api/debates/archived` | fictional (ROUTES-listed, never dispatched) | deprecated `debates.listArchived` |
| `GET /api/debates/batch/{param}/results` | fictional | deprecated `debates.getBatchResults` → `getBatchStatus()` (documented) |
| `GET /api/debates/statistics/agents` | fictional | deprecated `debates.getAgentStatistics` → `getStatsAgents()` (documented `GET /api/v1/debates/stats/agents`; the unversioned form 404s — fixed in review round) |
| `GET /api/debates/{param}/agent-statistics` | fictional | deprecated `debates.getDebateAgentStatistics` → `getStatsAgents()` |
| `GET /api/debates/{param}/audience/suggestions` | fictional | deprecated `audience.getSuggestions` → `listSuggestions()` (documented `GET /api/v1/audience/suggestions?debate_id=`) |
| `GET /api/debates/{param}/checkpoints` | **shadowed** (CheckpointHandler implements it; unreachable via route index) | deprecated `checkpoints.listForDebate` → `list({ debate_id })`; wire-or-remove |
| `GET /api/debates/{param}/health` | fictional | deprecated `debates.getHealth` → `listHealth()` (system-wide) |
| `GET /api/debates/{param}/quality` | fictional | deprecated `debates.analyzeArgumentQuality` → `getSummary()` |
| `GET /api/debates/{param}/reasoning` | **shadowed** (DebateInterventionHandler implements it; not registered in route index) | deprecated `debates.getReasoning`; wire-or-remove |
| `GET /api/debates/{param}/replay` | fictional | deprecated `replays.getFromDebate` → `replays.get()` (documented `GET /api/replays/{id}`) |
| `GET /api/debates/{param}/similar` | fictional | deprecated `debates.findSimilar` → `search()`/`compare()` |
| `GET /api/debates/{param}/spectate/public` | **shadowed** (DebateShareHandler implements the versioned GET; unreachable) | deprecated `debates.getPublicSpectate` → `getPublicDebate()`; wire-or-remove |
| `PATCH /api/debates/{param}/metadata` | fictional | deprecated `debates.updateMetadata` → `update()` (`PATCH /api/v1/debates/{id}` `metadata`) |
| `POST /api/debates/batch/{param}/cancel` | fictional (worse: `/cancel` suffix branch treats the literal segment `batch` as a debate id) | deprecated `debates.cancelBatch` → per-debate `cancel()` |
| `POST /api/debates/batch/{param}/retry` | fictional | deprecated `debates.retryBatch` → `submitBatch()` |
| `POST /api/debates/{param}/audience/suggestions` | fictional | deprecated `audience.submitSuggestion` → `createSuggestion()` (`POST /api/v1/audience/suggestions`) |
| `POST /api/debates/{param}/checkpoint` | **shadowed** (CheckpointHandler implements it; unreachable) | deprecated `checkpoints.createForDebate`; wire-or-remove |
| `POST /api/debates/{param}/evidence` | **mis-shaped** (lands on GET-only evidence-chain endpoint, silent no-op) | added `debates.injectEvidence()` targeting documented `POST /api/v1/debates/{id}/inject-evidence`; deprecated `addEvidence`; target route **wired server-side in the review round** (was itself unreachable) |
| `POST /api/debates/{param}/make-permanent` | fictional | deprecated `debates.makePermanent` |
| `POST /api/debates/{param}/notes` | fictional | deprecated `debates.addNote` (also deprecated sibling `getNotes`, same dead feature) |
| `POST /api/debates/{param}/restore` | fictional (no un-archive endpoint) | deprecated `debates.restore` |
| `POST /api/debates/{param}/tags` | fictional | deprecated `debates.addTags` → `update()` |

Counts: 22 fictional, 4 shadowed (server dispatch exists but unreachable), 1 mis-shaped (fixed additively).

## Server-side findings (for operator / follow-up; `DebateInterventionsHandler` since wired in the review round, see below — the rest not fixed here)

The route index (`handler_registry/core.py` `PREFIX_PATTERNS`) hands **every** `/api/debates/*` and `/api/v1/debates/*` dynamic path to `DebatesHandler` (its `can_handle` claims the whole subtree in the normalized pass). Handlers with genuine dispatch branches on debates paths are therefore unreachable, because they have no `PREFIX_PATTERNS` entry / `ROUTE_PREFIXES`:

- `CheckpointHandler` — `GET .../checkpoints`, `POST .../checkpoint`, `POST .../checkpoint/pause`
- `DebateInterventionHandler` — `GET .../reasoning` (spec documents it as **post**, a generator verb misclassification worth revisiting; the handler serves GET and 405s non-POST intervene)
- `DebateShareHandler` — `GET .../spectate/public` (spec documents **post**, same verb artifact), plus `/share`, `/share/revoke`
- `PublicDebateViewerHandler` looks affected by the same shadowing family (out of scope here)

Wiring any of these (one-line `PREFIX_PATTERNS`/`ROUTE_PREFIXES` additions + spec regen) would make the corresponding entries resolvable; that is an operator decision (wire-or-remove). No `scripts/generate_openapi.py` change is included in this PR to keep conflict risk with the concurrent python-debates worker (#9395) minimal.

## Gates

- `python scripts/verify_sdk_contracts.py --strict` — PASS (baseline regressions py=0 ts=0)
- `python scripts/check_contract_drift_ratchet.py --mode pr --base-ref origin/main --strict` — PASS (+0 across all counters)
- `npx tsc --noEmit` (pinned tsc 6.0.3 via `npm ci`) — PASS
- `npx vitest run src/namespaces/__tests__` — 1087/1087 PASS
- cross-SDK parity: new TS-only `/api/debates/{param}/inject-evidence` baselined in `scripts/baselines/cross_sdk_parity.json` until the python SDK adds its parallel (#9395 owns the python debates domain)

## Bookkeeping

- No SDK method removals; no `verify_sdk_contracts.json` prunes (deprecation alone does not resolve entries — per #9397 recipe, entries resolve on removal or on the route becoming documented)
- No hand-edits to generated spec files

## Reviewed design tradeoffs (review round, claude P1/P2/P3s)

Every reviewer claim was re-verified empirically before acting: a full-registry probe (all 327 handlers instantiated, `RouteIndex.build()` exactly as `_init_handlers` does) resolved each contested path, and the dispatch fall-through semantics in `handler_registry/__init__.py` were traced.

- **P1 (confirmed, chose WIRE over remove).** `POST /api/v1/debates/{id}/inject-evidence` resolved to `DebatesHandler` → slug 404, exactly as claimed: `RouteIndex.build()` stores `ROUTES` as literal exact keys, so the interventions handler's wildcard strings never match, and it had no prefix registration. The wiring stayed small — one `ROUTE_PREFIXES = ["/api/v1/debates/"]` line — because the design is already safe by construction: the route index verifies every prefix candidate with `can_handle()`, and `ExplainabilityHandler` (earlier-registered on the same prefix) rejects intervention suffixes, so dispatch lands on the interventions handler deterministically without stealing any existing route. This also revives the whole intervention family (`/pause`, `/resume`, `/nudge`, `/challenge`, `/intervention-log`), which was equally dead. `tests/server/test_debate_interventions_route_dispatch.py` pins resolution for all six routes, the non-stealing properties (explanation → explainability, plain path → debates), subset registry-order faithfulness, and the P2 characterization. Full-registry probe after the fix: all six resolve to `_debate_interventions_handler`. Removal was the fallback if wiring turned nontrivial; it did not.
- **P2 (confirmed, fixed).** `/api/debates/stats/agents` resolves to `DebatesHandler` (404); only `/api/v1/debates/stats/agents` reaches `DebateStatsHandler` (its `ROUTES` are versioned literals and exact-match keys never fire for the unversioned form). `getStatsAgents` now targets the versioned path; the `getAgentStatistics` deprecation note's cross-reference updated to match.
- **P3a (reviewer's attribution corrected by probe).** For the versioned `POST /api/v1/debates/{id}/evidence` the SDK actually sends, the 200 does NOT come from `DebatesHandler`'s `"/evidence"` suffix route: resolution lands on `ExplainabilityHandler` (its `can_handle` claims the `/evidence` suffix in the first prefix pass), whose base-class `handle_post` stub returns `None`, and dispatch falls through to its GET `handle()` → `_handle_evidence` → evidence chain, 200. `DebatesHandler._get_evidence` serves only the unversioned form. The note now states both mechanisms precisely.
- **P3b (reviewer's premise corrected).** `/reasoning` is NOT unimplemented: `DebateInterventionHandler` (singular, `aragora/server/handlers/debate_intervention.py`, distinct from the plural interventions handler) implements `_get_reasoning_summary` and IS registered (`_debate_intervention_handler`) — but is unreachable (`ROUTES = []`; its `DYNAMIC_ROUTES` regex is never consumed by the route index). The note now names the right module and mechanism, drops the "pending wiring fix (#9397)" phrasing, and marks it wire-or-remove for the operator. Not wired here: unlike P1 it was not the target of a new SDK method in this PR, and its `/intervene` sibling shares the pattern — that is one coherent follow-up decision.
- **P3c (fixed).** The three "route index hands every /api/debates/* path to DebatesHandler first" notes now describe the actual semantics: literal `ROUTES` exact-match ahead of a `can_handle`-verified prefix scan, with `DebatesHandler`'s prefix winning only when no earlier candidate claims the path.
- **Drive-by required by the gate:** touching `interventions.py` pulls it into CI's changed-file mypy gate, so its 5 pre-existing `str | None` arg-type errors are drained in the same commit (single narrowing pattern applied at all call sites).

Review-round gates rerun: verify_sdk_contracts --strict PASS, drift ratchet PR-mode +0 across all counters, parity trio PASS, generate+validate pipeline PASS (0 debates-domain param mismatches), pinned tsc 6.0.3 PASS, route-index/collision/ownership suites 108 passed, debates handler suites 1867 passed, interventions tests 35 passed.

### Round 2 (openai P2 — confirmed, fixed)

The round-1 wiring exposed a real hole the reviewer caught: once reachable, every POST intervention action called `_get_or_create_manager(..., create=True)` with **no debate-existence check** (the docstring claimed "Returns None if the debate does not exist" but never did). Arbitrary valid-looking IDs returned 200, minted global `InterventionManager` state for nonexistent debates, and could poison a future debate reusing the ID.

- **Fix:** new `_debate_exists()` gates manager creation, reusing the sibling handlers' existence resolution rather than inventing one (`create.py` cancel and `costs.py` both resolve via active state then storage): (1) existing intervention manager — managers are only ever created after the gate passes, so presence implies a previously validated debate, and this keeps resume/log working if state and storage move on; (2) active state via `get_state_manager().get_debate()`; (3) `get_storage().get_debate()`. Nonexistent debate → the handler's standard 404 envelope, no state created.
- **GET intervention-log had the same hole** (audited per review): it fabricated an empty 200 log for any ID. Now: unknown debate → 404; real debate with no interventions → empty log 200, unchanged.
- **Tests:** each POST action + GET log on a nonexistent ID → 404 with an explicit assertion that no manager state was created; empty-log 200 for existing debates; active-state-only and manager-only existence paths covered. Existing endpoint tests now run against a storage stub that knows the test debate IDs (`tests/handlers/test_debate_interventions.py`).
- Gates rerun after the fix: interventions + route-dispatch 44 passed, debates handler suites 1867 passed, changed-file mypy clean, ruff clean, verify_sdk_contracts --strict PASS, drift ratchet +0.

### Round 3 (openai P2 — confirmed, fixed)

Path-shape validation on the wiring: `can_handle` matched prefix + suffix only and `_extract_debate_id_from_path` accepted `len(parts) >= 5`, so `/api/v1/debates/victim/anything/pause` dispatched here and paused `victim`. Both layers now enforce the exact shape `/api/debates/{id}/{action}` (post version strip, exactly 5 segments): `can_handle` rejects extra-segment paths so they fall through to DebatesHandler's slug 404, and the extractor rejects them (400) as defense in depth. GET intervention-log shares both gates. Tests pin the malformed exemplar at extractor, endpoint (no manager state minted; a real `victim` debate provably NOT paused — the exact-shape pause still returns first-pause 200 afterward), and route-index resolution levels; unversioned form covered; happy paths unchanged. Gates rerun: 53 interventions/dispatch tests, 1867 debates handler tests, changed-file mypy, ruff, verify_sdk_contracts --strict, ratchet +0 — all green.

### Round 4 (openai P2 — confirmed, fixed)

`_debate_exists()` treated any durable storage hit as intervention-eligible, so POST actions against a COMPLETED stored debate minted a fresh `InterventionManager` (default RUNNING) and succeeded. The gate is now split: **log = existence, actions = liveness**. New `_debate_liveness()` classifies live/finished/missing — active server state first, then stored status against the canonical `ACTIVE_DEBATE_STATUSES` set (reused from `admin/dashboard_metrics`, not re-invented) plus `"paused"` so resume keeps working. The five POST actions go through `_acquire_live_manager()`: finished → the sibling cancel-on-completed convention (create.py precedent: 400 naming the status, chosen over 409 to match the existing handler convention), no manager minted; missing → 404. Storage status is consulted even when a manager already exists, so a stale manager for a since-completed debate refuses new actions; the manager-only fast path survives only when state and storage are both silent (legacy resume). GET intervention-log keeps the existence gate — history of a finished debate stays readable (200). Tests: all five POST actions on a completed stored debate → 400 "finished" with no manager state; GET log on completed → 200; stale-manager-after-completion refusal; `"paused"` stored status still accepts actions; live happy paths unchanged. Gates rerun: 61 interventions/dispatch tests, 1867 debates handler tests, changed-file mypy, ruff, verify_sdk_contracts --strict, ratchet +0 — all green.

### Round 5 (claude P2+P3, openai P2 — all confirmed, fixed)

1. **Authz (claude P2):** GET intervention-log had no permission gate — `handle_post` carried `debates:update` but the GET router only had `@handle_errors`, and the round-1 wiring made the route reachable, leaking intervener user_ids and nudge/challenge/injected-evidence text. `handle()` now carries `@require_permission("debates:read")` per the sibling read pattern (evidence/checkpoints/analysis), stacked under `@handle_errors` exactly like `handle_post`. Real-auth tests (`no_auto_auth` + `ARAGORA_TEST_REAL_AUTH`) pin 403 FORBIDDEN for both routers — verified empirically that `@handle_errors` converts the `PermissionDeniedError` into the 403 envelope, so the tests assert the result rather than the raise.
2. **Fail closed (claude P3):** `_debate_liveness` treated absent status as live. Flipped: absent/unparseable status → finished verdict, 400 "not live (status: unknown)", no manager minted. Tests cover a status-less storage row and an active-state object without a status attribute.
3. **Stored-paused resume (openai P2):** a fresh manager starts RUNNING, so stored-paused debates couldn't resume after manager loss. New public `InterventionManager.restore_paused()` (silent — no log entry, no event, no-op unless RUNNING) applied only when the manager is freshly created AND stored status is "paused"; an existing manager is the in-memory truth and is never overridden (tested: stored "paused" does not flip a live RUNNING manager). Tests: resume on stored-paused with no manager → 200 and subsequent nudge works; pause on stored-paused → 400 state error (already paused), distinct from a liveness refusal.

Gates: 67 interventions/dispatch tests + full intervention surface 138 + debates suites 1910 passed; ruff, changed-file mypy (both touched modules), verify_sdk_contracts --strict, ratchet +0 — all green. This closes every reviewer-named seam on the file: wiring, existence, shape, liveness, authz, resume — each covered by an explicit test.

Refs #9397, #9378


scarmani:

Head: b9cf659 (b9cf659f011e05f32b93b39e4071406092ac1657), committed 2026-07-22T23:06:42Z.
PR: #9402.
Model family: claude

Verdict: PASS

- [P3] `interventions.py` `_get_intervention_log` (~line 413): client-visible contract change — previously returned `200` + empty log for *any* ID, now returns `404` when the debate isn't resolvable via manager/state/storage. Intentional and tested, but any consumer polling the log during a debate's creation window (before state/storage registration) will now see 404 instead of an empty log. Worth confirming no dashboard/poller relied on the old always-200 behavior.
- [P3] `tests/server/test_debate_interventions_route_dispatch.py`: reachability is proven against a hand-built 4-handler `_SUBSET`, not the full `HANDLER_REGISTRY`. The order-preservation assertion helps, but any handler registered *before* `ExplainabilityHandler` that also claims the `/api/v1/debates/` prefix and accepts an intervention suffix could shadow this handler in production without these tests catching it. One integration-level resolution check against the real registry would harden the guard.
- [P3] `interventions.py` `_get_intervention_log`: minor redundancy — it calls `get_intervention_manager(debate_id, create=False)`, then `_debate_exists` re-runs that same `create=False` lookup as its first step.

dogfood: yes

scarmani:

Head: b9cf659 (b9cf659f011e05f32b93b39e4071406092ac1657), committed 2026-07-22T23:06:42Z.
PR: #9402.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

Head: 3c3f670 (3c3f6704f0e2f58457335d18d2cb451d6aee6bdc), committed 2026-07-23T03:34:18Z.
PR: #9402.
Model family: claude

Verdict: PASS

- [P3] `interventions.py` `_acquire_live_manager` — `fresh` is computed with a separate `get_intervention_manager(..., create=False)` call, then a second `create=True` call mints the manager; the two aren't atomic. Under concurrent requests on a stored-`paused` debate whose in-memory manager was lost (restart-recovery path), request B can read `fresh=True`/`status="paused"`, receive the manager that request A already created and `resume()`d to RUNNING, and then call `restore_paused()` — silently flipping it back to PAUSED and undoing A's resume. Narrow window, self-correctable by a later resume; a cleaner fix is to have `get_intervention_manager` return a `(manager, created)` tuple so creation detection is atomic.
- [P3] `interventions.py` `_debate_liveness._verdict` — fail-closed status coupling: a genuinely in-flight debate that is *absent from the in-memory state manager* but persisted with a status outside the hardcoded `_INTERVENTION_LIVE_STATUSES` set (any non-standard/intermediate status string) is classified `finished` → 400 "not live". This is the safe direction and is mitigated by the state-manager-first ordering (running debates live there with status `"running"`), but the intervention liveness set is now silently tied to the admin dashboard's `ACTIVE_DEBATE_STATUSES`; a future status added to the debate lifecycle but not to that set would refuse interventions on live debates.
- [P3] `scripts/baselines/cross_sdk_parity.json` — I could not execute the SDK-parity generator in this review to confirm the bumped `typescript_endpoint_count: 2276` matches generated output; the `+1` and the added `/api/debates/{param}/inject-evidence` entry are internally consistent with the new `injectEvidence()` method, and `sdk-parity` is a required CI check, so this is a verification note, not a defect.

The core changes are sound: the reachability fix genuinely wires the previously-dead intervention routes (all six resolve to `DebateInterventionsHandler`), the existence/liveness gate correctly closes the prior over-permissive `create=True` hole (arbitrary IDs could mint global intervention state), the GET log is now RBAC-gated (`debates:read`), `restore_paused()` is additive and lock-guarded, the exact-5-segment path enforcement blocks the `/{id}/anything/pause` acting-on-`{id}` vector, and the SDK `getStatsAgents` versioned-path change fixes a previously-404ing call. Tests cover malformed paths, completed-debate refusal, stale-manager override, and stored-paused recovery.

dogfood: yes

scarmani:

Head: 3c3f670 (3c3f6704f0e2f58457335d18d2cb451d6aee6bdc), committed 2026-07-23T03:34:18Z.
PR: #9402.
Model family: openai

Verdict: PASS
- [P3] `aragora/server/handlers/debates/interventions.py:170` POST requests to `/api/v1/debates/{id}/intervention-log` resolve to this handler, `handle_post()` returns `None`, and dispatch falls through to `handle()`, returning the log over POST. This does not bypass `debates:read`, but it widens the method contract unintentionally.

dogfood: yes

scarmani:

Head: 626cd7b (626cd7b2378f3b0767c6a1805f5613a2294d2b93), committed 2026-07-23T04:04:43Z.
PR: #9402.
Model family: claude

Verdict: PASS

- [P3] `interventions.py` `_acquire_live_manager`/`restore_paused` — the stored-`"paused"` restart-recovery branch (`fresh and status == "paused"`) only fires when state/storage reports status `"paused"`, but the pause endpoint (`_pause_debate`) never persists `"paused"` to state or storage (debates are written to durable storage on completion, not mid-pause). So the round-5 recovery is effectively unreachable in production and is exercised only via the dict-returning `_FakeStorage` in tests. It's harmless (a fresh RUNNING manager whose storage still says `"running"` correctly 400s on resume), but the recovery guarantee is more theoretical than the test suggests — consider having pause persist status, or downgrade the claim.
- [P3] `interventions.py` GET `handle` (~line 123) — the newly added `@require_permission("debates:read")` gate is not exercised by any test; the unit tests call `_get_intervention_log` directly, bypassing the decorator. A context-extraction regression on the GET auth path wouldn't be caught. Risk is low since it mirrors the already-working `handle_post` decorator, but a route-level test would close the gap.
- [P3] `interventions.py` `_get_intervention_log` — the log now returns 404 for an unknown debate ID where it previously returned `200` with an empty log. This is intentional and matches the POST existence gate, but it's a client-visible contract change; ensure SDK/consumers that treated 200-empty as "no interventions" tolerate the 404.

dogfood: yes

scarmani:

Head: 626cd7b (626cd7b2378f3b0767c6a1805f5613a2294d2b93), committed 2026-07-23T04:04:43Z.
PR: #9402.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9011, opened 2026-07-08
fix(review-queue): block parked exact heads

an0mium:
## Summary

Adds a shared exact-head park-record reader for repo-visible PR comments and wires it into both merge-packet and settle_one_pr.py so a PR with a standing current-head park cannot be emitted in admin_squash_order or produce a merge suggestion.

Recognized park records include:
- Current-head repeat-blocker park
- Current-head evidence blocker
- Evidence safety correction

A later exact-head supportive PASS does not clear the park. Only a later explicit operator lift/override for the same head clears it. REST fallback now treats issue-comment lookup as safety-critical and fails closed if comments are unavailable.

## Validation

- PYTHONDONTWRITEBYTECODE=1 python3 -m pytest tests/cli/commands/test_review_queue_park_records.py tests/scripts/test_settle_one_pr.py tests/cli/commands/test_review_queue.py -q -p no:cacheprovider
- python3 -m ruff check aragora/cli/commands/review_queue.py aragora/cli/commands/review_queue_rest_fallback.py aragora/cli/commands/review_queue_park_records.py scripts/settle_one_pr.py tests/cli/commands/test_review_queue.py tests/cli/commands/test_review_queue_park_records.py tests/scripts/test_settle_one_pr.py
- python3 -m ruff format --check aragora/cli/commands/review_queue.py aragora/cli/commands/review_queue_rest_fallback.py aragora/cli/commands/review_queue_park_records.py scripts/settle_one_pr.py tests/cli/commands/test_review_queue.py tests/cli/commands/test_review_queue_park_records.py tests/scripts/test_settle_one_pr.py
- git diff --check
- bash scripts/automation_pr_preflight.sh origin/main HEAD
- Live parser probe against #9005 comments at f80c63e526509fd4b8fb1ec2d2bd69b6faa9835e returns blocked=true for Evidence safety correction.

## Live #9005 note

#9005 is now closed unmerged, so the original parked-open merge-suggestion state can no longer be reproduced live. The new fixture tests cover the parked-open case modeled on #9005.

Draft because this changes merge-readiness helper behavior. Do not merge without normal queue authorization.

Issue #8353, opened 2026-06-13
Batched P3 follow-ups from funnel-substrate reviews (#8338/#8339/#8340)


an0mium:
Batched non-blocking P3 observations from the adversarial reviews on tonight's funnel-substrate PRs (#8338, #8339, #8340). None blocked merge; all are quality follow-ups. One small PR can address several.

## agent_bridge.py (from #8338 review)
- **Delivery cross-check**: first-iteration "no marker, no placeholder" short-circuits to `delivered=True`, so a launcher that died *before pasting* is attested delivered. Cross-check pane non-emptiness or a "Welcome to <agent>" banner before attesting delivery (extends #8317's scope from "staged paste" to "silent launcher death").
- **Placeholder regex anchoring**: `[Pasted Content N chars]` match is permissive; a confirmation dialog or scrollback echo of a *previous* paste in the window could trigger the Enter nudge. Anchor to a leading `> `/composer prefix.
- **Dead local**: `staged` is written in the poll loop but never read after exit (only `placeholder`/`saw_marker_only`/`capture_failed` drive the decision). Trim.
- **SIGPIPE comment**: clarify the inline comment — "scoped because launch is the entrypoint that calls it," not "restored on return."
- **Re-poll interval**: `_verify_prompt_submission` reuses caller `timeout_seconds` for re-poll sleeps (capped by `SUBMIT_VERIFY_NUDGE_REPOLL_INTERVAL`); document or give it a dedicated interval.
- **Test-fake hygiene**: `_FakeTmuxRunner` returns the last pane indefinitely on overrun; a stricter fake would fail loudly.

## pr_state_cache.py (from #8339 review)
- **`verify` should return `labels`** (or the reader contract should document that label routing must consult merge-packet). A lane keying on `boss-ready`/automation labels can act on stale labels even after the just-before-mutation `verify`.
- **Breaker string-normalization**: identical-error breaker keys on the full error string; GitHub error bodies carry request IDs/timestamps, so systemic 502s may never coalesce. Normalize on the HTTP-status prefix (`gh: HTTP <NNN>`). (Same lesson as #8316.)
- **Endpoint-dict pruning**: `{**prev, **new}` merge leaves dropped-tail page entries forever when pagination shrinks N→M. Eventual bloat, never correctness.

## identify_lane_owner.py (from #8340 review)
[...]

PR #9415, opened 2026-07-17
fix(automation): keep newer queued checks visible (#8353)

scarmani:
## Summary

- select the latest shared-cache check attempt by GitHub run id, with creation-time fallback
- prevent an older completed success from hiding a newer queued attempt with null start/completion timestamps
- preserve the existing `.aragora/pr-state-cache.json` reader schema

## Why

`pr_state_cache.py` previously ordered attempts by `completed_at` or `started_at`. A newly queued rerun has neither field, so an older success could remain visible as the latest state. That makes automation state look green while a newer attempt is pending.

This PR intentionally does not touch the Tier-4 settlement REST fallback because that file overlaps open PR #9011.

## Validation

- `pytest -q tests/scripts/test_pr_state_cache.py` (`39 passed`)
- `ruff check scripts/pr_state_cache.py tests/scripts/test_pr_state_cache.py`
- `ruff format --check scripts/pr_state_cache.py tests/scripts/test_pr_state_cache.py`
- `pre-commit run --files scripts/pr_state_cache.py tests/scripts/test_pr_state_cache.py`
- `git diff --check`

Refs #8353

### DIFF
--- a/scripts/pr_state_cache.py
+++ b/scripts/pr_state_cache.py
@@ -43,9 +43,9 @@ Delta detail pass:
 
 - Only PRs whose head moved since the cached entry (or with no cached checks)
   cost a ``repos/{repo}/commits/{sha}/check-runs?per_page=100`` call; the
-  latest run per check name (by completed/started timestamp) is stored as
-  ``{name: conclusion-or-status}``. At most ``--max-detail`` fetches per run;
-  over-cap PRs are annotated ``detail_deferred:<n>``.
+  latest run per check name (by run id, with creation-time fallback) is stored
+  as ``{name: conclusion-or-status}``. At most ``--max-detail`` fetches per
+  run; over-cap PRs are annotated ``detail_deferred:<n>``.
 
 Safety model (mirrors scripts/auto_evidence_cycle.py):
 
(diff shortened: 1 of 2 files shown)


an0mium:
## Exact-head model-evidence transport blocker

Head: `a0a6e31df0d64d101879298d475e2f4343d72ce3`

- PR is non-draft, `MERGEABLE/BLOCKED`, with all five non-quorum required checks green.
- OpenAI dry-run: `PASS`, focused dogfood present, `would_count=true`, no findings.
- Claude dry-run: unavailable because the subscription CLI reports exhausted usage credits and no OpenRouter fallback is configured.
- Grok alternate-family dry-run: returned no parsed verdict, so `would_count=false`.
- No partial evidence was posted and no CI was rerun.

Smallest unblock: restore one second countable reviewer family (Claude credits or a configured independent fallback), then run one fresh exact-head prepared dry-run before any apply.

scarmani:

Head: a0a6e31 (a0a6e31df0d64d101879298d475e2f4343d72ce3), committed 2026-07-17T22:59:46Z.
PR: #9415.
Model family: claude

Verdict: PASS

Review basis: I traced every branch of the new `extract_checks`/`_check_run_is_newer` logic against the existing fixture, the new fixture, and realistic GitHub payloads by hand (a sandbox permission denial prevented actually executing pytest, so this is a manual trace, not a test run — the function is pure and small enough for that to be conclusive). The core fix is sound: real check-run payloads always carry `id`, ids are creation-ordered, so a newer queued attempt with null `started_at`/`completed_at` now correctly wins over a stale completed run — the exact bug fixed. The unchanged test `test_extract_checks_keeps_latest_run_per_name` (id-less fixture) still passes under the new fallback since only `completed_at` is present in its runs. Tie-break via `candidate_index > current_index` preserves the old later-entry-wins semantics for exact duplicates. No security surface touched.

- [P3] `scripts/pr_state_cache.py:281` (`timestamp` helper) and the new test fixture both use `created_at`, but GitHub's check-runs REST object has no `created_at` field — only `started_at`/`completed_at`. Harmless in production (the id path always decides first), but the fallback's primary key and the fixture model a field the API never sends; the docstring's "fallback for incomplete fixtures or payloads" framing is doing real load-bearing work here.
- [P3] In the id-less fallback path, the fixed bug persists: a newer queued attempt with all-null timestamps serializes to `""`, which loses lexically to any older completed run's timestamp, so the queued attempt is still hidden when ids are absent. Only reachable with synthetic/incomplete data, but worth knowing the fix is id-dependent.
- [P3] Fallback ordering silently changed from completion-order (`completed_at or started_at`) to creation-order (`created_at or started_at or completed_at`) for id-less runs — e.g. run A (started 09:00, completed 12:00) vs run B (started 11:00, completed 11:30) now picks B where it previously picked A. Consistent with the fix's attempt-creation-order rationale, but a behavior change for any other fixture-driven consumers of `extract_checks`.

dogfood: yes

scarmani:

Head: a0a6e31 (a0a6e31df0d64d101879298d475e2f4343d72ce3), committed 2026-07-17T22:59:46Z.
PR: #9415.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9477, opened 2026-07-21
feat(transport): route exact OpenAI requests via VibeProxy

scarmani:
## Summary

- add exact-model OpenAI Chat Completions and Responses request support to the central VibeProxy transport policy
- route eligible non-streaming `OpenAIAPIAgent` Chat calls through that policy with one catalog/request deadline
- keep web search/tools, streaming, custom OpenAI gateways, unsupported capabilities, and protected surfaces direct
- preserve prefer-before-output fallback and required-mode fail-closed behavior
- update the no-inference diagnostic, operator guide, and static inference inventory

## Safety boundaries

- VibeProxy remains a transport, never a reviewer family
- no runtime/application inference requests were sent
- no auth or endpoint-pinning, CI/workflow, governance, evidence/settlement, public API/SDK, or other-provider changes
- port 8317 remains prohibited

## Validation

- `pytest -q tests/agents/api_agents/test_openai.py tests/agents/transports/test_vibeproxy.py tests/scripts/test_check_vibeproxy.py tests/scripts/test_check_inference_site_allowlist.py` (135 passed)
- `bash scripts/automation_pr_preflight.sh origin/main HEAD`
- `bash scripts/preflight_mypy.sh --diff-base origin/main`
- `python3 scripts/check_inference_site_allowlist.py --json`
- `python3 scripts/check_charter_compliance.py --base origin/main`
- ruff check + format check; `git diff --check`

Part of #9409.

## Reviewed design tradeoffs (rounds 1-7)

Seven independent model-review rounds ran against this PR. Every P0/P1-free
round produced genuine P2/P3 findings; all were fixed except two areas where
successive rounds issued mutually exclusive demands. The following decisions
are deliberate and settled; future reviews should ground on them:

1. **Request-leg timeout semantics.** The proxy request leg is bounded by
   `remaining_timeout()` evaluated at executor-thread start (raises instantly
   if queue wait exhausted the deadline; otherwise caps the socket wait). It is
   deliberately NOT wrapped in `asyncio.wait_for`: the executor job cannot be
   cancelled, so wall-clock abandonment can run a duplicate inference (proxy
   completes after the direct fallback starts). One review round demanded the
   `wait_for` (added), the next flagged the duplicate-inference hazard
   (removed). You cannot have both with a thread-based client; the
   no-duplicate-spend property wins. Discovery IS `wait_for`-bounded — it is
   idempotent and cheap, so an orphaned discovery call is harmless.

[...]

PR #9493, opened 2026-07-22
fix(tests): reap VibeProxy executor threads to stop cross-file pollution

scarmani:
## Summary
Fixes pre-existing test pollution on main: running `tests/agents/api_agents/test_openai.py` before `tests/swarm/test_quorum_evidence.py` in one process failed `test_collect_overall_timeout_does_not_wait_for_stuck_reviewer` and `test_collect_overall_timeout_fails_closed_and_ignores_late_results`, while each file passed alone.

## Root cause
The VibeProxy routing tests (#9477) submit work to the module-level `_PROXY_EXECUTOR` `ThreadPoolExecutor` in `aragora/agents/api_agents/openai.py`. Its worker threads (`vibeproxy-openai_*`) persist for the rest of the process. `quorum_evidence._reviewer_process_context()` then sees `threading.active_count() > 1` and — correctly, for fork safety in a threaded parent — selects a **forkserver** context instead of **fork**. Forkserver must pickle the worker target, and the quorum tests' locally defined `runner` closures are not picklable:

```
AttributeError: Can't pickle local object 'test_collect_overall_timeout_...<locals>.runner'
```

The production fork-safety check is legitimate; the pollution is the leaked test threads.

## Fix
Autouse fixture in `tests/agents/api_agents/conftest.py` (alongside the existing pollution-guard fixtures there) that monkeypatches a fresh per-test executor and shuts it down with `cancel_futures=True` at teardown, reaping its threads and restoring the pristine (thread-less) module executor.

## Verification
- `pytest tests/agents/api_agents/test_openai.py tests/swarm/test_quorum_evidence.py`: 310 passed (previously 2 failed)
- Thread probe after `test_openai.py`: `threading.active_count() == 1` (previously 3 with two `vibeproxy-openai_*` threads)
- Full `tests/agents/api_agents/`: 811 passed

## Reviewed design tradeoffs
- Test-side fixture over production change: the forkserver selection in `_reviewer_process_context()` is deliberate fork-safety behavior and must not be weakened; the leak is purely a test-process artifact.
- Per-test (function-scoped) swap over module-scoped teardown: threads spawn lazily, so a fresh executor is free, and each test starts from a clean thread state regardless of ordering within the directory.
- `cancel_futures=True` on shutdown so a test that leaves queued work behind cannot hang teardown.

### DIFF
--- a/tests/agents/api_agents/conftest.py
+++ b/tests/agents/api_agents/conftest.py
@@ -557,27 +557,6 @@ def _restore_create_client_session():
 
 
 @pytest.fixture(autouse=True)
-def _fresh_vibeproxy_executor(monkeypatch):
-    """Give each test a private VibeProxy executor and reap its threads.
-
-    The module-level ``_PROXY_EXECUTOR`` in ``api_agents.openai`` keeps its
-    worker threads alive for the rest of the process once a test routes a
-    request through VibeProxy.  Lingering threads flip
-    ``threading.active_count() > 1`` checks elsewhere in the suite — notably
-    ``quorum_evidence._reviewer_process_context()``, which then selects a
-    forkserver context and fails to pickle locally defined reviewer runners.
-    """
-    from concurrent.futures import ThreadPoolExecutor
-
-    from aragora.agents.api_agents import openai as openai_module
-
-    executor = ThreadPoolExecutor(max_workers=8, thread_name_prefix="vibeproxy-openai-test")
-    monkeypatch.setattr(openai_module, "_PROXY_EXECUTOR", executor)
-    yield
-    executor.shutdown(wait=True, cancel_futures=True)
-
-
-@pytest.fixture(autouse=True)
 def _allow_localhost_for_api_agents(monkeypatch):
     """Ensure localhost is allowed for API agent tests.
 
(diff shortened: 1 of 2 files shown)


scarmani:

Head: 6af82e7 (6af82e77606210d19a4edf81f676d0f5faf41c12), committed 2026-07-22T16:03:29Z.
PR: #9493.
Model family: claude

Verdict: PASS

- [P3] `tests/agents/api_agents/conftest.py:585-589` — the docstring's "fails loudly rather than hanging the suite" only half holds. I reproduced it: a worker that outlives the grace period is reported by `pytest.fail`, but `concurrent.futures.thread._python_exit` then joins it unconditionally at interpreter exit, so the pytest process still blocks for the full wedge duration (sim: 2s bounded teardown, 30s total process wall clock). Worth softening the claim, or noting that the real protection is the "block on an event, not a bare sleep" rule.
- [P3] `tests/agents/api_agents/conftest.py:604,606` — `executor._threads` is a private CPython attribute used unguarded, in an autouse fixture covering the whole directory. If it is ever renamed/absent, every test here errors in teardown rather than degrading. `getattr(executor, "_threads", ())` costs nothing.
- [P3] `tests/agents/api_agents/conftest.py:597` — `max_workers=8` duplicates the production pool size instead of deriving from it. `test_prefer_wedged_discovery_bounded_by_wall_clock` exists specifically to pin the "executor queue wait consumes the deadline" invariant documented at `openai.py:296-301`; if production later shrinks the pool, this test silently keeps exercising 8 workers. Prefer a shared constant or `openai_module._PROXY_EXECUTOR._max_workers`.
- [P3] `tests/agents/api_agents/conftest.py:571-575` vs `tests/agents/api_agents/test_openai.py:814-826` — the comment asserts the abandoned leg finishes "while sibling autouse patches (… create_client_session) are still in place", but the wedge is released in the `finally`, i.e. after the test-body `with patch(create_client_session)` has already exited. Harmless today (the resumed leg only touches the fake `WedgedClient`), but the stated invariant is stronger than what the code guarantees. The rationale also leans on pytest's undocumented same-scope autouse definition ordering — fine in practice, fragile if a fixture is ever reordered.

dogfood: yes

scarmani:

Head: 6af82e7 (6af82e77606210d19a4edf81f676d0f5faf41c12), committed 2026-07-22T16:03:29Z.
PR: #9493.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9496, opened 2026-07-22
fix(profiles): keep Claude pool alive via VibeProxy token sync

scarmani:
## Problem

The Claude Max profile pool ran at **0/12 healthy** — every profile expired,
including ones logged in hours earlier. The hourly `claude_pool_verify.py`
refresh side-effect (probe an expired token → CLI refreshes it) is correct in
isolation but was silently defeated.

## Root cause

Anthropic OAuth **single-use-rotates** the refresh token on every refresh: once
used, the old refresh token is revoked. The same accounts were being refreshed
by **three independent owners**:

1. VibeProxy's `cli-proxy-api` (batch-refreshes all its accounts on its own schedule),
2. aragora's hourly `claude_pool_verify.py` probe,
3. for two shared logins (max-08/09, max-12/13), a *second* aragora profile.

Whoever refreshed an account first got the new token; every other holder was
left with a revoked one → dead. Confirmed by account overlap: **9 of 12
profiles hold accounts VibeProxy also refreshes**, and two logins are shared
across profile pairs. `claude_pool_verify.py`'s own docstring predicted this
("canNOT revive a revoked refresh token — duplicate accounts, same-org seats,
or the same account used concurrently").

## Fix — consume VibeProxy's tokens, don't compete

VibeProxy stays alive because it is the **single owner** of each account's
refresh token. So aragora stops refreshing and instead copies VibeProxy's
already-fresh **access** token into the matching profile, written **without a
usable refresh token** (pure consumer). Consequences:

- aragora can never rotate VibeProxy's live token.
- the hourly `claude_pool_verify.py` becomes automatically safe — a profile with
  no refresh token cannot rotate anything even when probed.
- self-healing: if a sync cycle lags past expiry, the profile just reports
  expired until the next cycle; nothing is revoked.

## Contents

- `scripts/sync_claude_profiles_from_vibeproxy.py` — format-translating sync;
  pure-consumer default, idempotent, atomic `0600` writes, `--probe-after`.
- `scripts/install_claude_profile_sync_launchd.sh` — 30-min `launchd` timer.
  Deploys the script to `~/.aragora/bin` (outside any git checkout) so worktree
  TTL-cleanup and future merges never disturb the running daemon.
- `docs/guides/CLAUDE_MAX_POOL.md` — root cause, fix, runbook.
- 9 focused tests.

## Validation

- **0/12 → 11/12 healthy in one sync**, confirmed by `claude_pool_verify.py`
  live probe. The single holdout (`max-10` / `armand.tuzel@gmail.com`) is the
  one account VibeProxy does not hold — add it to VibeProxy or keep it on
  interactive re-login.
- Timer installed and its `RunAtLoad` fire probed green.
- 9 tests, ruff, mypy, automation preflight, diff-check clean.

## Notes / scope

- Local operator tooling; no runtime/product surface changes.
- Requires VibeProxy running with the Claude accounts connected.
- Follow-up worth considering: fold the two shared-login profiles into distinct
  accounts (the health checker already recommends this) so the pool has 13
  distinct-account slots rather than 11 + 2 duplicates.

## Reviewed design decisions (repair rounds)

Several review rounds refined the credential-safety model. These are settled;
future reviews should ground on them:

1. **Native-login guard, then idempotency.** A profile holding a real (non-blank)
   refresh token is a *native login* and is never blanked without `--force`,
   checked BEFORE the access-token idempotency check (so a coincidental token
   match cannot leave a live refresh token racing VibeProxy). Our own synced
   creds carry a blank refresh token and flow freely.

2. **Offline ambiguity → one-time `--force` bootstrap.** A live refresh token
   cannot be distinguished from a revoked one without a network call. So the
   first run over native/dead profiles requires `--force` to convert them into
   blank-refresh consumers; the daemon then runs without `--force` in steady
   state. An all-native result is NOT "healthy" — `main()` exits non-zero and
   points at the bootstrap, rather than silently no-op'ing.

3. **Interval ≤ refresh lead.** The timer runs every 5 min (≤ VibeProxy's
   ~10-min refresh lead) so a sync always lands between the new token appearing
   and the old one lapsing.

4. **Backups are native-only.** A `0600` `.bak` is written only when replacing a
   real native credential (under `--force`), never overwritten by one of our own
   blank-refresh creds.

### DIFF
--- a/docs/guides/CLAUDE_MAX_POOL.md
+++ b/docs/guides/CLAUDE_MAX_POOL.md
@@ -164,10 +164,13 @@ bash scripts/install_claude_profile_sync_launchd.sh
 This revives every profile VibeProxy can source — one profile per VibeProxy
 account. Requirements: VibeProxy running with the Claude accounts connected
 (auth files at `~/.cli-proxy-api/claude-*.json`) and the local config filled in.
-The sync writes a `0600` `.bak` before replacing a credential and refuses to
-overwrite a fresh **native** login (live token + real refresh token) without
-`--force`. Run the timer more often than VibeProxy's refresh lead (it refreshes
-~10 min before the ~8h access-token expiry); 30 min is the default.
+The sync backs up a **native** credential (`0600` `.bak`) before replacing it
+and refuses to overwrite a profile holding a real refresh token without
+`--force`. The timer interval must be **≤ VibeProxy's refresh lead** (it
+refreshes ~10 min before the ~8h access-token expiry) so a sync always lands in
+the window between the new token appearing and the old one lapsing; the default
+is **5 min**. A wider interval would leave a profile on an expired, blank-refresh
+token for up to `(interval − lead)` every ~8h.
 
 **One profile per VibeProxy account (important).** VibeProxy authenticates by
 **email** and holds exactly **one org per email**. Two profiles that are
--- a/docs/guides/CLAUDE_MAX_POOL.md
+++ b/docs/guides/CLAUDE_MAX_POOL.md
@@ -156,7 +156,7 @@ python3 scripts/sync_claude_profiles_from_vibeproxy.py
 
 # ONE-TIME bootstrap: profiles that still hold a native (or dead) refresh token
 # are protected by default, so convert them to VibeProxy-managed consumers once
-# with --force. After this they carry a blank refresh token and sync freely.
+# with --force. Run from the repo so --probe-after can find claude_profile.sh.
 python3 scripts/sync_claude_profiles_from_vibeproxy.py --apply --force --probe-after
 
 # Steady state (no --force): keeps the managed profiles fresh.
--- a/docs/guides/CLAUDE_MAX_POOL.md
+++ b/docs/guides/CLAUDE_MAX_POOL.md
@@ -151,15 +151,28 @@ chmod 600 ~/.aragora/claude_profile_sync.json
 ```
 
 ```bash
-# One-shot: sync every mapped profile from VibeProxy (dry-run first, then apply)
+# Preview the plan (dry-run):
 python3 scripts/sync_claude_profiles_from_vibeproxy.py
-python3 scripts/sync_claude_profiles_from_vibeproxy.py --apply --probe-after
+
+# ONE-TIME bootstrap: profiles that still hold a native (or dead) refresh token
+# are protected by default, so convert them to VibeProxy-managed consumers once
+# with --force. After this they carry a blank refresh token and sync freely.
+python3 scripts/sync_claude_profiles_from_vibeproxy.py --apply --force --probe-after
+
+# Steady state (no --force): keeps the managed profiles fresh.
+python3 scripts/sync_claude_profiles_from_vibeproxy.py --apply
 
 # Make it durable: a launchd timer (deploys the script to ~/.aragora/bin, outside
 # any git checkout so worktree TTL-cleanup and merges never touch it; seeds the
 # config from the example if absent)
 bash scripts/install_claude_profile_sync_launchd.sh
 ```
+
+> **First run needs `--force`.** The daemon runs `--apply` without `--force` and
+> refuses to overwrite any non-blank refresh token (it cannot tell a live native
+> login from a revoked one offline). Until you bootstrap once with `--force`, the
+> sync skips every native/dead profile and **exits non-zero** so the launchd log
+> flags it rather than silently reporting success.
 
 This revives every profile VibeProxy can source — one profile per VibeProxy
 account. Requirements: VibeProxy running with the Claude accounts connected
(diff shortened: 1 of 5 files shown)


scarmani:

Head: 86e2ab6 (86e2ab6d87afc2ad23569fb2f83b65bc382e6966), committed 2026-07-22T20:01:12Z.
PR: #9496.
Model family: claude

Verdict: PASS

Reviewed the sync script's guard ordering, write path, exit-code contract, the launchd installer, and the tests. The core safety properties hold: the native-login guard runs before the idempotency check, `--force` backs up only real native creds, the stale-source margin blocks a crashed proxy from installing a dead token, writes are 0600 + atomic via per-PID tmp + `os.replace`, and no token material is printed. No blocking defects found.

- [P3] `scripts/sync_claude_profiles_from_vibeproxy.py:~410` (`main`, `--json` branch) — `[r.__dict__ for r in results]` emits raw emails, bypassing `_redact_email`, which the human path applies with the justification "emails are PII here". Inconsistent redaction between the two output modes; `SyncResult.detail` for `skipped_native_only` also echoes operator-authored config text verbatim.
- [P3] `scripts/sync_claude_profiles_from_vibeproxy.py:~290` + `scripts/install_claude_profile_sync_launchd.sh:~95` — the comment promises a corrupt cred can "self-heal instead of erroring forever" under `--force`, but the installed plist runs `--apply` only, so that path is unreachable in the deployed daemon. Compounding: the tmp file is not `fsync`ed before `os.replace`, so a crash/power-loss can produce exactly the partial-file state that then wedges the profile until a human runs `--force`.
- [P3] `scripts/sync_claude_profiles_from_vibeproxy.py:~300` — the `.bak.corrupt` copy uses `read_text(errors="replace")`, so "Preserve the corrupt bytes … so nothing is lost" is false for any non-UTF-8 content. Also, if the original read failed with `OSError` (perm denied, not decode), this second read re-raises and the whole profile falls to `error` with a misleading message.
- [P3] `scripts/sync_claude_profiles_from_vibeproxy.py:translate_credential` — `oauth.setdefault("scopes"/"subscriptionType")` preserves whatever the *previous* credential held. The guide advertises repointing `sync_target` at a different VibeProxy account with "no native login needed", which leaves the old account's plan/scope metadata attached to the new account's token.
- [P3] `scripts/sync_claude_profiles_from_vibeproxy.py` `--keep-refresh` — it is effectively one-shot: the next run sees a non-blank `refreshToken`, returns `skipped_native_login`, and `main` exits 1 on every subsequent cycle until someone passes `--force`. Worth stating in the flag help alongside the existing UNSAFE note.
- [P3] `scripts/install_claude_profile_sync_launchd.sh:~75-100` — install-time values (`$HOME`-derived paths, `--python`, `--path`, `CLAUDE_PROFILE_ROOT`) are interpolated into a shell string nested inside XML with no escaping; a path containing `"`, `&`, `<` yields an invalid plist or a broken command. `--interval-seconds` is likewise unvalidated, so a non-numeric value emits `<integer>abc</integer>`.
- [P3] `scripts/install_claude_profile_sync_launchd.sh:~60` — it seeds the *example* config and then immediately `launchctl load`s with `RunAtLoad` + 300s interval, so a fresh install runs against `you+work@example.com` every 5 min and exits 1 each cycle. Safe (no source file exists → `skipped_no_source`, no write), but the 0600 log grows unbounded with no rotation. Consider skipping the load, or refusing to install, when the config is still the unedited example.
- [P3] `docs/guides/CLAUDE_MAX_POOL.md:~150` — "Run from the repo so `--probe-after` can find `claude_profile.sh`" is inaccurate: `_resolve_profile_tool` resolves relative to `__file__` (then `shutil.which`), not cwd. What matters is invoking the *repo copy* of the script, which is what the installer's echoed command actually does. Separately, the guide says "0/12 healthy … 9 of 12" while the script docstring says "0/13 healthy".
- [P3] `scripts/sync_claude_profiles_from_vibeproxy.py:_write_owner_only` — `0o600` applies only on creation; a pre-existing `.bak`/`.bak.corrupt` written by another tool keeps its prior (possibly looser) mode. `cred_path.parent.mkdir` also uses default umask perms for the `.claude` directory (file contents stay protected, so this is directory-listing exposure only).

dogfood: yes

scarmani:

Head: 86e2ab6 (86e2ab6d87afc2ad23569fb2f83b65bc382e6966), committed 2026-07-22T20:01:12Z.
PR: #9496.
Model family: openai

Verdict: PASS

- [P3] `scripts/sync_claude_profiles_from_vibeproxy.py:137` malformed config whose top-level JSON is not an object can raise `AttributeError` from `raw.get(...)` instead of the documented `ConfigError`, producing an unhandled traceback for a bad local config.
- [P3] `scripts/install_claude_profile_sync_launchd.sh:81` plist content interpolates paths/env-derived values without XML escaping beyond the literal `&&`; unusual values containing `&`, `<`, or `>` can generate an invalid launchd plist.

dogfood: yes

PR #9533, opened 2026-07-23
fix(stream): adapt legacy synchronous event emissions (#8227)

scarmani:
## Summary

- add a typed `emit_sync(...)` compatibility boundary to `SyncEventEmitter`
- register the existing `debate_early_terminated` emission in `StreamEventType`
- cover trickster, early-termination, and unknown-event behavior

## Runtime failure

A real local `aragora ask --crux-cards` run reached the low-novelty trickster path in `DebateConvergenceTracker`. That path receives the CLI's real `SyncEventEmitter` but calls the legacy keyword interface:

```text
event_emitter.emit_sync(event_type="trickster_intervention", ...)
AttributeError: 'SyncEventEmitter' object has no attribute 'emit_sync'
```

The adapter converts that established keyword shape into the authoritative typed `StreamEvent` schema at one boundary. This avoids rewriting the 27 existing static `emit_sync` callers and preserves the existing `emit(StreamEvent)` behavior. Unknown event names are logged and dropped rather than aborting an active debate.

Refs #8227 (ODR-4 crux-card dogfood path). This PR does not close the broader CLI/API/SDK/receipt issue.

## Validation

- `pytest -q tests/stream/test_emitter.py tests/debate/phases/test_convergence_tracker.py` - 81 passed
- adjacent stream/integration selection - 85 passed, 12 deselected
- direct real-emitter trickster smoke - passed
- `ruff check` and `ruff format --check` on touched files - passed
- mypy on touched source files - passed
- local pre-commit and pre-push hooks - passed
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` - passed
- `git diff --check` - passed

## Architecture note

The advisory charter checker reports two PROPOSED `CHR-E-004` path warnings and zero binding violations. `CHR-E-004` specifically excludes new server-local metrics, tracing, or HTTP-pool homes; this PR adds none. One warning is for the existing stream-emitter module and the other is a test import of that existing module, so these are path-only heuristic false positives rather than a new excluded concern.

## Scope

Draft only. No workflow, required-check, settlement, evidence, branch-protection, or public crux-contract change is included.

Issue #9122, opened 2026-07-10
Gate-level passes-in-isolation check before treating shard failures as real (pollution quarantine)


an0mium:
Measured Jul 9-10: tests/server/fastapi/routes/test_swarm_status.py fails in the server-rest xdist shard, passes in isolation (verified twice by independent lanes); CLI critique-store tests same pattern. These blocked #9073/#9047 for hours. Whack-a-mole quarantine is inferior to teaching the gate the check both conductors ran by hand.

## Design
When a test-fast shard fails a ready PR, a bounded follow-up job (or reconciler step) reruns ONLY the failed tests in isolation; if they pass in isolation, the failure is classified 'pollution-suspect' — reported, linked to a quarantine issue, and non-terminal for the live gate (the nightly pristine-main run remains the backstop for real reds).

## Acceptance
- Synthetic order-dependent failure → classified pollution-suspect with the isolation-pass evidence attached; real failure → stays terminal
- Auto-files/updates one quarantine issue per polluted test with occurrence count
- Decision-integrity applied to our own CI (the receipts say WHY a failure was disregarded)
- Work-mix: substrate serving settlement. Tier 2-3; ≤800 LOC

an0mium:
Boss loop exhausted 3 attempts without producing a deliverable. The issue may be too complex for autonomous workers.

an0mium:
Boss loop exhausted 3 attempts without producing a deliverable. The issue may be too complex for autonomous workers.

an0mium:
Boss loop exhausted 3 attempts without producing a deliverable. The issue may be too complex for autonomous workers.

an0mium:
Boss loop exhausted 3 attempts without producing a deliverable. The issue may be too complex for autonomous workers.

an0mium:
Boss loop exhausted 3 attempts without producing a deliverable. The issue may be too complex for autonomous workers.

PR #9537, opened 2026-07-23
fix(observability): keep SLO metric state coherent (#9122)

scarmani:
## Summary

- give the SLO metrics module a complete `_init_noop_metrics()` hook so global disabled-mode initialization cannot set `_initialized=True` while collectors remain `None`
- make every SLO recording surface verify the lazy-init flag and collector objects as one coherent state
- reset the SLO flag and collector objects atomically in the observability fixture
- add a regression that drives the real global no-op initializer before the recovery webhook path

## Incident evidence

This is the source/test-only repair for the SLO state-pollution failure classified on #9533: https://github.com/synaptent/aragora/pull/9533#issuecomment-5058332694

The affected test passed in isolation on both the PR head and its main control while the shard failed with `_initialized=True` and `SLO_LATENCY_HISTOGRAM=None`. The repair removes that invalid state instead of weakening or rerunning the optional gate.

Refs #9122. This PR addresses the concrete SLO pollution specimen; it does not claim to close the broader gate-level pollution-quarantine issue.

## Validation

- `python3 -m pytest -q tests/observability/test_slo_webhook_reliability.py::TestSLORecoveryWebhookReliability::test_recovery_path_survives_global_noop_initialization` — 1 passed
- `python3 -m pytest -q tests/observability/test_slo_webhook_reliability.py tests/observability/metrics/test_slo.py` — 31 passed
- `python3 -m pytest -q tests/observability` — 1,678 passed
- `python3 -m ruff check aragora/observability/metrics/slo.py tests/observability/conftest.py tests/observability/test_slo_webhook_reliability.py` — passed
- `python3 -m ruff format --check aragora/observability/metrics/slo.py tests/observability/conftest.py tests/observability/test_slo_webhook_reliability.py` — passed
- `python3 -m mypy aragora/observability/metrics/slo.py` — passed
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` — passed
- `git diff --check` — passed

## Scope

No workflow, required-check, branch-protection, evidence, settlement, or merge behavior changed. This is a net-negative hard-drain exception: it removes a demonstrated false-red source without adding a new program surface.

### DIFF
--- a/aragora/observability/metrics/slo.py
+++ b/aragora/observability/metrics/slo.py
@@ -224,6 +224,32 @@ SLO_LATENCY_HISTOGRAM: Any = None
 SLO_VIOLATION_MARGIN: Any = None  # How much over the threshold
 
 
+def _slo_metrics_ready() -> bool:
+    """Return whether the lazy-init flag and metric objects are coherent."""
+    return _initialized and all(
+        metric is not None
+        for metric in (
+            SLO_CHECKS_TOTAL,
+            SLO_VIOLATIONS_TOTAL,
+            SLO_LATENCY_HISTOGRAM,
+            SLO_VIOLATION_MARGIN,
+        )
+    )
+
+
+def _init_noop_metrics() -> None:
+    """Initialize the complete SLO metric set with no-op collectors."""
+    global _initialized
+    global SLO_CHECKS_TOTAL, SLO_VIOLATIONS_TOTAL
+    global SLO_LATENCY_HISTOGRAM, SLO_VIOLATION_MARGIN
+
+    SLO_CHECKS_TOTAL = NoOpMetric()
+    SLO_VIOLATIONS_TOTAL = NoOpMetric()
+    SLO_LATENCY_HISTOGRAM = NoOpMetric()
+    SLO_VIOLATION_MARGIN = NoOpMetric()
+    _initialized = True
+
+
 def init_slo_metrics() -> bool:
     """Initialize SLO Prometheus metrics lazily.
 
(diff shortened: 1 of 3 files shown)


scarmani:

Head: 72f2dca (72f2dcab9db369c6de30851e46bd01871a892f3c), committed 2026-07-23T12:42:23Z.
PR: #9537.
Model family: claude

Verdict: PASS

Verified against the repo: the fix targets a real bug — `_init_noop_metrics_all` (`aragora/observability/metrics/initialization.py:94-104`) calls a module's `_init_noop_metrics` hook if present, else just flips `_initialized = True`; before this PR slo.py had no hook, so global noop init left the flag set with all four metric globals `None`, and every record path then crashed on `None.labels()`. The new hook plus the self-healing `_slo_metrics_ready()` check both close that hole. Globals are declared correctly, `NoOpMetric` instances satisfy the readiness check, the conftest reset is coherent with the new logic, and the test's `km_query`/100ms/p99 assertion is deterministic against the 500ms default threshold.

- [P3] `aragora/observability/metrics/slo.py:306` — the reinit path can mask a duplicate-registration failure: if `_initialized` is True with a `None` metric after a *real* prometheus init previously succeeded, the retry re-creates collectors with the same names in the default registry, raising `ValueError`, which is caught and logged as "prometheus-client not installed" while silently downgrading live metrics to NoOp and orphaning the old collectors in the registry. Distinguish the ValueError case in the log message, or reuse/unregister existing collectors on reinit.
- [P3] `aragora/observability/metrics/slo.py` (init path, pre-existing) — lazy init remains lock-free; two threads first-recording concurrently can both enter the prometheus branch, and the loser's duplicate-registration `ValueError` overwrites the winner's real collectors with NoOp. Not a regression, but since this PR touched every call site, a module-level `threading.Lock` around `init_slo_metrics` would be cheap to add.
- [P3] `tests/observability/test_slo_webhook_reliability.py:492` — the test passes even if `_init_noop_metrics` were deleted from slo.py, because `_init_noop_metrics_all` falls back to setting `_initialized = True` and the record call self-heals via the reinit branch. Asserting all four metrics are non-None immediately after `_init_noop_metrics_all()` (before the record call) would pin the new hook specifically.

dogfood: yes

scarmani:

Head: 72f2dca (72f2dcab9db369c6de30851e46bd01871a892f3c), committed 2026-07-23T12:42:23Z.
PR: #9537.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9556, opened 2026-07-23
test(ranking): eliminate calibration integration skips

scarmani:
## Summary

- initialize the calibration integration-test database through the production `EloDatabase` schema path
- remove two broad exception-to-skip wrappers so both workflows execute and fail normally
- merge current `main` additively and regenerate the combined skip audit
- ratchet the current-main skip baseline from 84 to 82

## Why

The Jul 23 external-proof replan fired the quality-ratchet trigger because the skip baseline had risen above the month plan's 77 starting point and 72 target.

A live scan of open PR file surfaces found no duplicate skip-reduction unit for these calibration tests. The two tests were not unavailable: they instantiated the calibration path against an empty SQLite file, then converted the resulting missing-table failures into skips. Initializing through the canonical `EloDatabase` schema makes both tests execute and pass without changing runtime code.

Current `main` independently lowered the skip baseline to 84 for a different repair. The additive restack preserves that reduction and this PR's two removals, yielding the truthful combined count of 82.

## Validation

- `python3 -m pytest -q tests/ranking/test_calibration_engine.py tests/ranking/test_calibration_database.py tests/debate/test_calibration_database.py tests/scripts/test_audit_test_skips.py` — 88 passed
- `python3 scripts/audit_test_skips.py --count-only` — 82
- `python3 -m ruff check tests/ranking/test_calibration_engine.py scripts/audit_test_skips.py tests/scripts/test_audit_test_skips.py` — passed
- `python3 -m ruff format --check tests/ranking/test_calibration_engine.py scripts/audit_test_skips.py tests/scripts/test_audit_test_skips.py` — passed
- `git diff --check origin/main...HEAD` — passed
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` — passed
- pre-commit and pre-push hooks — passed, including mypy and gitleaks

## Scope

Test and generated skip-audit surfaces only. The final PR diff remains exactly:

- `tests/.skip_baseline`
- `tests/SKIP_AUDIT.md`
- `tests/ranking/test_calibration_engine.py`

No runtime, workflow, settlement, evidence, or merge-authority behavior changes.

PR #9561, opened 2026-07-23
test(epistemic): enforce proof-first manifest fixtures

scarmani:
## Summary

- replace two permissive missing-manifest skips with fail-closed assertions
- ratchet `tests/.skip_baseline` from 85 to 84
- regenerate the skip audit and preserve a concise baseline note

The proof-first claims manifest is committed governance evidence, so its disappearance should fail these tests instead of silently reducing coverage. This is separate from the Tier 3 calibration-skip work in #9556.

## Validation

- `python3 -m pytest tests/epistemic/test_truth_map.py tests/epistemic/test_claim_verifier.py -q` (38 passed)
- `python3 scripts/audit_test_skips.py --count-only` (84)
- `python3 -m ruff check tests/epistemic/test_truth_map.py tests/epistemic/test_claim_verifier.py`
- `git diff --check`
- pre-commit and pre-push hooks


scarmani:

Head: 463f993 (463f993af98a56735d0c76bee06890ae843c3f14), committed 2026-07-23T23:23:40Z.
PR: #9561.
Model family: claude

Verdict: PASS

Verification performed:
- `docs/status/claims/proof_first_claims.yaml` **is committed and git-tracked** (since #6051, Apr 16) → converting the `if not exists: pytest.skip` guards to `assert exists` is safe and correct. Fail-closed is an improvement: the silent skip previously masked a regression where the committed fixture goes missing.
- Ran `scripts/audit_test_skips.py --count-only` on main (pre-PR) = **86**. This PR removes exactly 2 `pytest.skip` calls → 86 − 2 = **84**, which matches the new `tests/.skip_baseline` (84) and the regenerated `SKIP_AUDIT.md` total exactly. Both marker-type (40+37+5+2) and category (29+18+17+9+6+4+1) tables sum to 84 — internally consistent.
- `assert` in test modules is pytest-rewritten (not stripped under `-O` the way prod asserts are), so no loss of failure messaging.

- [P3] `tests/SKIP_AUDIT.md:78` — the new note says it "lowers the baseline from 85 to 84," but the true pre-PR *count* was 86 (baseline 85 was already 1 under actual — CI's warn-if-exceeds would have been tripping). The net result (baseline now = actual = 84) is correct, but the narrative understates that it also absorbs a pre-existing +1 drift rather than being a clean −1 from two removed skips.
- [P3] `tests/SKIP_AUDIT.md:74-80` — the rewrite drops the prior baseline-progression history (68→75→77 with per-PR audit-table provenance). Losing that lineage weakens future auditability of baseline bumps; consider retaining a one-line history entry.

No correctness, security, or regression risks. The two [P3]s are documentation-precision nits only.

dogfood: yes

PR #9566, opened 2026-07-24
test(models): enforce provider pricing coverage

scarmani:
## Summary

- replace the vacuous provider-config catalog skip with a fail-closed coverage assertion
- reduce `tests/.skip_baseline` from 84 to 83
- regenerate `tests/SKIP_AUDIT.md` and preserve the ratchet provenance

This is a bounded, non-overlapping step toward the standing skip-baseline target of 72. It does not modify runtime pricing tables. Tier 3 PR #9556 remains untouched and separately human-gated.

## Validation

- `python3 -m pytest tests/models/test_catalog.py -q -rs` -> 42 passed, 6 expected debate-cost skips
- `python3 scripts/audit_test_skips.py --count-only` -> 83
- `python3 -m ruff check tests/models/test_catalog.py`
- `git diff --check`
- commit and pre-push hooks passed


scarmani:

Head: 9812814 (9812814525aafc87b8a6af95b749e3d131861c60), committed 2026-07-24T01:08:11Z.
PR: #9566.
Model family: claude

Verdict: PASS

- [P3] `tests/models/test_catalog.py:156-157` — `test_debate_costs_matches_catalog` still uses the same `pytest.skip("...presence not yet required")` pattern this PR is removing from `test_provider_config_matches_catalog`, leaving the two mirror tests with divergent strictness. Justified here (debate_costs only covers 2/8 enforced models, so tightening it would fail), but worth a tracking note so the asymmetry is deliberate rather than forgotten.
- [P3] `tests/models/test_catalog.py:161-172` — the assertion now makes a `provider_config` row a hard requirement for every enforced model. This is the intended fail-closing, but it couples future catalog additions to routing-pricing coverage: adding a new enforced model without a provider_config row will now redden this test rather than skip. Desirable per the file's philosophy; flagging only so it's an understood contract.

dogfood: yes

PR #9571, opened 2026-07-24
fix(merge): stale rollup rows must not read as green (P1) + inspect all quorum rows

scarmani:
## Summary

Recovered stranded work from the 2026-07-23 worktree cleanup pass (the commit
lived only in a local worktree with no PR ever opened), plus a P1 fix for a
defect that recovery exposed.

## Bug 1 — order-dependent prefilter

`_cheaply_promising()` in `scripts/auto_merge_quorum_green.py` iterated the
status-check rollup and **returned on the first row** matching the quorum check
name. Rollups keep multiple rows per name in arbitrary order, so a stale
`FAILURE` sorted ahead of the current `SUCCESS` made the function return
`False` and the auto-merge daemon skipped a PR that was in fact green. Now it
scans all matching rows, with case normalisation.

## Bug 2 — stale success could authorise a merge (P1, found in review)

**My original description of this PR claimed "the packet fetch remains the
authoritative gate, so no merge authorization is weakened." That was wrong**,
and openai's independent review caught it.

`context_from_gh()` collapsed the rollup by name with **last-write-wins**, so
`check_states` — which `decide_auto_merge()` reads directly to authorise a
merge — depended on GitHub's arbitrary row order. Widening the prefilter feeds
more PRs into that path, so the two behaviours compose into a real hazard.

Duplicates are the norm, not an edge case. On this PR nearly every check name
carries 2+ rows (`changes` had 6). And the vulnerable case occurred **live on
this PR while it was being settled**:

```
aragora-merge-quorum:
  [0] FAILURE  completed 2026-07-24T21:39:21Z   <- current
  [1] SUCCESS  completed 2026-07-24T18:43:02Z   <- stale draft-phase run
```

In that order, last-write-wins reported the quorum check **green** while the
current run had failed.

Rows are now reduced by **recency** (`completedAt`, then `startedAt`; ISO-8601
UTC, so lexical order is chronological). When recency cannot be established —
equal or absent timestamps — the reduction is **fail-closed**: a failing row
wins over a success it cannot be proven newer than.

## Testing

`tests/swarm/test_auto_merge_green.py` + `tests/scripts/test_merge_executor.py`
— **74 passed**. Five new tests; **four fail against the previous
implementation** and pass against this one. The fifth pins the case this fix
must *not* break.

## Reviewed design tradeoffs

- **Recency discriminator vs. "any failure wins":** the blunt version would be
  simpler and also fail-closed, but it would block every PR whose check was
  rerun to green — common after cancelled runs, which this repo hits often.
  `test_rerun_success_after_earlier_failure_reads_green` pins that behaviour.
- **Fail-closed on unrankable ties:** rows without timestamps cannot be proven
  current. Preferring the failing row can at worst cost an unnecessary skip;
  the alternative risks authorising a merge on unprovable evidence.
- **Keeping the prefilter widening:** it is safe *given* bug 2 is fixed. It can
  only cause additional packet fetches, never fewer, and the packet plus the
  now-order-independent `check_states` decide the outcome.
- **Widening vs. latest-row-only in the prefilter:** any-`SUCCESS` keeps the
  cheap stage cheap; correctness is enforced in the authoritative stage rather
  than duplicated in both.

### DIFF
--- a/aragora/swarm/auto_merge_green.py
+++ b/aragora/swarm/auto_merge_green.py
@@ -183,6 +183,50 @@ def first_error_line(stderr: str, stdout: str) -> str:
     return text.splitlines()[0]
 
 
+def _rollup_recency(item: dict[str, Any]) -> tuple[str, str]:
+    """Sortable recency for one rollup row; ``("", "")`` when unknown.
+
+    Timestamps are ISO-8601 UTC as returned by ``gh``, so lexical order is
+    chronological order. Rows missing both stamps rank lowest and are treated
+    as unrankable by the caller.
+    """
+    return (str(item.get("completedAt") or ""), str(item.get("startedAt") or ""))
+
+
+def _reduce_rollup_states(rollup: Any) -> dict[str, str]:
+    """Collapse a status-check rollup to one state per check name.
+
+    Newest row per name wins. When two rows for a name are not comparable by
+    recency (equal or absent timestamps), a failing state wins over a
+    non-failing one so an unprovable success can never authorise a merge.
+    """
+    best: dict[str, tuple[tuple[str, str], str]] = {}
+    for item in rollup or []:
+        if not isinstance(item, dict):
+            continue
+        name = item.get("name") or item.get("context")
+        if not name:
+            continue
+        name = str(name)
+        state = str(item.get("conclusion") or item.get("state") or item.get("status") or "")
+        recency = _rollup_recency(item)
+
+        previous = best.get(name)
+        if previous is None:
+            best[name] = (recency, state)
+            continue
+
+        previous_recency, previous_state = previous
+        if recency > previous_recency:
+            best[name] = (recency, state)
+        elif recency == previous_recency and state in _FAILING_CHECK_STATES:
+            # Same (or unknown) recency: prefer the failing row. Ties must not
+            # be resolved by input order, which is exactly the old defect.
+            best[name] = (previous_recency, state)
+
+    return {name: state for name, (_, state) in best.items()}
+
+
 def context_from_gh(view: dict[str, Any], packet_entry: dict[str, Any] | None) -> PRMergeContext:
     """Build a :class:`PRMergeContext` from a ``gh pr view`` payload + packet entry.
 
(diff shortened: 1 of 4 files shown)


scarmani:

Head: ad5b61c (ad5b61c6853c810fa1112084e8b4697369a233be), committed 2026-07-24T23:35:43Z.
PR: #9571.
Model family: claude

Verdict: PASS

I verified the surrounding context in the repo: the script's flow is prefilter → optional packet fetch → `context_from_gh` → `decide_auto_merge`, so `decide_auto_merge` re-checks the quorum state through the fixed reduction (auto_merge_green.py:150-164), making the loosened `_cheaply_promising` a cost-only change, not a safety change. `merge_executor.py` (scripts/merge_executor.py:298,340) also routes through `context_from_gh`, so it inherits the fix. The core reduction is order-independent, recency-ranked with startedAt-primary, and fail-closed on unrankable ties; the test matrix covers the live incident, the legitimate-rerun inverse, in-flight ranking both directions, and the end-to-end merge decision.

- [P3] Stale docstring contradicts the code: the new paragraph in `context_from_gh` (aragora/swarm/auto_merge_green.py, "reduced by **recency** (``completedAt`` then ``startedAt``)") states completedAt-first ordering, but `_rollup_recency` deliberately orders startedAt-first — the exact regression `test_in_flight_current_run_outranks_stale_completed_success` exists to prevent. Leftover from the first attempt; misleading in a safety-critical rationale.
- [P3] Fail-closed only holds at exact recency equality: a row with no timestamps (e.g., a StatusContext, which gh exports without `startedAt`/`completedAt`) loses to any timestamped row, so a timestamped stale CheckRun SUCCESS would outrank a current untimestamped FAILURE/PENDING sharing the same name (`_reduce_rollup_states`, `recency > previous_recency` branch). Requires a cross-typename name collision — theoretical for this repo's Actions-based checks — but the `_rollup_recency` docstring's "treated as unrankable by the caller" overstates the guarantee.
- [P3] `.upper()` normalization is bidirectional: a lowercase `"success"` row (REST-shaped input) now reads as green where the old code blocked it as `!= "SUCCESS"`. No live change for gh GraphQL data (always uppercase), and it strengthens `_FAILING_CHECK_STATES` matching for lowercase failures, but the docstring's "cannot be bypassed by casing" framing elides that casing now also normalizes *into* SUCCESS.

dogfood: yes

scarmani:

Head: ad5b61c (ad5b61c6853c810fa1112084e8b4697369a233be), committed 2026-07-24T23:35:43Z.
PR: #9571.
Model family: openai

Verdict: PASS

- [P3] `aragora/swarm/auto_merge_green.py:250` Docstring is stale/inconsistent: it says recency is reduced by `completedAt` then `startedAt`, but `_rollup_recency()` intentionally sorts by `startedAt` first. The code and tests align; update the comment to avoid future “fixes” reintroducing the in-flight rerun bug.

dogfood: yes

PR #9541, opened 2026-07-23
ci(benchmark-truth): use fine-scoped PAT for draft PR creation (#9519 option B)

scarmani:
## Summary

Implements founder-approved **option B** from #9519: the daily benchmark-truth-publication workflow runs green but the repository blocks the Actions token from creating PRs, so trust-loop surfaces (`docs/status/B0_BENCHMARK_TRUTH_STATUS.md`, `TW03_RESCUE_PRODUCTIZATION_STATUS.md`) go stale and `benchmark-truth-publication/<run_id>` branches strand (14 as of Jul 22).

- The PR-creation step authenticates with a `BENCHMARK_TRUTH_PR_TOKEN` repo secret when present — a fine-grained PAT scoped to this repo only (`pull_requests: write` + `contents: read` + mandatory `metadata: read`) — falling back to `github.token`, so the change is a no-op until the secret is provisioned (secret was added 2026-07-23).
- Both `find_draft_pr` polls tolerate a failing `gh` (`|| true`) so a dead PAT cannot abort the step under `set -e` before the degrade path runs.
- Any non-zero `gh pr create` exit degrades to the branch-backed handoff with a targeted notice (policy-blocked) or warning (token expired/mis-scoped/transient); a missing PR after a *successful* create still hard-fails.
- Least privilege preserved: all other workflow steps stay on `github.token`.

## Reviewed design tradeoffs

- **Option A rejected** (repo-wide 'allow Actions to create PRs') in favor of a fine-scoped PAT, per the no-standing-admin-keys credential architecture.
- **In-workflow degradation escalation was added and then deliberately removed** after three review rounds each found a real defect in it (absolute-count false reds; pre-PAT backlog tripping the first degraded run; stale `strand-check` refs surviving on the persistent `clean: false` mac-studio runner after operator pruning). Persistent-degradation detection belongs to the **Stage-Gate Conductor**, the external observer that filed #9519 — a workflow whose own credential is the broken piece cannot reliably self-report without false reds. The degrade path is therefore simple and green: handoff + notice/warning.
- **Failure-classification crux (claude r2 vs openai r6) resolved by classifying `gh pr create` failures**: policy block → notice + green handoff; recognizable credential death (401/403/Bad credentials — the planned ≤1y PAT-expiry mode) → warning + green handoff; unrecognized failures (transient API error, script regression) → handoff still written but the run goes red, self-healing next daily run if transient. Green-forever and red-forever are both avoided.
[...]

Issue #9519, opened 2026-07-23
[stage-gate] Benchmark truth publication auto-PR path broken — proof-surface cadence stalled ~6 days before W3 EU AI Act gate


an0mium:
Filed by the Stage-Gate Conductor (audit-only). No `boss-ready` / `autonomous` labels intentionally — this is a bounded operator/permissions issue, not a dispatch task.

## Evidence

- Last operator-driven refresh of `docs/status/B0_BENCHMARK_TRUTH_STATUS.md` and `docs/status/TW03_RESCUE_PRODUCTIZATION_STATUS.md` on `main` was **2026-07-16** (PRs #9491 and #9356). Today is 2026-07-22 — surface is ~6 days stale heading into the W3 EU AI Act GPAI/Art-50 gate (Jul 30) required by `docs/plans/2026-07-09-thirty-day-external-proof-month.md`.
- `.github/workflows/benchmark-truth-publication.yml` (daily 13:20 UTC, self-hosted `mac-studio` runner) runs successfully today but **cannot open the follow-up PR** — repository setting blocks it with:
  > _GitHub Actions is not permitted to create or approve pull requests._
- Workflow falls back to pushing `benchmark-truth-publication/<run_id>` branches and writing a handoff note to the step summary. Diagnosis in the body of the last operator PR (#9356) counts **14 stranded branches** on origin (May 19–27 cluster + Jul 10–16 cluster).
- Latest B0 surface (`docs/status/B0_BENCHMARK_TRUTH_STATUS.md`, updated `2026-07-22T14:01:55Z` via the generator, not via a landed PR) reports headline `truth_success_rate: 100.0%` while proxy metrics show `13/13` corpus rows resolving to `issue_already_resolved` and `unique_issues_succeeded: 0`. The primary metric is technically honest (verified graduated cohort), but the corpus is exhausted and no fresh execution proof is being generated.

## Why this is drift, not maintenance

The current gate (`docs/status/NEXT_STEPS_CANONICAL.md`) explicitly holds the Foreman/proof-loop as a background obligation while the ODR tranche executes in the foreground:

> Do not restock queue work to compensate for stale or dirty observer surfaces; fix the observer and the publication path first.

[...]

PR #9578, opened 2026-07-24
fix(triage): promote daily benchmark-publication drafts to ready (#9519)

scarmani:
## Summary

Closes the last structural gap from #9519: the daily proof-surface publication PR opens automatically (option B / PAT, #9541) but then **sits forever as a draft** because nothing promotes it — so keeping the surface fresh still depends on a human noticing.

Root cause, traced through every merge path:

| Path | Draft handling |
|---|---|
| `auto_merge_quorum_green.py` | filtered at `list_open_pr_numbers` (`:83`), again at `:141`, again in the pure core (`aragora/swarm/auto_merge_green.py:97`) |
| `merge_executor.py` | reuses the same helper (`:87`) |
| `merge_arbiter.py` | explicit "draft PR: never auto-merged" (`:443-464`) |
| `settle_one_pr.py` | draft skip at `:413` and `:849` |

So a draft publication PR is unmergeable by every automation, by design. `pr_ready_triage.py` is the one mechanism built to promote drafts — but `DEFAULT_BRANCH_PREFIXES` was `("codex/",)`, so it never looked at `benchmark-truth-publication/*`.

This adds that prefix to the default list.

## Reviewed design tradeoffs

- **Prefix widening vs. making the workflow open a non-draft PR:** the workflow change is Tier-4 and would also skip the draft-CI stage that currently catches bad publications cheaply. The triage prefix is the smaller, in-idiom change and keeps the draft→verify→promote sequence intact.
- **Prefix widening vs. a new publication-specific settle lane:** rejected as new settlement tooling; composing the existing primitive is the standing preference for this area.
- **Does this widen what can be promoted?** No. Stage 1 only decides *which drafts are looked at*; stage 2 still requires a merge-packet tier in 0–2 with no unresolved dissent and no human-settlement flag, and the never-promote governance markers still veto. A test pins that a Tier-4 publication PR is still refused.

## Verification

- `pytest tests/scripts/test_pr_ready_triage.py` — 59 passed (2 new: default-prefix coverage, and tier-gate still enforced for the new prefix)
- `ruff` + `mypy` clean via pre-commit

## Note on the surrounding lane

Promotion is necessary but not sufficient for a fully unattended daily lane: quorum evidence still needs a local reviewer pass, and the `ctl-merge-executor` daemon is currently self-halted by `~/.aragora/MERGE_EXECUTOR_HALT` (written 2026-07-17 for a `Generate & Validate` failure on main that has since gone green). Both are operator-side and out of scope here.

Refs #9519


scarmani:
Held pending #9582.

Your P2 is correct and is the ordering constraint: promoting these drafts fires the ready-only `build` job, and the publisher never stages its `docs-site/` mirrors, so the promoted PR would go red by construction. #9582 fixes that at the source (sync + stage the two mirrors in the publisher, hard-fail on missing node). This PR should land after it.

P3s addressed in `aa4f4dda06`: rationale comment reworded to say promotion is *manual* rather than claiming the PRs are invisible to every path (they have been merging, by hand), with `check_branch_mutation_policy.py` cited as the provenance guard for the branch namespace; and the tier-gate test now asserts the packet was probed and the rejection reason is the tier, so it fails if the prefix regresses out of the defaults rather than passing on a stage-1 prune.

Your P3 on the freshness of the underlying data is tracked separately in #9580 — the rev-7 restock is blocked upstream because no open candidate has dispatch history.

scarmani:
## Closing — this is the wrong mechanism, and the substantive half already landed

Four review rounds each found a *different* structural mismatch in the same selection path:

1. **Starvable ordering** — 16 open `codex/` drafts against a `--max-scan` of 15 meant the publication PR (always newest, hence last) would never be probed. Fixed with round-robin.
2. **Prefix ≠ provenance** — fork PRs report the bare head branch name, so a fork branch could inherit the namespace's promotion rights. Fixed by pruning cross-repository heads.
3. **Promotion would red the PR** — resolved by #9582 landing.
4. **Oldest-first is inverted here** — it would promote the *stalest* daily snapshot and strand today's.

Each fix was correct, and each round surfaced a new mismatch, because they share one root cause: **`pr_ready_triage.py` models a FIFO queue of worker drafts, and the daily publication PR is a different kind of object** — a single, self-superseding, newest-wins artifact with exactly one live instance per day. Making a FIFO promoter handle it correctly requires per-namespace ordering, per-namespace provenance and per-namespace freshness rules — i.e. re-deriving a second tool inside the first. That is the fragile abstraction the repo's own treadmill rule says to remove rather than patch.

**What actually closed the drift:** #9582 (merged, `d09607ec`). The publisher now stages its own `docs-site` mirrors, verified by dispatch run [30134524906](https://github.com/synaptent/aragora/actions/runs/30134524906) producing #9590 with both mirrors staged automatically. That was the recurring red — and the reason a human had to touch every daily PR.

**What remains** is a smaller, cleaner design decision than this PR: either the publisher opens its PR non-draft (which `check_branch_mutation_policy.py` currently forbids, so that policy would need revisiting), or a dedicated one-shot settle step handles the single live publication PR directly. Both are better homes for newest-wins semantics than a FIFO promoter. Tracking on #9519 rather than forcing this through.

The two generally-useful fixes found here are worth keeping independently of the prefix change — cross-repository heads should never be auto-promoted, and one busy namespace should not starve another. If we want them, they should land as their own PR without the `benchmark-truth-publication/` prefix, since they stand on their own merits.

Closing. Branch retained.

PR #9570, opened 2026-07-24
docs(status): refresh recurring trust-loop surfaces

an0mium:
## Summary

Automated refresh of recurring trust-loop benchmark and rescue status surfaces.

- updates benchmark truth artifacts and scorecards
- updates rescue productization outputs
- refreshes the rendered status pages

Source workflow run: https://github.com/synaptent/aragora/actions/runs/30099191641

PR #9542, opened 2026-07-23
feat(status): surface corpus-exhaustion note on B0 benchmark truth surface

scarmani:
## Summary

Honesty fix for the recurring B0 proof surface, per #9519's finding: `docs/status/B0_BENCHMARK_TRUTH_STATUS.md` leads with a `100.0%` verified headline while the current window's proxy metrics show 13/13 rows neutral (`issue_already_resolved`) and `unique_issues_succeeded: 0` — the rev-6 corpus is exhausted and generating no fresh execution proof. The generated surface never stated that plainly; the honest framing lived only in PR bodies (#9356, #9427), which don't survive onto the surface.

Because the surface is regenerated byte-authoritatively by the daily `benchmark-truth-publication` workflow, a hand-edited note would be clobbered on the next run. This PR instead teaches the renderer to emit the note itself:

- `scripts/render_benchmark_truth_status.py`: when the proxy window is fully neutral (attempted > 0, 0 successes, 0 failures), render a **Corpus exhaustion note** directly under the Truth Metrics headline table: the verified rates reflect the previously graduated cohort, not fresh autonomy proof, and the corpus revision is exhausted until restocked.
- `docs/status/B0_BENCHMARK_TRUTH_STATUS.md`: re-rendered with the note (2-line diff; otherwise byte-identical to the instrument's own output merged in #9538 today).
- Tests: exhausted window → note present above Proxy Metrics; fresh successes → note absent.

The note disappears automatically once a restocked corpus (TW-01 rev-7) produces a fresh success or failure — no manual cleanup.

## Reviewed design tradeoffs

- Renderer-emitted vs hand-edited note: hand edits are overwritten daily by the workflow; only a generator change is durable.
- Placement under the headline table (not in Proxy Metrics): the mission of the note is that the 100% headline cannot be read without the exhaustion context.
- Trigger is fully-neutral-window (succeeded == 0 and failed == 0 and neutral == attempted): a window with any fresh success *or* fresh failure is producing execution evidence and should not carry the note.

## Verification

- `pytest tests/scripts/test_render_benchmark_truth_status.py tests/scripts/test_benchmark_truth_publication_workflow.py` — 34 passed
- `python3 scripts/check_docs_consistency.py` — no findings
- `python3 scripts/probe_proof_surface_freshness.py` — b0 + tw03 fresh, exit 0
- `ruff check` / `ruff format --check` / `mypy` on changed files — clean
- Re-render against main's published JSON artifacts is byte-identical except the 2 new lines

[...]

PR #9582, opened 2026-07-24
ci(benchmark-truth): sync docs-site mirrors in the publisher (#9519)

scarmani:
## Summary

The daily publication PR has been red on `build` by construction, and nobody noticed because a human quietly fixed it each time.

`docs-build.yml` runs `node scripts/sync-docs.js` then `git diff --quiet` ("Ensure docs are synced") and fails closed. The publisher renders `docs/status/B0_BENCHMARK_TRUTH_STATUS.md` and `TW03_RESCUE_PRODUCTIZATION_STATUS.md`, but its `git add` list never included the corresponding `docs-site/docs/contributing/*.md` mirrors — so every daily PR arrives with mirror drift. Merging in that state also hands the red to the next unrelated `docs/**` PR.

Evidence this is recurring, not hypothetical:
- Today's #9570 arrived with 10 files; the two mirrors were added by a follow-up human commit (`7ad0e2cc3`) before it could merge.
- Yesterday's #9542 hit the identical failure and needed the same manual `sync-docs.js` + commit.
- The reviewer independently flagged it on #9578 as a P2 blocker for automating promotion.

This makes the sync part of publishing: run `sync-docs.js` before staging, stage the two mirrors, and hard-fail if `node` is missing rather than publishing known-drifted surfaces.

## Reviewed design tradeoffs

- **Fix in the publisher vs. in the promoter:** the drift is created at publication time; fixing it downstream would leave the manual step for anyone who publishes by hand.
- **Hard-fail vs. skip on missing node:** skipping would silently reintroduce the exact drift this removes. The runner already probes `node --version` in its prerequisites step, so the toolchain is expected to be present.
- **Stage the two named mirrors vs. all of `docs-site/`:** naming them keeps the commit scoped to what this workflow renders and avoids sweeping in unrelated regenerated site files.
- **No untrusted input** is interpolated into the added shell (static commands only).

## Verification

- `pytest tests/scripts/test_benchmark_truth_publication_workflow.py` — 9 passed, including a new test pinning that the sync runs *before* `git add`, that both mirrors are staged, and that a missing toolchain errors.
- Behaviour matches what the last two publications needed by hand.

## Tier

Workflow change — Tier 4, operator settlement required. Not auto-settled.

Unblocks #9578 (auto-promoting these drafts fires the ready-only `build` job, which fails by construction until this lands).

Refs #9519

### DIFF
--- a/.github/workflows/benchmark-truth-publication.yml
+++ b/.github/workflows/benchmark-truth-publication.yml
@@ -348,6 +348,20 @@ jobs:
         run: |
           git config user.name "github-actions[bot]"
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+          # The rendered status pages are mirrored into docs-site, and
+          # docs-build.yml fails closed when a docs/** change lands without its
+          # mirror ("Ensure docs are synced"). Publishing without this leaves
+          # every daily PR red on `build` and, once merged, hands that red to
+          # the next unrelated docs PR — so the sync belongs to the publisher,
+          # not to whoever happens to notice. Hard-fail on a missing toolchain
+          # rather than publishing known-drifted surfaces.
+          if ! command -v node >/dev/null 2>&1; then
+            echo "::error::node is required to sync docs-site mirrors"
+            exit 1
+          fi
+          ( cd docs-site && node scripts/sync-docs.js )
+
           git add \
             docs/benchmarks/benchmark_corpus_freshness.json \
             docs/benchmarks/rescue_productization.json \
@@ -355,7 +369,9 @@ jobs:
             docs/status/TW03_RESCUE_PRODUCTIZATION_STATUS.md \
             docs/status/generated/benchmark_scorecards \
             docs/status/generated/benchmark_truth_artifacts \
-            docs/status/generated/rescue_productization
+            docs/status/generated/rescue_productization \
+            docs-site/docs/contributing/b0-benchmark-truth-status.md \
+            docs-site/docs/contributing/tw03-rescue-productization-status.md
 
           if git diff --cached --quiet; then
             echo "No recurring trust-loop surface changes to commit."
--- a/.github/workflows/benchmark-truth-publication.yml
+++ b/.github/workflows/benchmark-truth-publication.yml
@@ -332,17 +332,23 @@ jobs:
 
       - name: Sync docs-site mirrors
         shell: bash
-        # Deliberately uncredentialed and separate from the publish step: the
-        # rendered status pages are mirrored into docs-site, and docs-build.yml
-        # fails closed when a docs/** change lands without its mirror ("Ensure
-        # docs are synced"). Publishing without this leaves every daily PR red
-        # on `build` and, once merged, hands that red to the next unrelated
-        # docs PR. Running the same generators docs-build runs — in the same
-        # order — is what makes the published tree satisfy that gate; keeping
-        # it out of the publish step keeps repository JavaScript away from the
-        # PAT and makes a sync failure attributable in the run log.
+        # Scope: repair the mirror drift this workflow itself creates, nothing
+        # more. The status pages it renders are mirrored into docs-site, and
+        # docs-build.yml fails closed when a docs/** change lands without its
+        # mirror — so without this every daily PR is red on `build` and, once
+        # merged, hands that red to the next unrelated docs PR.
+        #
+        # This deliberately does NOT re-implement docs-build's full closure
+        # (doc_stats, capability-matrix, legacy-entrypoint and CLI-reference
+        # checks). Chasing parity here means re-deriving another job's gate
+        # inside a publisher that cannot see it change, and pre-existing drift
+        # from elsewhere is not this workflow's to absorb. docs-build stays the
+        # authority; this only stops the publisher from being a drift source.
+        #
+        # Uncredentialed and separate from the publish step so repository
+        # JavaScript never runs alongside the PAT, and a sync failure is
+        # attributable in the run log.
         run: |
-          python3 scripts/doc_stats.py --write
           cd docs-site && node scripts/sync-docs.js
 
       - name: Publish workflow summary
@@ -370,9 +376,6 @@ jobs:
           git config user.name "github-actions[bot]"
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
 
-          # Stage the whole mirror tree, not just this workflow's two pages:
-          # the gate in docs-build.yml is a whole-tree `git diff --quiet` after
-          # running the same generators, so a partial stage still reds `build`.
           git add \
             docs/benchmarks/benchmark_corpus_freshness.json \
             docs/benchmarks/rescue_productization.json \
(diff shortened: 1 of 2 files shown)


scarmani:
## Round 3 — removed the replication instead of widening it

Round 2's P1 and P2 were both consequences of one choice: making the publisher satisfy `docs-build`'s whole gate. Each round then surfaced another piece of that closure I'd failed to replicate — `doc_stats` rewriting embedded counts in `README.md`/`STATUS.md` outside the guard's pathspec, untracked files invisible to `git diff`, and three further checks the gate runs (capability matrix, legacy entrypoints, CLI reference). That is the nitpick-treadmill signature, and the cure is to remove the fragile abstraction rather than patch it a third time.

The publisher now repairs only the drift **it** creates:

| Round 2 finding | Resolution |
|---|---|
| P1 — guard pathspec narrower than the gate | Guard removed; `docs-build` is the authority on tree cleanliness |
| P2 — allowlist + fail-closed = detect-and-die | Removed; `git add -u docs-site/docs` stages what the sync actually rewrote |
| P3 — `git diff` blind to untracked | Moot: no guard |
| P3 — `-A` sweeps untracked leftovers under `clean: false` | Now `-u` (tracked-only) |
| P3 — replicates only half the closure | No longer attempts replication; scope stated explicitly in the step comment |
| P3 — test locks in the narrow pathspec | Test rewritten; asserts step order and tracked-only staging, and pins that `-A docs-site` is *not* used |

`doc_stats.py --write` is gone — it was added in round 2 to mirror the gate and was precisely what pushed the blast radius past the guard. The publisher renders two status pages; their mirrors are `sync-docs`' job alone. Pre-existing drift originating elsewhere is not this workflow's to absorb.

Retained from round 2 because they were correct and independent of the replication question: `actions/setup-node@v4` (pinned 20, matching docs-build) so node is provisioned rather than asserted on the self-hosted runner, and the sync living in its own uncredentialed step so repository JavaScript never runs alongside `BENCHMARK_TRUTH_PR_TOKEN`.

Note: two backticked fragments were eaten by the shell in commit `5d6913b`'s message ("Stages  — every mirror…"). The diff is unaffected; not force-pushing to fix a cosmetic message defect.

scarmani:
## Round 4 — dispatch verification found a job-killing defect

Your P2 was right, and acting on it caught something CI could not have.

**First dispatch from this branch ([run 30134420462](https://github.com/synaptent/aragora/actions/runs/30134420462)) failed.** The job died *inside* `actions/setup-node` — 13 seconds in, with Setup Node still `in_progress` and all eleven later steps `pending`. `continue-on-error` could not rescue it because what failed was the runner, not the step. **Merging round 3 would have ended the daily publication entirely, every day** — a strictly worse outcome than the red `build` this PR set out to fix.

So provisioning was the wrong remedy for round 2's "asserted, not provisioned" finding. The mac-studio runner already has **node v23.5.0**, probed every run by `Ensure local toolchain on PATH`, and `sync-docs.js` uses only `fs`/`path`. `actions/setup-node` is gone; the sync uses ambient node and degrades with a `::warning::` if node is ever absent. A test pins that no `setup-node` step returns.

**Second dispatch ([run 30134524906](https://github.com/synaptent/aragora/actions/runs/30134524906)) succeeded end to end**, and produced [#9590](https://github.com/synaptent/aragora/pull/9590):

- `Sync docs-site mirrors`: success
- `Commit, publish, and verify draft PR`: success
- **14 files staged, including both `docs-site/docs/contributing/*.md` mirrors** — automatically, with no human commit. Today's #9570 needed a hand-pushed mirror commit to reach the same state.

That is the behaviour change this PR exists for, demonstrated on the real runner rather than asserted by a YAML string test.

Also fixed from round 3: the sync no longer aborts the publication (it runs after `--ensure-issues` has already mutated issues, so aborting would discard the day's surfaces — the degrade-and-warn contract now applies), and two comments that overclaimed were corrected — staging the sync output *does* absorb unrelated mirror drift, and moving the sync out of the publish step narrows the PAT's exposure but not the checkout token that persists in `.git/config` for the whole job.

#9590 is a genuine second publication for today; it can settle normally once this lands.

scarmani:

Head: 3e0fa7b (3e0fa7b518caa17bcad89e7cb778ad6d0baee0f2), committed 2026-07-24T23:43:08Z.
PR: #9582.
Model family: claude

Verdict: PASS

Verified locally: workflow YAML parses, step order is render(11) → sync(12) → summary(13) → commit(14), triggers are `schedule`/`workflow_dispatch` only (no fork exposure), `sync-docs.js` requires only `fs`/`path` (lines 12–13) and writes solely under `DEST_DIR = docs-site/docs` with no unlink/rm — so the ambient-node claim and the `docs-site/docs` pathspec both hold. `git add -u` is also consistent with the authority gate, which is `git diff --quiet` (docs-build.yml:120), i.e. tracked-only. Both mirrors it touches are tracked. All 10 tests pass.

- [P3] `.github/workflows/benchmark-truth-publication.yml:407` — staging mirrors before the emptiness check changes what the early exit means: on a day with no status delta, unrelated docs-site drift already on main is enough to make `git diff --cached --quiet` false (line 409), producing a branch + draft PR titled "refresh recurring trust-loop surfaces" that refreshes no surface. Self-limiting (the first such PR repairs the drift), but it adds PR/branch noise to the exact accumulation the Stage-Gate Conductor watches. The comment names the absorption, not this consequence.
- [P3] `.github/workflows/benchmark-truth-publication.yml:364` — the degrade path warns but does not roll back. `sync-docs.js` writes per-file in a loop, so a mid-loop crash leaves a partially regenerated mirror set that `git add -u` then commits and publishes. Not worse than the status quo for `build` (red either way), but the commit carries a half-sync. A `git checkout -- docs-site/docs` in the failure branch would make degrade mean "no-op" rather than "commit whatever got written".
- [P3] `tests/scripts/test_benchmark_truth_publication_workflow.py:214` — ordering is asserted only against the commit step. Moving the sync *above* "Publish tracked trust-loop surfaces" (step 11) keeps both new tests green while syncing mirrors from pre-render sources — i.e. reintroduces this exact bug silently. Add `names.index("Publish tracked trust-loop surfaces") < names.index("Sync docs-site mirrors")`.
- [P3] `.github/workflows/benchmark-truth-publication.yml:364` — `assert "||" in sync_run` (test line 240) matches any `||` anywhere in the block; it does not pin the fallback to the node invocation. Minor, but it is the only guard on the non-fatal contract.
- [P3] `.github/workflows/benchmark-truth-publication.yml:407` — `-u` cannot stage a *newly created* mirror (`sync-docs.js` mkdir/writes new dests at lines 941–945, 978–982). Harmless today (all DOC_MAP dests this workflow touches are tracked, and docs-build's own gate is tracked-only), but if DOC_MAP or the category list grows, the new mirror stays untracked across runs — `clean: false` means it never gets swept, and nothing warns.

dogfood: yes

scarmani:

Head: 3e0fa7b (3e0fa7b518caa17bcad89e7cb778ad6d0baee0f2), committed 2026-07-24T23:43:08Z.
PR: #9582.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

PR: #9582
Exact head: 3e0fa7b518caa17bcad89e7cb778ad6d0baee0f2
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #9582 is non-draft and live exact-head checks/merge-packet remain otherwise green.

PR #9506, opened 2026-07-22
feat(cli): expose crux cards via --crux-cards flag

scarmani:
## Summary

Closes the named W3 EU AI Act bundle gap from #9427: *"crux-cards receipt — needs API-key debate with `enable_crux_cards=True` — no CLI flag exposes it"*. A crux-cards decision receipt can now be produced without writing Python.

- Add `--crux-cards` flag (default off) to `aragora ask`, wired through `protocol_overrides` to `DebateProtocol.enable_crux_cards` — same pattern as the existing cross-pollination feature flags.
- `DecisionReceipt.from_debate_result` (used by `--decision-integrity`) and the auto-persisted `~/.aragora/receipts` receipt both carry the crux block via shared helpers (`crux_cards_from_metadata`, `receipt_schema_version`).
- **Uniform local-only rule:** `--crux-cards` requires local execution. `--graph`/`--matrix` rejected unconditionally; explicit API config (`--api`, `--api-url`, exported `ARAGORA_API_URL`) rejected only when the run is not already local (`--local`/`--demo`/`ARAGORA_OFFLINE`); auto-discovery case forces local with a stderr `Note:`.
- **Receipt schema 1.2 + unified, version-gated hash:** crux receipts stamp `schema_version` 1.2; the single canonical recipe `compute_receipt_artifact_hash()` (shared by `DecisionReceipt._calculate_hash` and `aragora verify`) binds `cruxes` whenever present and binds `schema_version` only for 1.2-stamped receipts. Legacy 1.1 receipts — including #9414-era crux receipts persisted before the stamp existed — keep their exact stored hashes. `aragora verify` requires the full artifact hash for crux/1.2 receipts (legacy 16-char checksum fallback stays pre-crux-only).

## Bundle runbook invocation

```bash
aragora ask "<decision question>" --agents anthropic-api,openai-api --local --crux-cards --decision-integrity
```

(`--local` is optional — `--crux-cards` forces local with a stderr Note — but kept explicit in the runbook for clarity.) The persisted receipt (and the decision-integrity package receipt) carry the `cruxes` block at schema_version 1.2 and verify under both `aragora verify` and `aragora receipt verify`.

## Reviewed design tradeoffs

Round 6 (compat edges), all addressed:

[...]

PR #9427, opened 2026-07-20
docs(compliance): W3 EU AI Act bundle skeleton (Jul 2026)

scarmani:
## Summary

Assembles the **W3 EU AI Act bundle** (thirty-day external-proof plan, W3 Jul 23-30; publish target **Jul 30**, content-complete **Jul 27**) as a draft under `docs/compliance/bundles/2026-07-eu-ai-act/`.

The branch now carries bundle files only and is additively restacked on main `9076464c796b4eb0fd708c7080d209771f83d29f`. Production remains unavailable under #9391, so the bundle can ship either with a production-signed receipt (Variant A) or with the included locally signed receipt plus the dated production-signing gap statement (Variant B, current default).

## Contents

| Artifact | File | Status |
|---|---|---|
| Bundle index + Variant A/B contingency | `README.md` | present |
| Art. 14 oversight pack (JSON + Markdown) | `oversight-pack.{json,md}` | present |
| Source receipt: real 3-agent debate (#8263 packaging decision, 2026-06-22) | `receipts/2026-06-22-packaging-decision-8263.receipt.json` | present |
| Variant B signed ODR (Ed25519, bundle-local key) + public key | `packaging-8263.signed-local.odr.json`, `local-signing.pubkey.pem` | present |
| Verification walkthrough | `verification-walkthrough.md` | present |
| Rekor transparency-log procedure | `rekor-note.md` | present; external publication pending operator |
| ODR-2 #8225 closure draft | `odr2-closure-draft.md` | present; founder review pending |
| Variant A production-signed receipt | - | pending production recovery (#9391) |
| Crux-cards receipt | - | pending one credentialed debate with `enable_crux_cards=True` |
| Founder earned-claim review | - | pending founder review |

The bundle index classifies each remaining gate as infrastructure/operator, external-publish/operator, credentialed machine work, or founder review. None is silently treated as complete.

## Exact-head validation

Head: `f953991e1c247940b094e25021ab1d96568eb35f`

- Metrics/doc generation: three resolved-index passes; second and third passes produced no diff.
- `pytest -q tests/scripts/test_regenerate_metrics.py tests/scripts/test_doc_stats.py tests/scripts/test_docs_site_sync_links.py`: **67 passed**.
- `pytest -q tests/compliance/test_oversight_pack.py tests/compliance/test_oversight_fetch.py`: **41 passed**.
- `PYTHONPATH=aragora-verify/src pytest -q aragora-verify/tests`: **106 passed**.
[...]

PR #9583, opened 2026-07-24
docs(compliance): correct crux-cards gap attribution in W3 bundle (#9581)

### LINKED ISSUE #9581: Crux cards can never be produced: belief network built without factor edges (blocks W3 exit criterion)
## Summary

Crux cards cannot be produced by a standard debate. Both code paths that build
the `BeliefNetwork` for crux detection add **claims only** and never add the
**factor edges** (`RelationType.SUPPORTS` / `CONTRADICTS`) that
`CruxDetector.compute_disagreement_scores()` requires. With no factors,
`total_disagreements` is always `0`, no crux ever clears any threshold, and the
`cruxes` block is never attached.

This makes the `--crux-cards` CLI flag (#9506) and the crux-cards receipt
feature (#8227 / #9414) inert for standard debates at any threshold, with any
agents, on any question.

## Affected call sites

- `aragora/debate/crux_cards.py::_network_from_messages` — `add_claim` only
- `aragora/debate/phases/winner_selector.py::analyze_belief_network` (~line 361) — `add_claim` only

Note `ctx.belief_network` is only populated when `enable_km_belief_sync=True`
(`aragora/debate/orchestrator_runner.py:806`), so by default `crux_cards` always
takes the message-fallback path above.

## Why it can never fire

`CruxDetector.compute_disagreement_scores()` walks
`self.network.node_factors[node_id]` and groups incoming evidence by author,
requiring `len(author_beliefs) >= 2` to score any disagreement. `add_claim()`
creates nodes via `add_node_from_claim` and never creates factors, so
`node_factors` is empty for every node and the loop body never executes.

## Reproduction

Two real debates via `aragora ask --agents claude,codex --crux-cards` (2 and 3
rounds, majority consensus, deliberately contested prompt) both produced
receipts with `schema_version: 1.1` and **no** `cruxes` block.

Isolated proof that the detector is fine and only the wiring is missing —
same claims, same detector, only factor edges added:

```python
n = BeliefNetwork(max_iterations=3)
n.add_claim('c1', '...', 'claude'); n.add_claim('c2', '...', 'codex'); n.add_claim('c3', '...', 'claude')
# without these three lines: cruxes=0, disagreements=0
n.add_factor('c2', 'c1', RelationType.CONTRADICTS, strength=0.9)
n.add_factor('c3', 'c1', RelationType.SUPPORTS,    strength=0.9)
n.add_factor('c3', 'c2', RelationType.CONTRADICTS, strength=0.9)
CruxDetector(n).detect_cruxes(top_k=5, min_score=0.1)
# => cruxes: 3, disagreements: 2
```

Also note `contesting_agents` came back `[]` even for detected cruxes, which
looks like a second, smaller defect worth checking alongside the fix.

## Impact

- **Blocks a W3 exit criterion** of the thirty-day external-proof month:
  "crux cards in ≥1 published receipt"
  (`docs/plans/2026-07-09-thirty-day-external-proof-month.md`).
- The July EU AI Act bundle currently attributes this gap to needing "a real
  (API-key) debate run"
  (`docs/compliance/bundles/2026-07-eu-ai-act/README.md`). That attribution is
  wrong — credentials are not the blocker — and should be corrected before the
  bundle is published, per its own honesty contract.
- #9414 shipped crux cards into the receipt schema as a W2 deliverable; the
  schema plumbing is real, but nothing can populate it.

## Suggested fix direction

Populate relations between proposer/critic claims before running the detector.
Per the project's "use real intelligence, not regex" principle, stance
detection between two claims should be model-based rather than keyword
matching. The critic→proposer relationship in a standard debate is already
structurally known (a critique targets a specific proposal), so a minimal
correct fix may be able to derive edges from message structure and use a model
only for the SUPPORTS-vs-CONTRADICTS polarity.

Acceptance: a standard 2-agent debate on a contested question yields a receipt
whose `cruxes` block has ≥1 item with non-empty `contesting_agents`.

scarmani:
## Summary

Corrects a false gap attribution in the July EU AI Act bundle before it is
published (W3 exit criterion, publish target Jul 30). The bundle's own honesty
contract says absence must be recorded accurately, never implied away — this
fixes a case where the recorded cause was wrong.

## What was wrong

The bundle's "Crux cards (pending)" section claimed:

1. "no CLI flag exposes `enable_crux_cards` yet" — false since #9506 merged
   2026-07-22 (after the bundle was assembled on 07-20): `aragora ask --crux-cards`.
2. "generating one requires a real (API-key) debate run" — false. Credentials
   are not the blocker.

The real blocker is a product defect, filed as #9581: both paths that build the
`BeliefNetwork` for crux detection (`crux_cards._network_from_messages` and
`winner_selector.analyze_belief_network`) call `add_claim` but never
`add_factor`, so `CruxDetector.compute_disagreement_scores()` finds
`total_disagreements == 0` and no crux clears any threshold. Crux cards
therefore cannot be produced by any standard debate, with any agents, at any
threshold.

## Evidence

- Two real provider-backed debates on 2026-07-24
  (`aragora ask --agents claude,codex --crux-cards`, 2 and 3 rounds,
  deliberately contested prompts, majority consensus): both receipts came back
  `schema_version: 1.1` with no `cruxes` block.
- Isolated test: the same three claims produce **0** cruxes with claims alone
  and **3** cruxes once `SUPPORTS`/`CONTRADICTS` factor edges are added by
  hand — the detector is sound, only the wiring is missing.
- Production re-probed directly on 2026-07-24; `api.aragora.ai/health` returned
  no response, so Variant A stays blocked and the status-check date is refreshed.

## Scope

Bundle README only — no other bundle artifact changes, and the signed ODR bytes,
digests and signatures are untouched. Per the bundle's PR-hygiene rule, no
generated surfaces are included.

The gap itself stays **open**. This PR does not close it or weaken it; it only
makes the stated cause true, and re-classifies it from
"machine-capable, credentialed" to "product defect" — which also makes it the
one remaining W3 gap that is neither operator-held nor blocked on AWS.

## Reviewed design tradeoffs

- **Correct the note vs. fix #9581 first and skip the note:** the bundle has a
  Jul 30 publish target and #9581 needs model-based stance detection to do
  properly (per the repo's "use real intelligence, not regex" principle).
  Publishing with a knowingly-wrong cause was the one unacceptable option, so
  the honest note lands now and the fix proceeds on its own timeline.
- **Re-attribute vs. quietly drop the claim:** leaving the gap listed with an
  accurate cause is strictly more useful to an auditor than removing the row,
  and matches how the bundle already handles the Variant A / AWS gap.

Closes nothing. Tracks #9581. Part of the W3 bundle (#9427).

### DIFF
--- a/docs/compliance/bundles/2026-07-eu-ai-act/README.md
+++ b/docs/compliance/bundles/2026-07-eu-ai-act/README.md
@@ -4,9 +4,12 @@
 **Plan:** [`docs/plans/2026-07-09-thirty-day-external-proof-month.md`](../../../plans/2026-07-09-thirty-day-external-proof-month.md) (W3, Jul 23–30)
 **Scope:** GPAI / Art. 50 transparency posture ahead of the **Aug 2, 2026** EU AI Act GPAI deadline, plus the Article 14 human-oversight evidence pack (#8230 / ODR-6).
 **Assembled:** 2026-07-20
-**Status checked:** 2026-07-21 against current main. Issue #9391 remains open;
-its latest recorded external probe (2026-07-20) still classifies production as
-unreachable.
+**Status checked:** 2026-07-24 against current main. Issue #9391 remains open
+and production was re-probed directly on 2026-07-24 (`https://api.aragora.ai/health`
+returned no response), so Variant A stays blocked and Variant B remains the
+shipping default. The crux-cards gap was re-attributed on the same date after
+direct testing — see "Crux cards" below and
+[#9581](https://github.com/synaptent/aragora/issues/9581).
 
 This bundle packages what Aragora can already prove — signed, dissent-preserving
 decision receipts with third-party offline verification — into one auditable
@@ -27,7 +30,7 @@ with an honest pending status and the exact step that closes it.
 | 8 | Rekor transparency-log note | `rekor-note.md` | present (log entry itself: **pending** — submission is an external publish, held for operator) |
 | 9 | ODR-2 (#8225) closure comment draft | `odr2-closure-draft.md` | present (**pending-founder-review** — do not post until reviewed) |
 | 10 | Signed **production** receipt — **Variant A** | — | **pending-prod** (blocked on AWS reinstatement, [#9391](https://github.com/synaptent/aragora/issues/9391)) |
-| 11 | Crux-cards receipt (`cruxes` block populated) | — | **pending** (see "Crux cards" below) |
+| 11 | Crux-cards receipt (`cruxes` block populated) | — | **pending — blocked by product defect [#9581](https://github.com/synaptent/aragora/issues/9581)** (see "Crux cards" below) |
 | 12 | Founder earned-claim review of the bundle | — | **pending-founder-review** (W3 exit criterion) |
 
 ## Remaining exit gates
@@ -36,7 +39,7 @@ with an honest pending status and the exact step that closes it.
 |---|---|---|
 | Production-signed receipt | infrastructure/operator | Restore production access under #9391, then run the documented Variant A export and independent verification. |
 | Rekor entry | external publish/operator | Review `rekor-note.md`, publish the digest once, and record the returned UUID. |
-| Crux-cards receipt | machine-capable, credentialed | Run one real provider-backed debate with `enable_crux_cards=True`, verify the populated `cruxes` block, and add the receipt. |
+| Crux-cards receipt | **product defect** | Fix [#9581](https://github.com/synaptent/aragora/issues/9581) (belief network built without factor edges), then run one debate with `aragora ask --crux-cards`, verify the populated `cruxes` block, and add the receipt. Not blocked on credentials or on AWS. |
 | Earned-claim and ODR-2 text | founder review | Review this bundle and `odr2-closure-draft.md`; publication and issue closure remain separate operator actions. |
 
 The repository can prepare and validate these artifacts, but none of the four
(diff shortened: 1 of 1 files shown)


scarmani:

Head: 20f4e57 (20f4e57fa88f05cf46e70528e70ffc3d043b78bd), committed 2026-07-24T21:58:26Z.
PR: #9583.
Model family: claude

Verdict: PASS

- [P3] `docs/compliance/bundles/2026-07-eu-ai-act/README.md` ("Gap re-attributed" section / exit-gates row): the claim "the detector itself is sound" and the recipe "Fix #9581, then run one debate" are now under-scoped — issue #9644 (opened after this PR's Jul 24 status check) documents additional defects (dissent attribution: propagate-before-measure, missing author stance), so fixing #9581 alone may yield a populated but misattributed `cruxes` block. The doc is honestly dated and the "verify the populated `cruxes` block" step gives a catch point, so this is non-blocking; reference #9644 alongside #9581 at the next status check.
- [P3] Same section: "merged 2026-07-22" for #9506 is local-time (GitHub records 2026-07-23T02:43Z); harmless, but a compliance bundle read by external auditors could state the UTC date to avoid a spurious discrepancy.

dogfood: yes

scarmani:

Head: 20f4e57 (20f4e57fa88f05cf46e70528e70ffc3d043b78bd), committed 2026-07-24T21:58:26Z.
PR: #9583.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9075, opened 2026-07-09
feat(models): GPT-5.6 Sol + Claude Fable 5 frontier upgrade

an0mium:
Founder-directed (Jul 9): GPT → **gpt-5.6-sol** wherever beneficial; Claude Opus → **claude-fable-5** where quality-per-call dominates, **Opus 4.8** elsewhere. Part of #9069; stacked on #9073 (retarget to main after it merges).

## Moved to Sol (price-neutral: $5/$30 same as 5.5, Terminal-Bench 88.8 vs 82.7)
openai-api agent default + registry, legacy-upgrade map targets, `reviewer` role pin, selector profile.

## Moved to Fable 5 ($10/$50 — quality-per-call roles only)
`judge`/`security_auditor`/`compliance_auditor` role pins; claude CLI default (Max subscription — zero marginal cost); OR identity alias; selector profile.

## Deliberately NOT moved
Codex-CLI reviewer harness + quorum reviewer map stay on gpt-5.5 until ~Jul 23 (14-day rule — day-0 models don't drive merge-authority evidence). Opus 4.8 keeps bulk debate roles (2x cost unjustified at volume).

## Guardrails
Billing + routing pricing rows for both models (receipts and the Pareto optimizer never see $0.00 — the #9064 lesson). 688 tests green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9588, opened 2026-07-24
feat(models): Claude Opus 4.8 -> Claude Opus 5 frontier bump

scarmani:
## Summary

- Bumps the Anthropic frontier pin repo-wide from `claude-opus-4-8` to **`claude-opus-5`** (`$5/$25` per MTok — identical economics, so no cost delta), across the canonical catalog, all runtime price mirrors, role pins, agent configs, harness defaults and call sites.
- **Retains `claude-opus-4-8`** in the catalog, every price mirror and all legacy-remap keys — it is still Active upstream and is Opus 5's documented fallback target for cyber-classifier refusals.
- Sizes `max_tokens` for Opus 5's **adaptive-thinking-by-default** behaviour at the four raw call sites that omit a `thinking` field (see *Behavioural change* below) — this is the part a string swap cannot express.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Bug fix (non-breaking change that fixes an issue)
- [x] Documentation update

## Reviewed design tradeoffs

Please read this section before filing findings — the three most flaggable choices here were made deliberately.

### 1. The 14-day availability rule (`soak_until`) is WAIVED for this bump

Opus 5 is a **day-0 release**. Under `ModelSpec.soak_until` it would normally carry `soak_until=2026-08-07`, which would bar it from merge-authority evidence and routing candidate enumeration — exactly the constraint settled on #9075, where flipping a pin early was a convergent review finding.

**This was an explicit operator decision, not an oversight.** The spec carries `soak_until=None` with an inline comment recording the waiver and the exact one-line revert (`soak_until=date(2026, 8, 7)`).

*Known consequence, disclosed:* `DEFAULT_CLAUDE_MODEL` in the VibeProxy evidence transport is now `claude-opus-5`, and VibeProxy sits first in the quorum reviewer fallback chain — so merge-authority evidence for this PR can be produced by the unsoaked model it adopts. The CLI reviewer path is unaffected (it pins `claude-fable-5`). Flagging rather than silently relying on it.

### 2. Opus 4.8 is kept, not replaced

Blind-replacing every `claude-opus-4-8` string would have de-registered a model that is still served. It stays in `CATALOG`, `_PRICE_PER_MTOK`, `PROVIDER_PRICING`, `MODEL_PRICING`, `DEFAULT_PROVIDER_RATES`, `MODEL_TIERS`, `MODEL_PROFILES` and as a legacy-remap **key** (only remap *values* moved to Opus 5).

### 3. Historical records were deliberately not rewritten

`docs/status/**`, `docs/plans/**`, `docs/superpowers/**`, `docs-site/**`, `docs/specs/examples/*.odr.json` and the vibeproxy burn-in capture still say Opus 4.8. Rewriting a dated status report or a **signed receipt** would falsify the record and break the ODR signing/verification fixtures.

## ⚠️ Behavioural change: adaptive thinking is on by default

This is the part a model-ID swap cannot express, and it produced **two distinct classes of breakage**. Opus 4.8 ran **without** thinking when the `thinking` field was omitted; Opus 5 runs **adaptive thinking**.

**Round 2 (post-review):** both reviewers returned CHANGES-REQUESTED and were right — the same hazard existed at sites the first pass missed, because the initial grep looked for `content[0].text` and missed the raw-JSON form `data["content"][0]["text"]`, and because `domain_matcher` had its budget raised but not its parse. Rather than patch eight sites, round 2 adds a shared module — **`aragora/models/compat.py`** (`first_text_block()` + `rejects_sampling_params()`/`strip_sampling_params()`, 24 regression tests in `tests/models/test_compat.py`). Sites fixed in round 2: `verification/formal.py` (×2 parses + a 512 budget), `evolution/evolver.py`, `routing/domain_matcher.py` (parse), `computer_use/claude_bridge.py` (temperature), `agents/api_agents/anthropic.py` (central temperature strip + legacy fallback parse), and the P3 `rescue_planner` comment.

### (a) `content[0].text` raises `AttributeError` — hard crash

With thinking on, the response's **first content block is a `ThinkingBlock`, not a `TextBlock`** (blocks are still emitted even though `display` defaults to `"omitted"`, just with empty text). Four raw call sites indexed blindly:

| Site | Was |
|---|---|
| `swarm/value_estimator.py:414` | `response.content[0].text.strip()` |
| `swarm/issue_upgrader.py:553` | `response.content[0].text.strip()` |
| `nomic/task_decomposer.py:1571` | `response.content[0].text if response.content else ""` |
| `scripts/nomic_live_fire.py:28` | `resp.content[0].text` |

All four now scan for the text block instead of indexing position 0, matching the pattern `api_agents/anthropic.py` already uses. `documents/folder/agent_filter.py` has the same shape but is Gemini-backed and unaffected.

**These were the "pre-existing" mypy `union-attr` warnings on those exact lines** — dormant on Opus 4.8 (which never emitted thinking blocks), live on Opus 5. Fixing them removes ~28 of the repo's standing mypy errors as a side effect.

### (b) `max_tokens` now covers thinking + response

Six call sites had budgets sized for a no-thinking model:

| Site | Before | After |
|---|---|---|
| `routing/domain_matcher.py` | **200** | 4096 |
| `swarm/value_estimator.py` | 256 | 4096 |
| `swarm/issue_upgrader.py` (direct) | 512 | 4096 |
| `swarm/issue_upgrader.py` (OpenRouter) | 512 | 4096 |
| `swarm/rescue_planner.py` | 500 | 4096 |
| `nomic/task_decomposer.py` (×2) | 1024 | 8192 |

The domain matcher was the worst: it asks for JSON in 200 tokens, so thinking would consume the whole budget, hit its `if not response.content` branch and return `None` — **silent degradation**. `max_tokens` is a ceiling, not a spend commitment, so headroom costs nothing unused. Budget-raising was chosen over `thinking: {type: "disabled"}` because several sites go through OpenRouter (which won't accept Anthropic-specific config) and the installed `anthropic==0.83.0` predates Opus 5's typed `effort`/`adaptive` values.

### (c) `temperature` is rejected with a 400 on Opus 4.7+

`swarm/rescue_planner.py` sent `"temperature": 0.1`. Non-default sampling parameters return a **400** on Opus 4.7 and later. Removed.

**No test caught any of this** — every one of these paths mocks the API, so the suites stayed green while the real calls would have failed. Found by auditing call sites against the migration guide rather than by running tests.

## Catalog fan-out (why the diff is wide)

`ENFORCED_MODELS = tuple(CATALOG)`, so adding one `ModelSpec` **automatically enforces it against every runtime mirror** via `tests/models/test_catalog.py`. Rows were added to `pdb/_PRICE_PER_MTOK`, `billing/usage`, `services/metering_models`, `billing/debate_costs` and `billing/optimizer`. `routing/provider_config` needed none — it projects from the catalog.

## Incidental, separable: a live provider reprice

Cataloguing Opus 5 requires refreshing `catalog_snapshot.json`, which is a **live capture that must not be hand-edited**. The refresh caught a real reprice:

```
moonshotai/kimi-k2.7-code   0.75/3.50  ->  0.82/3.75  per MTok
```

Reconciled across the catalog and all mirrors, plus the summed expectation in `tests/pdb/test_real_invoker.py` (4.25 → 4.57). Unrelated to the Opus bump; called out so it can be reviewed (or reverted) independently.

## Pre-existing bugs corrected rather than propagated

- `AGENTS.md` documented the `claude` CLI agent default as `claude-opus-4-8`; the registration in `cli_agents.py` pins `claude-fable-5`. Corrected to the truth instead of writing a new wrong value.
- `triage/issue_evaluator.py` `DEFAULT_PANEL` carried `0.015/0.075` per-1k cost estimates — stale **Opus 4.1-era ($15/$75)** rates. Now `0.005/0.025`.
- `review/protocol.py` named a `claude-opus-*-1m` variant; no such Opus 5 model exists (it is natively 1M), so the fabricated suffix was dropped.

## Verification

Model id, pricing, context/output limits verified against the provider model page; the OpenRouter slug `anthropic/claude-opus-5` verified against the **live** catalog rather than inferred from the 4.8 dotted spelling.

| Gate | Result |
|---|---|
| `tests/agents` + `tests/billing` + `tests/pdb` | **8634 passed** |
| gauntlet / governance / review_queue | 3060 passed |
| catalog / pins / routing / verticals / handlers | 1686 passed |
| metering / server / oracle stream | 342 passed |
| `check_canonical_metrics.py --all` | 10 pass / 0 fail |
| `tests/test_openapi_regeneration.py` | passed |
| ruff + ruff-format | clean |
| mypy | 48 errors across 9 changed files — **byte-identical to `origin/main`** |

The mypy note matters: the gate is changed-file scoped, so this PR surfaces pre-existing errors in files it touches. Verified against a baseline `origin/main` worktree — **none are introduced by this change**.

Generated artifacts were regenerated via their own generators (`export_openapi`, `generate_openapi` + the three `add_openapi_*` passes, `generate_sdk_types`, `generate_python_sdk_types`, `model_catalog_drift --refresh`), never hand-edited.

## Checklist

- [x] Code follows project style (`ruff` + `ruff format` clean via pre-commit)
- [x] All existing tests pass
- [x] Documentation updated (`AGENTS.md`, `docs/architecture/MODEL_CATALOG.md`)
- [x] Commit message follows conventional commit format
- [x] No hardcoded secrets or credentials
- [x] No new `# type: ignore` comments


scarmani:

Head: 9dc2e57 (9dc2e5728961d250c631155ca740a6a4abd69b56), committed 2026-07-27T15:20:08Z.
PR: #9588.
Model family: claude

Verdict: PASS

The migration is materially complete and internally consistent. I verified in the worktree at head 9dc2e57: no remaining `content[0].text` / `["content"][0]["text"]` parses in production code (the only hits are comments and the compat docstring); every direct-Anthropic call site that sends sampling params now routes through `strip_sampling_params` (including both streaming payload builders in `anthropic.py`, and the OpenRouter path correctly keys off `payload["model"]` so quota-fallback re-sends as Opus 5 are stripped); `first_text_block` correctly excludes thinking/tool blocks in both SDK-object and raw-JSON shapes, and opus-4-5/sonnet-4-6/haiku ids correctly do NOT match the modern-Claude regex (so temperature is never silently dropped for models that accept it). Pricing is consistent at $5/$25 across all five tables (catalog, snapshot, billing/usage, metering, real_invoker, model_selector), `model_pins.py` role pins are re-pointed, and empty-text-means-failure handling was added where "" would previously have masked an error (research_phase falls back to OpenRouter, formal.py returns None/low-confidence, evolver warns instead of returning an empty refinement). Targeted suites pass: 653 passed, 2 skipped across the compat, catalog, anthropic/openrouter agent, vibeproxy, real_invoker, swarm-config, verticals, cli_agents, and computer_use test files.

- [P3] `aragora/agents/registry.py:100` — the docstring example registers the `claude` CLI agent with `default_model="claude-opus-5"`, but the actual registration is `claude-fable-5` (`aragora/agents/cli_agents.py:829`, and AGENTS.md documents Fable 5). The example was stale before this PR and is still stale after touching it.
- [P3] `aragora/models/catalog.py` (claude-opus-5 spec) — `soak_until=None` waives the 14-day availability rule for a day-0 model, granting it immediate merge-authority-evidence and routing eligibility, while `model_pins.py` simultaneously holds the reviewer pin on gpt-5.5 citing that same rule (#9075). The waiver is explicitly operator-directed and documented in place with a one-line reinstatement path, so this is an observation for the record, not a blocker — but quorum reviews will now be adjudicated partly by a model with zero soak history.
- [P3] `aragora/nomic/task_decomposer.py:1573`, `aragora/swarm/issue_upgrader.py:556`, `scripts/nomic_live_fire.py:27` — hand-rolled inline `next(... if b.type == "text")` scans duplicate `aragora.models.compat.first_text_block`; the sweep's other ~10 sites use the shared, tested helper. Behaviorally equivalent today, but future block-shape fixes will only land in the helper.
- [P3] `aragora/server/question_classifier.py:303` — still pins `claude-opus-4-7` with a hand-rolled `hasattr(block, "text")` scan; safe against thinking blocks and sends no sampling params, but it's now two generations behind the frontier pin and bypasses both `model_pins` and the compat helper this PR established.

dogfood: yes

scarmani:

Head: 9dc2e57 (9dc2e5728961d250c631155ca740a6a4abd69b56), committed 2026-07-27T15:20:08Z.
PR: #9588.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

PR: #9588
Exact head: 9dc2e5728961d250c631155ca740a6a4abd69b56
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #9588 is non-draft and live exact-head checks/merge-packet remain otherwise green.

PR #9560, opened 2026-07-23
fix(security): raise live PostCSS floor

scarmani:
## Summary
- raise the `aragora/live` direct PostCSS development floor and override from `^8.5.10` to `^8.5.22`
- regenerate the lockfile so production resolves PostCSS 8.5.22 and nanoid 3.3.16
- remediate GHSA-6g55-p6wh-862q in the live frontend audited by Security Gate

## Validation
- `npm ci --ignore-scripts`
- `npm audit --omit=dev --audit-level=high` (0 vulnerabilities)
- `npm run lint`
- `npm run build` (228 routes)
- `npm test -- --runInBand` (257 suites, 4025 tests passed)
- `python3 scripts/report_code_quality.py --check`
- `bash scripts/automation_pr_preflight.sh origin/main HEAD`
- `git diff --check origin/main...HEAD`

This is intentionally separate from #9505 and #9545. It fixes the deterministic live-frontend security failure before either PR is restacked.

PR #9505, opened 2026-07-22
chore(runtime): align frontend containers on Node 24 LTS

scarmani:
## Summary

- align every approved frontend container surface on `node:24.18-alpine`
- retain the merged dependency remediation from #9484: Next.js `16.2.11` and sharp `0.35.3`
- add a regression assertion covering every frontend Node image pin
- leave package manifests, lockfiles, workflows, and protected files unchanged relative to current main

## Base reconciliation (2026-07-23)

PR #9505 was reconciled with main at `6015cd8fd9343e51cbfcf31db072b6574666c9d4`. Main later advanced to `db4451d007fee27629d24a62402dfa27495e7c10` in the unrelated `aragora/swarm/boss_worker_lifecycle.py`; the PR remains mergeable and its effective diff remains exactly four files:

| Surface | Current main | This PR |
| --- | --- | --- |
| `aragora/live/Dockerfile` | Node 20.11 | Node 24.18 |
| `deploy/Dockerfile.frontend` | Node 20.11 | Node 24.18 |
| `docker-compose.dev.yml` | Node 20.19 | Node 24.18 |
| `tests/ci/test_docker_compose_dev.py` | asserts Node 20.19 | asserts Node 24.18 and cross-surface consistency |

All seven frontend image references resolve to `node:24.18-alpine`.

## Verification

- `python -m pytest tests/ci/test_docker_compose_dev.py -v` — 3 passed
- `docker compose -f docker-compose.dev.yml config --quiet`
- `docker manifest inspect node:24.18-alpine`
- TypeScript SDK locked install and build under exact Node `v24.18.0`
- live dashboard locked install and `npm run build:local` under exact Node `v24.18.0` — Next `16.2.11`, typecheck passed, 228/228 routes generated, standalone output built
- sharp `0.35.3` runtime smoke under exact Node `v24.18.0` — PNG encoded successfully
- `npm ls next sharp --all` — Next `16.2.11`, sharp `0.35.3`
- `bash scripts/automation_pr_preflight.sh origin/main HEAD`
- push-time repository hooks — passed

## Risk and gate status

This is a major runtime-image alignment and remains Tier 4. Rollback is the runtime-pin commit; no workflow or package rollback is required. All evidence and checks from the previous head are stale after reconciliation. Fresh CI is required on exact head `016717dd465e70d8d0fa4ef54f68f6ffd13b4b93`; quorum evidence, settlement, and merge are intentionally outside this update lane.

## Inherited audit disclosure

[...]

PR #9589, opened 2026-07-24
fix(deps): clear dev-tier npm advisories in aragora/live

scarmani:
## Summary

Clears the four dev-only `npm audit` advisories in `aragora/live` that remained after
the PostCSS floor landed in #9560, and removes a pre-existing `brace-expansion` override
that was silently breaking `minimatch@10` at runtime.

These advisories were **non-blocking** — the CI gate runs `npm audit --omit=dev --audit-level=high`
and all four are dev-tier — so this is hygiene plus one real latent-bug fix, not an incident.

## Advisories cleared

| Package | Advisory | Sev | Fix |
|---|---|---|---|
| `@babel/core` | [GHSA-4x5r-pxfx-6jf8](https://github.com/advisories/GHSA-4x5r-pxfx-6jf8) — arbitrary file read via `sourceMappingURL` | low | override `^7.29.7` (stays in 7.x) |
| `js-yaml` | [GHSA-h67p-54hq-rp68](https://github.com/advisories/GHSA-h67p-54hq-rp68), [GHSA-52cp-r559-cp3m](https://github.com/advisories/GHSA-52cp-r559-cp3m) — merge-key quadratic DoS | high | override `^4.3.0` + nested `^3.15.0` |
| `@redocly/openapi-core` | flagged only via its `js-yaml` dep | moderate | resolved by the above |
| `brace-expansion` | [GHSA-3jxr-9vmj-r5cp](https://github.com/advisories/GHSA-3jxr-9vmj-r5cp) — exponential-time expansion DoS | high | override removed → natural resolution |

## The `brace-expansion` finding

The existing `"brace-expansion": "^2.0.3"` override (added in `f60d5d4075`) hoisted a single
2.1.0 across all four `minimatch` copies in this tree. But those four consume `brace-expansion`
through **three mutually incompatible export shapes**:

| Consumer | Call site | Needs |
|---|---|---|
| `minimatch@10.2.4` (typescript-estree) | `(0, brace_expansion_1.expand)(p, {max})` | v5 **named** export |
| `minimatch@9.0.9` (glob/jest) | `__importDefault(require(...)).default(p)` | v1/v2 callable default |
| `minimatch@5.1.9` (redocly), `minimatch@3.1.5` (eslintrc) | `var expand = require(...); expand(p)` | v1/v2 callable default |

`brace-expansion@2.x` is `module.exports = expand` with **no `.expand` property**, so on `main`
today `minimatch@10` throws on every pattern parse. Isolated repro (`minimatch@10.2.4` +
`overrides: {"brace-expansion": "2.1.0"}`, i.e. main's exact config):

```
minimatch@10 + brace-expansion@2.1.0 -> BROKEN: (0 , brace_expansion_1.expand) is not a function
```

Raising the override to `^5.0.8` inverts the breakage (v5 fixes `minimatch@10` and breaks the
other three). No single pinned version works, so the override is removed and each subtree
resolves its declared-compatible version — `1.1.16 / 2.1.2 / 2.1.2 / 5.0.8`. All four verified:

```
brace-expansion = 5.0.8
  PASS  minimatch@3 (eslintrc)     PASS  minimatch@5 (redocly)
  PASS  minimatch@9 (glob/jest)    PASS  minimatch@10 (ts-eslint)
```

## Reviewed design tradeoffs

- **Global `brace-expansion` bump to `^5.0.8` (rejected).** Yields `npm audit` → 0 vulnerabilities,
  but the tree is *broken at runtime* in 3 of 4 minimatch copies. `npm audit` checks versions, not
  export-shape compatibility, so a clean audit here would have been actively misleading. Correctness
  over a green number.
- **Global `^2.1.2` (rejected).** Clears GHSA-3jxr-9vmj-r5cp but leaves `minimatch@10` broken and
  still carries the residual advisory. Strictly worse than removing the override.
- **Overriding `minimatch` itself to 10.x everywhere (rejected).** Would let all consumers use
  `brace-expansion@5` and fully clear the residual, but forces a major API change into `glob@10`
  (declares `minimatch ^9`), eslintrc, and redocly. Disproportionate risk for a dev-only DoS.
- **Reported-count optics.** Full `npm audit` goes 4 → 29, but that is **1 root cause + 28 cascade**
  entries; the count rises only because the residual now sits in nested subtrees whose parents get
  attributed. The blocking gate is unchanged at 0.
- **`js-yaml` nested override.** `@istanbuljs/load-nyc-config` uses the 3.x `safeLoad` API removed in
  4.x, so a flat `^4.3.0` would break jest coverage config loading; 3.15.0 is the 3.x fix line.

## Residual (documented, deliberate)

[GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg) — `brace-expansion`
unbounded expansion → OOM (high) — is fixed **only** in 5.0.8, which the `minimatch@3/5/9` subtrees
cannot accept. It is dev-only and does not reach the blocking gate. It clears once `glob`,
`@eslint/eslintrc`, and `@redocly/openapi-core` ship minimatch versions built against v5.

## Verification

CI runs Node 20; `brace-expansion@5.0.8` declares `engines: "20 || >=22"` — satisfied.

```
npm ci --ignore-scripts && npm audit --omit=dev --audit-level=high   -> found 0 vulnerabilities, exit 0
npm run lint      -> clean (--max-warnings 0)
npm test          -> 257 suites / 4025 tests passed, 1 skipped
npm run build:ci  -> success (full route manifest)
```

Lockfile refreshed and re-verified reproducible from a clean `npm ci`.

## Scope

Touches only `aragora/live/package.json` and `aragora/live/package-lock.json`.
Unrelated to #9505 (frozen at `016717dd`), which is not modified.

### DIFF
--- a/aragora/live/package-lock.json
+++ b/aragora/live/package-lock.json
@@ -165,13 +165,13 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.29.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.0.tgz",
-      "integrity": "sha512-9NhCeYjq9+3uxgdtp20LSiJXJvN0FeCtNGpJxuMFZ1Kv3cWUNb6DOhJwUvcVCzKGR66cw4njwM6hrJLqgOwbcw==",
+      "version": "7.29.7",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.7.tgz",
+      "integrity": "sha512-Aup7aUOfpbAUg2ROOJN6Iw5f9DMBlzu0mIkm/malLQFN/YQgO48wCj0Kxa3sEHJvPVFg7siR+qRInwXd2qhQKw==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
-        "@babel/helper-validator-identifier": "^7.28.5",
+        "@babel/helper-validator-identifier": "^7.29.7",
         "js-tokens": "^4.0.0",
         "picocolors": "^1.1.1"
       },
@@ -180,9 +180,9 @@
       }
     },
     "node_modules/@babel/compat-data": {
-      "version": "7.29.0",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.29.0.tgz",
-      "integrity": "sha512-T1NCJqT/j9+cn8fvkt7jtwbLBfLC/1y1c7NtCeXFRgzGTsafi68MRv8yzkYSapBnFA6L3U2VSc02ciDzoAJhJg==",
+      "version": "7.29.7",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.29.7.tgz",
+      "integrity": "sha512-locTkQyKvwIEgBzVrn8693ebc97F2U8ZHjbXwDXJ5Fn2TCpNwTlKcaKLkdHop5c/icOFE7qt7Q9JC5hnKNa6Gg==",
       "dev": true,
       "license": "MIT",
       "engines": {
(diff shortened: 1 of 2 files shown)


scarmani:

Head: 5edf60e (5edf60e94d0306a0564eaaf4c028f5e3e77eeb4d), committed 2026-07-24T23:42:16Z.
PR: #9589.
Model family: claude

Verdict: PASS

- [P3] `aragora/live/package-lock.json` (top-level `brace-expansion` → 5.0.8, `balanced-match` → 4.0.4): the newly hoisted copies declare `engines.node: "20 || >=22"` / `"18 || 20 || >=22"`. On Node 18 (or with `--engine-strict` anywhere in CI/Docker) install will warn or fail; the Node 24 image migration (#9505) isn't merged at this head, so confirm every environment that runs `npm ci` in `aragora/live` is already on Node ≥20.
- [P3] `aragora/live/package.json` (overrides): the global `"js-yaml": "^4.3.0"` override force-upgrades every transitive js-yaml to v4 (which removed the v3 `safeLoad` API) except the single `@istanbuljs/load-nyc-config` carve-out. The lock diff shows load-nyc-config as the only v3 consumer today, so this is correct now, but any future dependency pulling js-yaml@3 will be silently broken at runtime rather than at install time — worth a comment or a periodic `npm ls js-yaml` check.
- [P3] Dropping the global `brace-expansion` override restores per-major nesting (1.1.16 under minimatch@3, 2.1.2 under the v5/v9 minimatch copies, 5.0.8 hoisted for minimatch@10) — this is the right cure for the export-shape breakage, and all pinned versions are at or above the ReDoS-patched floors (≥1.1.12 / ≥2.0.2). Note the v1-line security floor is now held only by the lockfile, no longer by an override, and the long-unmaintained `concat-map@0.0.1` returns as a required dep of brace-expansion@1 (expected, no advisory).
- [P3] Since the previous state proved "npm audit green ≠ working tree," attach runtime evidence to the PR (a passing `next build` + test run from this head), not just a clean `npm audit` — the lockfile shape looks consistent, but only an actual build exercises all four minimatch copies.

dogfood: yes

scarmani:

Head: 5edf60e (5edf60e94d0306a0564eaaf4c028f5e3e77eeb4d), committed 2026-07-24T23:42:16Z.
PR: #9589.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9593, opened 2026-07-25
fix(settle): surface reviewer findings so a dissent is diagnosable in one run

scarmani:
## The gap

When a reviewer dissented, `settle_pr` reported only *countability codes*:

```
problems=['blocking_or_negative_verdict', 'no_counted_model_family', 'no_counted_model_reviewer']
```

Those say a reviewer dissented but never **why**. The reviewer's actual text was
fetched and then discarded by `summarize_collect`, so diagnosing any dissent
required a second full `collect_quorum_evidence` run — ~15 minutes of real model
reviewers — purely to re-read something already retrieved.

Hit twice while settling #9571.

## Why it mattered more than the delay

With no visible reason, a dissent is indistinguishable from reviewer flake — and
these reviewers *are* non-deterministic (on #9571 openai returned opposite
verdicts on an identical commit 20 minutes apart). The cheap response is to
re-run until a PASS appears. On #9571 that "flake" was a genuine **P1 in
merge-authorization code**: a stale rollup row could read as green. It was
caught only by paying the 15 minutes to read the body.

A gate whose dissents are expensive to read is a gate people learn to re-roll.

## Change

- **`extract_finding_lines()`** — reports every real `[Pn]` line at any severity.
  Reuses the gate's own `_semantic_review_lines` / `_strip_decoration` /
  `_NO_FINDING_HEADS` handling, so a fenced or blockquoted `[P2]` example that
  the gate ignores is never reported as a live finding, and `[P2] None` stays a
  non-finding. Severity is deliberately unfiltered — `highest_blocking_severity`
  is `[P0]`/`[P1]`-only *by design*, and a `[P2]`-only dissent is exactly the
  common case a human needs to read.
- **`summarize_collect`** carries `findings` + the verbatim `body` per item.
- **`settle_pr`** prints findings inline under each reviewer.
  `--show-review-bodies` adds the full text (opt-in; bodies are several KB each,
  and the default JSON is often captured into logs).

## Safety

The gate module change is **45 insertions, 0 deletions** — purely additive. The
diagnostic markers (`_ANY_PRIORITY_MARKER`) are separate constants from the gate
markers (`_PRIORITY_MARKER`), so widening what is *reported* cannot widen what is
*blocking*. No existing gate logic is touched.

## Testing

347 settle/quorum tests and 103 gate-parser tests pass. New tests use the
**verbatim** claude and openai reviews from #9571, covering plain `[P1]` and
markdown `**[P2]**` forms, all-severity reporting, fenced/no-finding exclusion,
missing-body tolerance, and marker normalisation.

Before / after on that real payload:

```
- openai   verdict=changes_requested  problems=['blocking_or_negative_verdict', ...]
+ openai   verdict=changes_requested  problems=['blocking_or_negative_verdict', ...]
+     [P1] `auto_merge_green.py:_rollup_recency` - Recency sorts by `completedAt` before …
+     [P2] `auto_merge_green.py:_reduce_rollup_states` - Equal/unknown recency only prefers …
```

## Reviewed design tradeoffs

- **Findings inline vs. full bodies by default:** bodies run several KB per
  reviewer and `--json` output gets captured into logs, so verbatim text is
  opt-in while the actionable `[Pn]` lines are always shown.
- **Reusing the gate's line splitter vs. a fresh regex:** a separate parser would
  drift and could report findings the gate ignores (fenced examples), which is
  worse than no diagnostic. Reuse guarantees the view matches the gate's own
  reading.
- **Reporting all severities vs. mirroring the blocking filter:** mirroring would
  hide `[P2]`/`[P3]`-only dissents — the exact case that sends an operator back
  for a second run, so it would reproduce the gap this closes.

Found while settling #9571.

### DIFF
--- a/aragora/cli/commands/review_queue_comment_verdicts.py
+++ b/aragora/cli/commands/review_queue_comment_verdicts.py
@@ -167,6 +167,12 @@ _DEFAULT_BLOCKING_MARKER_STRIP = re.compile(r"^(?:\*\*)?\[(?:p0|p1|p2)\](?:\*\*)
 _PRIORITY_MARKER = re.compile(r"^(?:\*\*)?\[(?P<sev>p0|p1)\](?:\*\*)?(?:\s|$|[:.;—–-])", re.I)
 _PRIORITY_MARKER_STRIP = re.compile(r"^(?:\*\*)?\[(?:p0|p1)\](?:\*\*)?\s*", re.I)
 
+# Any-severity variants, used ONLY by the diagnostic :func:`extract_finding_lines`.
+# Kept separate from the gate markers above so widening the diagnostic view can
+# never widen what counts as blocking.
+_ANY_PRIORITY_MARKER = re.compile(r"^(?:\*\*)?\[(?P<sev>p\d)\](?:\*\*)?(?:\s|$|[:.;—–-])", re.I)
+_ANY_PRIORITY_MARKER_STRIP = re.compile(r"^(?:\*\*)?\[p\d\](?:\*\*)?\s*", re.I)
+
 
 def _strip_decoration(text: str) -> str:
     return re.sub(r"^(?:[#>\-*+\s]+|\d+[.)]\s+)+", "", text.strip())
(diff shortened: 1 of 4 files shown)


scarmani:

Head: e560c1e (e560c1e5a76565a7124c4685b4b390c05e57c39e), committed 2026-07-25T00:03:08Z.
PR: #9593.
Model family: claude

Verdict: PASS

- [P3] `aragora/swarm/settle_plan.py:24` — layering inversion: a previously stdlib-only pure library module (`swarm/settle_plan.py`, advertised as "performs no I/O, fully unit-testable") now imports from `aragora.cli.commands.*`. It works today (both package `__init__`s are lazy and the verdict module imports only `re` — verified importable), but library←CLI coupling invites a future cycle when someone adds a top-level import to `cli/__init__.py`; the shared parsing helpers belong in a non-CLI module.
- [P3] `scripts/settle_pr.py:149-151` — with `--show-review-bodies` set and zero parsed findings on a non-counting item, the hint "(no [Pn] finding lines parsed — see --show-review-bodies)" still prints immediately above the full body it tells you to go get. Cosmetic; gate the hint on `not args.show_review_bodies` (or just body absence).
- [P3] `scripts/settle_pr.py:145-153` — reviewer bodies/finding lines are echoed verbatim to the operator's terminal; a hostile PR review comment can embed ANSI/control sequences. Findings lines print by default, bodies only opt-in, and this matches existing verdict/problems printing, so observation only — consider stripping C0/escape chars before printing.

Verified, not merely assumed: the `body` pop at `scripts/settle_pr.py:265` executes before both the `--json` and `_render_human` paths (render is at line 269), so bodies never leak into default output; `summarize_collect` has no other callers, so no artifact writer starts persisting bodies; and I ran the new extraction logic against the main-branch helpers it reuses — all six test bodies plus adversarial edges (`[P2]:` bare colon, `[P1] None of the inputs…` kept, `[P2] None` dropped, fenced/bolded/lowercase markers) behave exactly as the tests assert and stay consistent with `_priority_finding_severity`'s head-normalization. The blocking gates (`_PRIORITY_MARKER`, `_DEFAULT_BLOCKING_MARKER`) are untouched, and the widened `p\d` marker is confined to the diagnostic function.

dogfood: yes

scarmani:

Head: e560c1e (e560c1e5a76565a7124c4685b4b390c05e57c39e), committed 2026-07-25T00:03:08Z.
PR: #9593.
Model family: openai

Verdict: PASS
- [P3] `aragora/cli/commands/review_queue_comment_verdicts.py:381` `extract_finding_lines` misclassifies explicit non-findings written with punctuation immediately after the marker, e.g. `[P3]: N/A` or `[P2] - None`, because `head` is computed before stripping the marker separator. This is diagnostic-only, so it can produce noisy surfaced findings but does not widen merge blocking.

dogfood: yes

scarmani:

PR: #9593
Exact head: e560c1e5a76565a7124c4685b4b390c05e57c39e
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #9593 is non-draft and live exact-head checks/merge-packet remain otherwise green.

PR #9594, opened 2026-07-25
fix(merge): treat untimestamped rollup rows as unrankable, not oldest

scarmani:
## Summary

Residual fail-open from #9571, flagged in that PR's own review and confirmed
against the merged code.

## The defect

`_rollup_recency` returns `("", "")` for a row with no timestamps. Tuple
comparison then ranks that row as merely the **oldest**, rather than as
*unrankable* — so the fail-closed tie-break, which only fires on exactly-equal
recency, never engaged against a timestamped row:

```
[SUCCESS started=21:00 completed=21:05, FAILURE (no stamps)]  ->  SUCCESS
```

Verified against merged `main` before this change.

This is not hypothetical. Check *runs* carry `startedAt`/`completedAt`, but
commit *statuses* (the `context`/`state` shape — e.g. `aragora/human-settlement`)
carry **neither**, so any check name appearing in both shapes hits it.
`decide_auto_merge` reads this map to authorise an unattended merge, and for
required checks other than quorum it is the only gate.

It also made #9571's docstring claim — "when recency cannot be established the
reduction is fail-closed" — an overclaim, since that only held for exactly-equal
tuples. The docstring is corrected here too.

## Fix

Rows are compared only when **both** carry timestamps *and* those timestamps
differ. Otherwise the pair is not comparable and any non-success state wins.

## Testing

84 tests pass. Both new bug-detecting tests **fail against merged `main`** and
pass here; each asserts in both input orders so none can pass by accident of
ordering. `test_rankable_rerun_still_wins_after_unrankable_guard` pins the
legitimate rerun-to-green case that this must not regress.

## Reviewed design tradeoffs

- **Unrankable vs. "oldest":** treating a missing timestamp as oldest is what
  produced the fail-open. A row with no stamps cannot be proven newer *or*
  older, so the only safe reading is "cannot order these".
- **Requiring both stamps vs. inferring from one:** a partial stamp could be
  used to guess an ordering, but guessing is what this whole reduction exists to
  eliminate in the merge-authorisation path.
- **Cost:** strictly more conservative — it can add an unnecessary skip, never an
  unauthorised merge. A skipped PR is picked up on the next pass.

Follow-up to #9571. Found via the reviewer-findings diagnostic in #9593.

### DIFF
--- a/aragora/swarm/auto_merge_green.py
+++ b/aragora/swarm/auto_merge_green.py
@@ -228,7 +228,15 @@ def _reduce_rollup_states(rollup: Any) -> dict[str, str]:
     States are upper-cased so this reduction cannot be bypassed by casing,
     matching the normalisation the cheap prefilter applies.
     """
-    best: dict[str, tuple[tuple[str, str], str]] = {}
+    # Three disjoint pools, so no row can be displaced out of the reduction by a
+    # row it is not comparable to. Folding everything into one slot is what let a
+    # veto vanish: an unrankable PENDING landing on a FAILURE incumbent matched
+    # no branch and was dropped, after which a later SUCCESS outranked the
+    # FAILURE and the name read green while the PENDING was still deciding.
+    rankable: dict[str, tuple[tuple[str, str], str]] = {}
+    veto: dict[str, str] = {}
+    unrankable_success: set[str] = set()
+
     for item in rollup or []:
         if not isinstance(item, dict):
             continue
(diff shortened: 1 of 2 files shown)


scarmani:

Head: 3207d8d (3207d8d99fb80bcd27debbb65282747b1975c17f), committed 2026-07-27T15:09:54Z.
PR: #9594.
Model family: claude

Verdict: PASS

Review grounded against the full module at head 3207d8d (not just the diff): I traced `_reduce_rollup_states` → `context_from_gh` → `decide_auto_merge` and both production callers (`scripts/auto_merge_quorum_green.py`, `scripts/merge_executor.py`), ran the touched test file at head (57/57 pass), and ran an adversarial permutation probe of my own against the edge cases the shipped tests don't cover.

The core redesign is sound and strictly fail-closed relative to the parent: the three-pool split genuinely closes the dropped-veto hole (an unrankable non-success can no longer be displaced out of the reduction), the precedence rule never opens a new SUCCESS pathway (precedence 0 can never win a tie or override a veto), and the final veto-application guard preserves the `_FAILING_CHECK_STATES` blocker in `decide_auto_merge` that the PR title targets. A lone unrankable commit-status SUCCESS resolving green is correct, since GitHub keeps only the latest status per context. The cheap prefilter's any-success read is over-inclusive only in the safe direction (packet + fixed reduction remain the gate).

- [P3] `aragora/swarm/auto_merge_green.py:327` — the `context_from_gh` docstring still says rows are "reduced by **recency** (``completedAt`` then ``startedAt``)". That is the reverse of the actual start-primary ordering, directly contradicts this PR's own `_rollup_is_rankable` docstring and `test_completed_at_only_row_is_unrankable` rationale, and omits the new veto/precedence semantics. Pre-existing text, but it documents the exact mechanism this PR reworks — worth fixing in the same change.
- [P3] `aragora/swarm/auto_merge_green.py:204` — `completedAt` as the secondary sort key re-admits the "completed stale SUCCESS outranks a still-deciding re-run" hazard when two runs share a `startedAt` second: probe shows `("T","")` in-flight PENDING loses to `("T","T2")` completed SUCCESS in every order. GitHub stamps are second-granularity, so this is a real (if narrow) race the equal-recency precedence branch doesn't reach because the tuples aren't equal. Pre-existing behavior, unchanged by this PR; dropping the secondary key (making same-second a precedence tie) would close it.
- [P3] `aragora/swarm/auto_merge_green.py:277` — the comment claims the surviving veto representative "does not depend on row order", but at *equal* precedence (PENDING vs QUEUED; likewise FAILURE vs CANCELLED in the rankable tie branch, probe: `{'CANCELLED','FAILURE'}`) it is still first-seen. The merge *decision* is order-independent — the states are decision-equivalent in `decide_auto_merge` — only the blocker message text varies, so this is a comment overstatement, not a defect.

dogfood: yes

scarmani:

Head: 3207d8d (3207d8d99fb80bcd27debbb65282747b1975c17f), committed 2026-07-27T15:09:54Z.
PR: #9594.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

Head: 3207d8d (3207d8d99fb80bcd27debbb65282747b1975c17f), committed 2026-07-27T15:09:54Z.
PR: #9594.
Model family: claude

Verdict: PASS

- [P3] `aragora/swarm/auto_merge_green.py:194` (`_rollup_recency` / equal-recency tie-break) — the precedence tie-break only engages on *exactly equal* recency tuples. Two rows sharing the same `startedAt` where one has completed (`(T, c)`) and one is in flight (`(T, "")`) compare unequal, so a completed stale `SUCCESS` still outranks a same-second in-flight rerun and reads green while it's deciding. Needs identical start stamps to the second, so it's a narrow window, and it's pre-existing behavior rather than a regression of this diff — but it's the one remaining fail-open seam in the ordering and worth a follow-up (e.g. treat equal-`startedAt` as a tie for precedence purposes).
- [P3] `aragora/swarm/auto_merge_green.py:~290` (veto pool comment) — the comment claims the surviving veto representative "does not depend on row order," but among equal-precedence states (e.g. `FAILURE` vs `TIMED_OUT`, both 2; `PENDING` vs `QUEUED`, both 1) first-seen still wins, so the surviving *state string* is order-dependent. Decision-equivalent under `decide_auto_merge` (set membership / `!= SUCCESS`), so impact is limited to nondeterministic blocker text, but the comment overstates the guarantee.
- [P3] design note, `aragora/swarm/auto_merge_green.py:~300` — an unrankable `SUCCESS` can never clear a name that has *any* rankable row, so a check whose re-run reports green only via a commit status (dual-transport shape) stays permanently blocked by a lingering stale rankable `FAILURE`. Fail-closed and consistent with this module's chosen stance, but it's a liveness deadlock shape to keep in mind operationally.

The core redesign is sound: the three-pool partition is decision-level order-independent (verified against the permutation tests' properties), the veto/precedence interaction correctly refuses to let a transient mask a terminal failure (`decide_auto_merge` blocks non-required checks only on `_FAILING_CHECK_STATES`, confirmed at `aragora/swarm/auto_merge_green.py:49,159`), unrankable `SUCCESS` is correctly confined to `setdefault`, and the `context`/`state` shape is genuinely reachable (`name = item.get("name") or item.get("context")` at line 263). Test coverage over `itertools.permutations` is the right property for this bug class.

dogfood: yes

scarmani:

Head: 3207d8d (3207d8d99fb80bcd27debbb65282747b1975c17f), committed 2026-07-27T15:09:54Z.
PR: #9594.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9484, opened 2026-07-22
fix(deps): remediate inherited sharp advisory in aragora/live

scarmani:
## Summary

Remediates the inherited high-severity `sharp <0.35.0` advisory in `aragora/live` and the subsequently reported high-severity Next.js advisories while preserving existing application behavior.

Merged from exact head `5db1c124f230d2cef9c4bb9b3ac1ea20df1220a7` after the operator-approved Next.js scope expansion, exact-head validation, independent model review, and Tier-1 quorum.

## Product changes

- Adds npm override `sharp: ^0.35.3` and regenerates the lockfile.
- Uses the operator-approved exact 29-entry sharp closure: 27 sharp/@img entries plus `@emnapi/runtime` 1.8.1→1.11.2 and the nested sharp `semver` 7.7.4→7.8.5 update.
- Updates Next.js from 16.2.9 to the security-fixed 16.2.11 patch, with the root dependency declaration and 10 Next-family lock entries updated.
- Aligns all three `aragora/live/Dockerfile` stages from `node:18.19-alpine` to `node:20.11-alpine`, satisfying the Node ≥20.9 engine floor shared by Next 16.2.11 and sharp 0.35.3.
- Reconciles the durable plan and learnings with both approved scope expansions.
- Does not modify workflows, protected files, SDK packages, or application source.

## Validation

- `npm audit --omit=dev --audit-level=high --json`: pass, zero vulnerabilities.
- `npm ls next sharp --all`: `next@16.2.11` with `sharp@0.35.3`, valid tree.
- Local lint, Jest, production build under Node 20.11.0/npm 10.2.4, and sharp PNG runtime smoke: pass.
- Exact-head CI: required checks and the full suite completed successfully before merge, including npm/Python security scans, build, type checking, three frontend E2E shards, integration gates, baseline determinism, and final quality gates.
- Exact-head independent review: Claude PASS with two non-blocking P3 follow-ups; OpenAI PASS with no findings.
- Tier-1 merge quorum: satisfied on the exact head before merge.

## Residual risk and follow-up

- sharp 0.35.x remains outside Next 16.2.11 declared `^0.34.5` optional range and is an upstream breaking line. Remove the override when Next.js declares support for sharp 0.35.x; verify image optimization during production rollout.
- The pre-existing `docker-compose.dev.yml` Node 18 runtime mismatch was correctly kept outside this PR. It is covered by separate PR #9505, which must be refreshed after this merge because it overlaps `aragora/live/Dockerfile`.
[...]

PR #9597, opened 2026-07-25
fix(docker): delete unbuildable live/Dockerfile, fix frontend healthcheck, gate in CI

scarmani:
## Summary

`aragora/live/Dockerfile` could not build and had been broken on `main` indefinitely. Nothing built it — `docker.yml` only builds images on tags and nightly, and `deploy-frontend.yml`'s docker jobs are gated behind a manual `target=docker` dispatch input — so the breakage was invisible.

This deletes it as a duplicate of the working `deploy/Dockerfile.frontend`, fixes a real healthcheck defect that surfaced while validating the replacement, and adds a PR-time CI gate so this class of breakage cannot recur silently.

Empirically verified with Docker 29.6.2. Pre-existing on `main`; unrelated to any in-flight PR.

## Why delete rather than repair

Three independent defects, not one:

1. `aragora/live/package.json` declares `"@aragora/sdk": "file:../../sdk/typescript"`, which resolves **outside** an `aragora/live`-scoped build context. npm cannot reify the tree and reports a misleading `EUSAGE` — "The \`npm ci\` command can only install with an existing package-lock.json" — even though `/app/package-lock.json` is copied successfully (522,753 bytes, `lockfileVersion=3`, 1029 entries).
2. The `deps` stage installed production-only, but `next build` needs devDependencies: `postcss.config.js` requires `@tailwindcss/postcss` and `next.config.js` requires `@next/bundle-analyzer`. The builder stage would fail even with the SDK present.
3. `npm run build` resolves to `build:runtime`, which never sets `NEXT_OUTPUT=standalone`, so `.next/standalone` is never emitted — and the runner stage copies exactly that.

Repairing means reproducing `deploy/Dockerfile.frontend` line for line. That file is already canonical (both compose stacks plus three build sites in `docker.yml`) and builds green nightly. Maintaining two definitions already cost us once, in the sharp remediation (#9484).

`deploy-frontend.yml`'s `deploy-docker` job is repointed at `deploy/Dockerfile.frontend` with root context. `deploy/README.md`'s build command is corrected.

Note on the deprecated `npm ci --only=production`: it disappears with the file. `deploy/Dockerfile.frontend` uses `npm install --legacy-peer-deps` by necessity — the `sed` rewrite of the SDK path desyncs `package-lock.json`, so `npm ci` in any form would reject it.

## Real defect found: the frontend image was permanently unhealthy

Writing the smoke test surfaced this in the *surviving* Dockerfile:

- Docker sets `HOSTNAME` to the container ID, and Next.js standalone `server.js` binds to `$HOSTNAME`. The server bound to the container's eth0 address (`172.17.0.3`) instead of `0.0.0.0`. The deleted Dockerfile had `ENV HOSTNAME="0.0.0.0"`; this one did not.
- After fixing that, still unhealthy: busybox wget resolves `localhost` to `::1` first while the server binds IPv4 only, so the probe is refused. `http://127.0.0.1:3000/` succeeds.

Both fixed. The identical `localhost` probe is corrected in `deploy/docker-compose.yml` and `deploy/demo/docker-compose.yml`, where the frontend service sits next to a `service_healthy` gate.

## Verification

| Check | Before | After |
|---|---|---|
| `docker build --target deps -f aragora/live/Dockerfile aragora/live` | `EUSAGE` failure | file deleted |
| `docker build -f deploy/Dockerfile.frontend .` | exit 0 | exit 0 (112 MB) |
| `docker inspect -f '{{.State.Health.Status}}'` | `unhealthy` | `healthy` |

Rebuilt no-cache on the rebased head after #9589's override changes.

## CI gate

New `.github/workflows/frontend-image-build.yml`, path-filtered to `aragora/live/**`, `sdk/typescript/**`, and `deploy/Dockerfile.frontend`. It builds the image, asserts no `aragora/live`-scoped Dockerfile reappears, and asserts the container's own HEALTHCHECK reaches `healthy` — a plain "does it build" check would not have caught the bind-address bug. It does **not** push; publishing stays with `docker.yml` and `deploy-frontend.yml`. Names are run-scoped with no published port, since the `aragora` runners are shared.

This PR triggers that check on itself.

## Reviewed design tradeoffs

- **Delete vs. repair** — repair was rejected because a correct repair is byte-equivalent to the existing working file; two definitions is the condition that produced this bug.
- **New workflow vs. adding `pull_request` to `docker.yml`** — `docker.yml` builds backend + frontend + operator multi-arch; re-adding a PR trigger would rebuild all three on every PR and restore noise that #613 deliberately removed. A path-filtered frontend-only job is the minimum that closes the gap.
- **Health-status assertion vs. an external HTTP probe** — probing from the host would have passed against the broken image, since Docker port-forwards to the container IP. Asserting the image's own HEALTHCHECK is what makes the regression visible.
- **Scope** — `deploy-docker` and `docker.yml` publish frontend images under two different names (`aragora-frontend` vs `aragora/frontend`). Consolidating those is out of scope here.

## Protected surface

Touches `.github/workflows/**` (tool change per `docs/AGENT_OPERATING_CONTRACT.md`). Operator approved the workflow changes before this PR was opened.

## Known follow-up

The build emits `EBADENGINE` warnings — several deps want Node `>=20.19` while the image pins `node:20.11-alpine`. Not failing today; worth a separate bump.


scarmani:
## Status: CI green, quorum satisfied, awaiting operator settlement

**CI:** all checks pass on head `5e6dad9`. The only red is `aragora-merge-quorum`, which stays red until Tier-4 evidence is posted by the operator settle path.

**Quorum:** collected against exact head `5e6dad9`, two heterogeneous families, zero dissent.

| Family | Verdict | Counts |
|---|---|---|
| claude (anthropic) | PASS | yes |
| openai | PASS | yes |

`settle_pr.py` reports `quorum_satisfied: true`, `route: operator_human_settlement`. Sole remaining blocker is the human risk settlement — expected, since this touches `.github/workflows/**`, which the operating contract classes as ACR-ineligible (never auto-merges).

### Review findings addressed

- **[P2, openai]** — pinning the gate to the `aragora` self-hosted label makes it depend on fleet availability. Accepted and fixed in `94ba6f1`: moved to `ubuntu-latest` (already the majority convention here, 169 jobs vs 75). The job needs no secrets, no registry login, and no self-hosted resources, and `type=gha` cache is runner-agnostic. Verified: passes in 5m8s on GitHub-hosted vs 5m52s self-hosted.
- **[P3, claude]** — smoke-test comment still cited "the `aragora` runners are shared" after the move. Fixed in `5e6dad9`.
- **[P3, claude]** — health poll budget of 90s exactly equalled Docker's verdict floor (`interval=30s` × `retries=3`), risking boundary flake. Raised to 120s in `5e6dad9`.

### Outstanding P3s — deliberately not addressed here

Non-blocking and non-counting per the tiered gate; noted so they aren't lost:

1. `deploy/Dockerfile.frontend` installs with `npm install --legacy-peer-deps` and never copies `package-lock.json`, so builds are not reproducible. This is currently forced: the `sed` rewrite of the `file:` SDK path desyncs the lockfile, so `npm ci` would reject it. Fixing it properly means restructuring how the SDK is vendored — a design change beyond this PR.
2. `docker.yml`'s frontend publish step passes no build-args, so images from the tag/nightly lane bake in the Dockerfile's `localhost` defaults. Pre-existing; tracked as a follow-up.
3. Stale references to `aragora/live/Dockerfile` remain in `docs/plans/npm-sharp-security-remediation-20260722.md` and `docs/elves/npm-sharp-remediation-20260722/learnings.md`. Left intact deliberately — they are dated historical records of work as it happened, not live documentation.

Stopping the review loop here rather than re-grounding evidence for a further round of P3s.

scarmani:

Head: 59666ba (59666ba5603194848a915f53b948004440561e4c), committed 2026-07-28T03:06:57Z.
PR: #9597.
Model family: claude

Verdict: PASS

Verification performed (head `59666ba` resolved locally on `claude/strange-swartz-0da194`): grepped the whole head tree for `aragora/live/Dockerfile` references in code/config — only the intentional guard in `frontend-image-build.yml` and a test comment remain; confirmed `docker.yml` (tag/nightly publish lane) and both compose files already build `deploy/Dockerfile.frontend` from the correct root contexts; confirmed the deploy-frontend Vercel lane carries the production `NEXT_PUBLIC_*` values (lines 88–89, 190–191) that the removed build-args comment claims; recounted all four changed metrics at the head tree (5,516 test files / 224,983 test functions / 94 workflows / 1,106 docs md — all exact); ran `pytest tests/ci/test_docker_compose_dev.py` (3 passed); read `.dockerignore` (excludes `node_modules/`, `aragora/live/node_modules/`, `aragora/live/.next/`, does not exclude `sdk/typescript`, so the root-context COPY of the SDK works).

- [P3] `.dockerignore:60-62` — the `node_modules/` pattern only matches the repo root, and there is no entry for `sdk/typescript/node_modules`. Verified by reading `.dockerignore` at head. A local `docker build` on a tree where someone ran `npm install` inside `sdk/typescript` will copy that stale `node_modules` into the builder via `COPY sdk/typescript/ /app/sdk/typescript/` (image bloat, potential module-resolution shadowing from the symlinked SDK dir). CI checkouts are clean, so no pipeline impact; adding `sdk/typescript/node_modules/` to `.dockerignore` would close it.
- [P3] `.github/workflows/frontend-image-build.yml` smoke test — unverified edge: if the healthcheck-to-healthy latency ever exceeds 120s on a slow shared runner (first probe fires at ~30s given `interval=30s`, so headroom is real but not proven under load), the gate flakes red. Only an actual slow-runner run would verify; the 60×2s budget and the comment's 90s-floor reasoning look sound.

The load-bearing risks I went hunting for — a publish or compose lane still pointing at the deleted Dockerfile, `.dockerignore` breaking the root-context SDK copy, the Vercel-lane claim in the build-args removal comment being wrong, or the doc metric bumps not matching the tree — all verified clean.

dogfood: yes

scarmani:

Head: 59666ba (59666ba5603194848a915f53b948004440561e4c), committed 2026-07-28T03:06:57Z.
PR: #9597.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

PR: #9597
Exact head: 59666ba5603194848a915f53b948004440561e4c
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #9597 is non-draft and live exact-head checks/merge-packet remain otherwise green.

PR #9635, opened 2026-07-27
fix: update deletion_coordinator.py

scarmani:
(cherry picked from commit c28844f3bb1df344653fc4add10de0a10f05ea24)

### DIFF
--- a/aragora/deletion_coordinator.py
+++ b/aragora/deletion_coordinator.py
@@ -32,6 +32,8 @@ from uuid import uuid4
 
 logger = logging.getLogger(__name__)
 
+_EXPECTED_DELETION_ERRORS = (OSError, RuntimeError, TypeError, ValueError)
+
 
 # ============================================================================
 # Protocols and Types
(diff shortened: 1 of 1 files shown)


scarmani:
Closing this exact head as a preservation-first invalid-scope disposition; the branch is retained and no files or worktrees are deleted.

A fresh exact-head Claude + OpenAI dry run at `6ed420e4f1fa05e3be493c97d1cf759806f24a2c` produced independent CHANGES-REQUESTED P2 findings with the same root cause: the patch narrows intentional outermost deletion-coordinator safety nets, allowing ordinary unexpected deleter/database exceptions to escape instead of recording FAILED/partial `CascadeResult` state, continuing the due-deletion batch, and audit-logging the failure. Existing tests also encode the failure-to-result contract with plain `Exception` inputs.

No evidence was posted and no repair is appropriate in this queue-drain lane: deletion/GDPR failure semantics are destructive/high-risk, while the proposed exception narrowing is not a safe Tier-1 cleanup. Preserve the current broad fail-to-result behavior; any future reland must start from current main with an explicit semantic contract and tier-correct review.

PR #9637, opened 2026-07-27
fix: update rubric.py

scarmani:
(cherry picked from commit 5027e6f7fda595276d498dcc0383e589eb639b58)

### DIFF
--- a/aragora/essay/rubric.py
+++ b/aragora/essay/rubric.py
@@ -187,7 +187,7 @@ async def evaluate_essay(
 
     try:
         response = await judge_agent.generate(prompt)
-    except Exception:
+    except (OSError, RuntimeError, ValueError):
         logger.exception("Judge agent failed during essay evaluation")
         return EssayScore()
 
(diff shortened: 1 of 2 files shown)


scarmani:
Closing under the evergreen no-treadmill/churner rule; the branch and both reviewed heads are preserved, with no deletion or force update.

Head `876b9022f77707c910af10e2a35af598aaacd554` narrowed the essay fallback catch and received independent P1/P2 findings that `AgentError` subclasses escaped the established empty-score fallback. One authorized repair added `AgentError`, producing head `5f08fd820b15094b87c046cd70532fde1ab46a46`. A fresh exact-head dry run then found a new same-area P2: `FallbackTimeoutError` from an exhausted slow `AgentFallbackChain` still escapes instead of returning the established empty `EssayScore()` fallback.

No evidence was posted at either dissenting head. The one-repair budget is exhausted, and repeated same-abstraction findings show that enumerating selected exception classes is the fragile design. Any reland should start from current main and specify the fallback boundary contract explicitly rather than adding another exception to this branch.

Issue #9638, opened 2026-07-27
Autonomous codex workers cannot commit in managed worktrees: --add-dir git dir is gated on admin_approved


scarmani:
Found while running a bounded dispatch batch for #9580. Filing separately because it blocks all autonomous codex work in managed worktrees, not just the corpus task.

## Symptom

Four independent issues (#5788, #5791, #5792, #5808) dispatched to the boss loop returned the **identical** outcome: `worker_outcome=acceptance_gate_failed`, `worker_status=needs_human`, `files_changed=1`, `tests_run=0`. Four different files, four different defect classes, one failure mode — that uniformity is what prompted the dig.

## What actually happened

The workers did the work. From #5808's captured reasoning:

```
Validation passed:
- `ruff check aragora/debate/session.py`: passed
- Bare `except ...: pass` patterns: none found
- `pytest -q tests/debate/test_session.py`: 79 passed
```

Then the commit failed:

```
fatal: Unable to create '.../.git/worktrees/swarm-00c6341e-micro-2/index.lock':
Operation not permitted
exited 128
```

The acceptance gate saw no commit and rejected the deliverable.

## Root cause

Codex ran with `sandbox: workspace-write [workdir, /tmp, $TMPDIR]`. The workdir is the linked worktree `/Users/armand/Development/aragora/.worktrees/codex-auto/swarm-<id>`, but a linked worktree's `.git` is a *file* pointing at `/Users/armand/Development/aragora/.git/worktrees/<name>/` — **outside every writable root**. So `git add`, `git commit`, even `git commit --allow-empty` fail: the index lock cannot be created.

The launcher already knows about this — `_resolve_worktree_gitdir` (`aragora/swarm/worker_launcher.py:820-861`) exists precisely to hand that directory to codex via `--add-dir`, and its docstring spells out the failure. But at `worker_launcher.py:693` it is gated:

```python
git_dir = self._resolve_worktree_gitdir(worktree_path) if admin_approved else ""
if git_dir:
    cmd.extend(["--add-dir", git_dir])
```

`admin_approved` comes from an explicit `admin_approved` key on the work order or metadata (`worker_launcher.py:732-736`), which the boss-loop dispatch path does not set. So the mitigation is present but never active for ordinary autonomous runs.

## Why this matters beyond one failed batch

1. **Every autonomous codex worker in a managed worktree can analyse and edit but cannot commit.** Work is performed, then discarded at the gate. The loop reports `needs_human` on work that may well have been correct.
[...]

PR #9633, opened 2026-07-27
fix: update performance_monitor.py

scarmani:
(cherry picked from commit 0206cb3a635cb1188e90dde960474fdf31b8c76a)

PR #9631, opened 2026-07-27
fix: update session.py

scarmani:
(cherry picked from commit 4f7ec5a0f90f9787bc79deb5e968f84be2fc2936)

Issue #9580, opened 2026-07-24
[TW-01] rev-7 corpus restock is blocked upstream: zero open candidates have dispatch history


scarmani:
Filed from the #9519 follow-through (audit-only, no `boss-ready`/`autonomous` labels — corpus membership is founder-owned and must never be auto-dispatched or auto-settled, per `docs/status/B0_CORPUS_REV6_RECONCILIATION.md` L7-9).

## Why this is filed

The B0 proof surface now states plainly that the corpus is exhausted (renderer note added in #9542): all 13 rows resolve to `issue_already_resolved`, `unique_issues_succeeded: 0`. The standing follow-up has been "restock TW-01 to rev-7 so the proxy no-rescue signal stops reading all-neutral."

**That restock is currently blocked, and not by paperwork.** The blocker is upstream of the corpus file.

## The blocker

`tests/benchmarks/test_corpus_freshness.py:193-215` enforces the rev-3 honesty invariant: every `expected_status: in_progress` entry must already have at least one `worker_outcome` row in `.aragora/overnight/boss_metrics.jsonl` — i.e. it must have been dispatched to the autonomous boss loop *before* it may join the corpus. This is the mechanical anti-cherry-picking gate, and it is the right gate.

Cross-checking the vetted staging pool (`tests/benchmarks/corpus_rev4.json`, 25 entries not yet promoted) against live issue state and the dispatch ledger:

| Candidate state | Count | Eligible for rev-7? |
|---|---|---|
| OPEN, **no** dispatch history | 15 | No — fails the dispatch invariant |
| CLOSED, has dispatch history | 10 | No — already graduated/closed |
| **OPEN + has dispatch history** | **0** | — |

The eligible set is empty. Every open candidate lacks dispatch history; every dispatched candidate is already closed.

Open-but-undispatched: #5188, #5788, #5791, #5792, #5793, #5794, #5801, #5808, #5809, #5810, #5811, #5812, #5821, #5823, #5825, #5883.

Supporting signal: `.aragora/overnight/boss_metrics.jsonl` was last written **2026-07-12** (723 rows, only 29 tagged `B0-cohort`). Dispatch against this cohort is dormant, which is the actual reason no fresh execution evidence exists — corpus exhaustion is the symptom, not the cause.

## Required order of operations

[...]

PR #9639, opened 2026-07-27
fix(acceptance-gate): regression criteria are not a request for tests (#9638)

scarmani:
## Summary

An entire class of autonomous refactor work could not pass the acceptance gate, no matter how correct the deliverable was.

`_criterion_requires_tests` (`aragora/swarm/acceptance_gate.py`) matched any criterion mentioning "test" — including **"Existing tests still pass"**. That is a regression guard, not a request for new test files, and it is the last acceptance criterion on every auto-generated exception-narrowing and silent-exception issue. So those issues always tripped `test_presence_missing` unless the worker invented an unrelated test file.

## Evidence this is live, not theoretical

Dispatched #5788, #5791, #5792, #5808 through the boss loop. All four produced plausible fixes and **published PRs** — #9633, #9635, #9637, #9631 — and all four were then marked `acceptance_gate_failed` / `needs_human`. Four different files, four defect classes, one identical verdict; ledger rows show `publish_action: pr_created`, `sanitizer_outcome: accepted`, `has_deliverable: true`.

Reproduced directly against #5788's real criteria:

```
requires_tests? True
passed: False   failure_classes: ('test_presence_missing',)
```

Two compounding effects: the ledger records success as failure (so any capability metric built on `worker_outcome` understates the loop — this is why I am **not** building a rev-7 corpus on that window, see #9580), and the loop halts on `needs_human`, collapsing throughput to one issue per invocation.

## Reviewed design tradeoffs

- **Narrowing the match vs. rewriting the issue templates:** templates would fix new issues only and leave every existing one ungateable; the gate is the correct place.
- **Preserving the module's conservative default.** The docstring states false negatives are cheaper than false positives, and that is retained: criteria are classified in order — explicit request for tests wins (so "add tests and keep existing tests passing" still demands tests), regression-only criteria are skipped, and **anything else mentioning tests still requires them**. Only phrasings that clearly guard existing tests are exempted.
- **Why an action verb is required for the creation pattern:** the bare noun ("unit tests") also appears in "existing unit tests still pass", so matching nouns alone would reintroduce the bug.
- **Not touched:** the `test_presence` check itself, the file-scope and file-creation checks, and the `(new)`-marker path.

## Verification

- `pytest tests/swarm/test_acceptance_gate.py` — 40 passed, including new coverage: the exact #9633 scenario now passes; 7 regression phrasings parametrised as non-triggering; 5 creation phrasings parametrised as still-triggering (so the real "asked for tests" case, e.g. #5188, is unweakened); and an unanticipated-phrasing case pinning the conservative fallback.
- `pytest tests/swarm/test_dispatch_followups.py tests/swarm/test_worker_launcher.py` — 178 passed, no regressions in the gate's callers.

## Note on the original #9638 diagnosis

That issue was filed naming a sandbox/git-permission problem as the cause. That problem is real — the validation micro-task cannot write its `git commit --allow-empty` marker because a linked worktree's git metadata sits outside the codex sandbox's writable roots — but it is **not** what failed these runs. The gate never inspects a marker commit; it compares changed paths against acceptance criteria. I have corrected the issue accordingly. The git-permission finding stands on its own and is deliberately **not** fixed here: granting `--add-dir <repo>/.git` would hand the sandbox write access to `.git/hooks` and `.git/config`, which is a sandbox escape, and an existing test pins that gating as intentional.

Refs #9638, Refs #9580

### DIFF
--- a/aragora/swarm/acceptance_gate.py
+++ b/aragora/swarm/acceptance_gate.py
@@ -59,6 +59,33 @@ _TEST_CRITERION_RE = re.compile(
     re.IGNORECASE,
 )
 
+# Criteria that explicitly ask for test files to be written. Requires an
+# action verb: the bare noun ("unit tests") also appears in regression
+# phrasings like "existing unit tests still pass".
+_TEST_CREATION_RE = re.compile(
+    r"\b(?:add|writ(?:e|ing)|creat(?:e|ing)|includ(?:e|ing)|extend(?:ing)?|"
+    r"introduc(?:e|ing)|updat(?:e|ing)|cover)\b[^.;]{0,48}?\b(?:tests?|coverage)\b",
+    re.IGNORECASE,
+)
+
+# Criteria that only require ALREADY-EXISTING tests to keep passing. These are
+# regression guards, not requests for new tests, and on their own they must not
+# trigger the test-presence check — otherwise every pure-refactor issue becomes
+# structurally ungateable. The auto-generated narrowing/silencing issues all
+# carry "Existing tests still pass", so this is the common case, not an edge
+# case (#9638).
+_TEST_REGRESSION_ONLY_RE = re.compile(
+    r"\b(?:"
+    r"existing\s+(?:unit\s+|integration\s+)?tests?|"
+    r"(?:all|current)\s+tests?\s+(?:still\s+)?pass(?:es|ing)?|"
+    r"tests?\s+(?:still\s+|continue\s+to\s+|should\s+still\s+)?pass(?:es|ing)?|"
+    r"no\s+(?:new\s+)?test\s+(?:failures|regressions)|"
+    r"(?:do(?:es)?\s+not|don't|doesn't)\s+break\s+(?:existing\s+)?tests?|"
+    r"without\s+breaking\s+(?:existing\s+)?tests?"
+    r")\b",
+    re.IGNORECASE,
+)
+
 # Pull explicit pytest target file paths out of a criterion.
 _PYTEST_TARGET_RE = re.compile(
     r"(?:pytest|python\s+-m\s+pytest)\s+([^\s'\"`]+\.py)",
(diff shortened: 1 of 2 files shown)


scarmani:

Head: 0fd1cdb (0fd1cdb551b3f05be43ec3ef429b57de2eb13521), committed 2026-07-27T15:17:20Z.
PR: #9639.
Model family: claude

Verdict: PASS

Verified locally at 0fd1cdb: `tests/swarm/test_acceptance_gate.py` 48 passed; `test_dispatch_followups.py` + `test_issue_scanner.py` 67 passed. Branch touches only the 2 files. No ReDoS surface (no nested quantifiers; alternation is literal-separated `\s+`), subtraction spans are `\b`-delimited and replaced with a space so no word-gluing, and `_pytest_targets_from_criteria` / `file_creation` still read raw text — defense in depth preserved.

- [P3] `aragora/swarm/acceptance_gate.py:81` — alternative 2 (`tests?\s+(?:still|continue to|should still)\s+pass`) is not qualifier-aware, so it contradicts the comment at lines 71–72 ("a bare 'tests pass' is likewise not matched… 'the new tests pass' presupposes tests that must be written"). Probed: `"New tests pass"` → demands tests (as documented), but `"New tests still pass"` and `"Newly added tests still pass"` are fully subtracted → residue `"New "` → no demand. The stated invariant holds only for the exact `pass` form, not `still pass`. Low-realism phrasing, but the comment overstates the guarantee.
- [P3] `aragora/swarm/issue_scanner.py:378` (latent coupling, not a live defect) — the `test_coverage` scanner, whose entire purpose is creating a new test file, leads with `"All tests pass"`, which this change now subtracts. It stays gated only incidentally, via sibling criteria `"At least 8 test functions"` / `"Tests complete in under 10 seconds"` (both confirmed to still demand tests). Editing that criteria list would silently un-gate the one class that must write tests. Worth a pinning test in `tests/swarm/test_acceptance_gate.py` that feeds the scanner's actual `test_coverage` criteria list and asserts `test_presence_missing`.
- [P3] `tests/swarm/test_acceptance_gate.py` — `test_unanticipated_test_phrasing_keeps_conservative_default` docstring ("Neither pattern matching must still fall through to requiring tests.") is ungrammatical and misdescribes the mechanism: `"The assert helper is exercised end to end"` *does* match `_TEST_CRITERION_RE` via the `assert` token; only the guard regex misses. Say "the guard does not match, so the conservative fallback fires."
- [P3] `tests/swarm/test_acceptance_gate.py:~150` — `test_regression_only_criterion_does_not_demand_new_tests` asserts only `"test_presence_missing" not in failure_classes`; the parametrized sibling uses the stronger `"test_presence" not in checks_run`. Prefer the stronger assertion in the headline #9638 case so a future change that runs the check but passes it for the wrong reason still fails.

dogfood: yes

scarmani:

Head: 0fd1cdb (0fd1cdb551b3f05be43ec3ef429b57de2eb13521), committed 2026-07-27T15:17:20Z.
PR: #9639.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9641, opened 2026-07-27
fix(quorum): ungrounded reviewers must not gate merges [Tier 4]

scarmani:
## Problem

A reviewer reached over a **single-shot API transport** (VibeProxy, a family API, OpenRouter) gets the prompt text and **no tools**. It cannot read the repo or reach the network, so it cannot verify any claim whose truth lives outside the diff — while the prompt's severity contract requires that any `[P1]`/`[P2]` block the merge. The cheapest way to satisfy that contract is a confident guess about the surrounding repository, and those guesses come back wrong.

This is structural, not model-specific. A base-image bump, a dependency pin, an infra alignment — any diff whose correctness lives *outside* the diff — will always admit a manufactured blocking finding.

## Evidence

[#9505](https://github.com/synaptent/aragora/pull/9505), a four-file container base-image bump, was blocked by three ungrounded reviews whose blocking findings were **all false**:

| Finding | Reality |
|---|---|
| `node:24.18-alpine` does not exist | It does — pulled, digest `sha256:a0b9bf06…`, built container reported `v24.18.0` |
| 24.18 is not an LTS line | Node 24 has been Active LTS since Oct 2025 |
| `--only` was removed in npm 9 | Still omits devDependencies under npm 11 (verified by running it) |

Both **grounded CLI reviewers passed the same head.** Re-collected with the default transport, #9505 got a clean 2/2 and merged as `35d66d03d3`.

I verified the capability difference directly: the CLI reviewer, asked about a file in no diff, correctly reported the absent `engines` field in `aragora/live/package.json`. The proxy transport cannot read anything.

## Changes

- `ReviewerResult`/`EvidenceItem` carry `grounded`. An ungrounded review from a family that **has** a CLI transport (`GROUNDED_TRANSPORT_FAMILIES`: claude, openai, grok, gemini) never counts toward quorum and never creates blocking dissent — the same both-directions treatment advisory-only families already get. It still posts, parses, and lints.
- **Families with no CLI harness are exempt** (mistral, hermes, Chinese-routed). Demoting their only transport would delete them from the reviewer pool and strand Tier 0-2 quorums that legitimately count them today. Deliberately narrower than "no ungrounded reviewer anywhere".
- **Reviewers go CLI-first.** This also closes a latent trap: any machine with `OPENAI_API_KEY` set silently routed *every* OpenAI review through the ungrounded API, invisibly.
- **Prompt clause** — do not assert facts about files you were not shown; tag unverifiable concerns `[P3]` and name what would verify them.
- `grounded` threaded through every `EvidenceItem` construction path (fresh collect, `from_raw`, clone, relint) so a prepared artifact cannot smuggle authority back in.

## Compatibility

**Default behavior is unchanged.** In direct mode the transport policy returns without touching the proxy, so the grounded CLI runs exactly as before. `vibeproxy-required` semantics are also unchanged — it still reaches the proxy first and fails closed, never touching the CLI or an OpenRouter fallback (regression caught by the existing test during development and fixed).

Prepared artifacts written before this field default to `grounded=True` rather than fail-closed. Unlike `severity_gated`, ungrounded lowers *both* counting and blocking, so neither default is uniformly stricter — and demoting every legacy artifact would strand in-flight prepared packets mid-settlement. Every live collect path sets the field explicitly. Called out for reviewer attention.

## Reviewed design tradeoffs

- **Demote vs. remove the transport.** VibeProxy is not an independent fallback — it also refreshes the access tokens the Claude CLI profiles consume (`com.aragora.claude-profile-sync`, blank refresh tokens per #9496), so if it dies the CLI expires within hours anyway. Demoting authority rather than deleting the transport keeps diagnostics and is reversible.
- **Scope of the carve-out.** Marking *all* API transports non-counting is more principled but shrinks the reviewer pool for API-only families. Left as a separate roster decision, documented in `REVIEW_AUTHORITY_PRINCIPLES.md`.

## Tests

371 passed (`tests/swarm/test_quorum_evidence.py` + `tests/governance/test_tiered_merge_gate_quorum_policy.py`). Six new tests cover the grounding contract in both directions, the API-only carve-out, and prepared-artifact roundtrip. Eight existing tests updated to the new contract.

Governance: `docs/REVIEW_AUTHORITY_PRINCIPLES.md` gains a *Transport grounding* section; changing `GROUNDED_TRANSPORT_FAMILIES` is Tier 4 under the existing family-additive rule.


scarmani:

Head: d7f1064 (d7f1064d1ccafe95d9aa71001606a220f6d5b2ec), committed 2026-07-27T17:43:33Z.
PR: #9641.
Model family: claude

Verdict: PASS

Verified against the checkout (I read `aragora/agents/transports/vibeproxy.py`, `claude_vibeproxy.py`, and the unchanged regions of `quorum_evidence.py` / the test file at the PR base): `run_claude_vibeproxy` really is an `attempted=False` no-op in direct mode; `TransportMode.REQUIRED.value == "vibeproxy-required"` matches the config-error fallback in `_claude_transport_mode_is_required`, and both sides degrade identically on `VibeProxyConfigurationError`, so mode resolution cannot diverge from the transport's own fail-closed behavior; `from_env()` is env-parsing only (no network); Codex CLI runs `--sandbox read-only --ephemeral`, so expanding its use to key-present machines is safe; at apply time `would_count`/`verdict` are recomputed from the body while `grounded` is preserved, and the `__post_init__` choke point re-demotes, so the roundtrip holds and problems don't duplicate.

- [P3] `aragora/swarm/quorum_evidence.py` (`_coerce_grounded_flag`): the missing-key default of `True` means an artifact can recover counting authority by simply *omitting* `grounded` — the exact recovery the explicit-`null` sentinel is documented as preventing. Verified: `grounded` is the only authority-bearing field trusted verbatim at apply (everything else is relinted). The legacy tradeoff is documented, but the "cannot smuggle an ungrounded review back into counting_families" comment overstates coverage, and there is no sunset/version stamp to close the omission path once in-flight packets drain. (Net-new forgery power is nil — a forger could always delete the item — so this is a rationale/comment gap, not a gate hole.)
- [P3] `_run_claude_reviewer` failure path: the removed elapsed-charging logic means a wedged CLI now consumes its full budget (600s default), then the proxy re-derives a full attempt budget from `reviewer_timeout=timeout` (verified in `claude_vibeproxy.py`), then the API runs — worst-case wall-clock roughly doubles versus the old proxy-elapsed charging. Bounded and failure-path-only, but worth a combined deadline.
- [P3] `_run_openai_reviewer`: when both Codex CLI and the API fallback fail, the API's error is silently dropped (returns `result` from the CLI untouched), unlike the claude path which composes both errors — a doubly-failed OpenAI leg will only ever show the Codex error.
- [P3] Required-mode consequence: under `vibeproxy-required`, every Claude review is `grounded=False`, so the claude family can never produce countable evidence — Tier 2+ quorums needing the western-frontier claude leg are unsatisfiable on such a machine. This matches the directive (required = proxy-or-nothing, ungrounded = advisory), but nothing logs or documents that required mode now yields zero countable claude signal; an explicit warning at collect time would prevent silent quorum starvation.
- [P3] `tests/swarm/test_quorum_evidence.py` (`test_claude_reviewer_prefer_failure_uses_direct_path`, base line 406): still passes under CLI-first (verified the body — CLI succeeds with the full 600s before the proxy is touched), but it no longer exercises the prefer-failure→direct fallback its name and comments describe; the proxy stub is dead weight. Rename or fold into the new prefer-mode tests.

dogfood: yes

scarmani:

Head: d7f1064 (d7f1064d1ccafe95d9aa71001606a220f6d5b2ec), committed 2026-07-27T17:43:33Z.
PR: #9641.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

PR: #9641
Exact head: d7f1064d1ccafe95d9aa71001606a220f6d5b2ec
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #9641 is non-draft and live exact-head checks/merge-packet remain otherwise green.

PR #9414, opened 2026-07-17
feat(debate): crux cards behind enable_crux_cards — receipts name the load-bearing dissent (#9046 phase 1)

scarmani:
## Summary

W2 designated product feature per the 30-day plan (work order on #9046, implements #8227 phase 1).

Standard debates already run `CruxDetector` in the majority-consensus path, but the output never reached the `DecisionReceipt`: every standard-debate ODR export carried the absent marker `no crux set recorded for this decision`. This wires the existing detector into the receipt path behind a new `DebateProtocol.enable_crux_cards` flag (**default OFF**).

- `aragora/protocols/debate.py`: `enable_crux_cards` flag (reuses `crux_finder_top_k` / `crux_finder_min_score` tuning)
- `aragora/debate/crux_cards.py` (new): `build_crux_cards()` — prefers the belief-analysis network on ctx, falls back to building one from proposer/critic messages (mirrors `WinnerSelector.analyze_belief_network`); returns `None` when nothing clears `min_score`
- `aragora/debate/phases/consensus_phase.py`: flag-gated best-effort attach after consensus, regardless of outcome (cruxes matter most when consensus **fails**); stored on `result.metadata['crux_cards']`; can never break the phase
- `aragora/gauntlet/receipt_models.py`: additive `DecisionReceipt.cruxes` block with per-crux dissent attribution (author, contesting_agents, affected_claims, component scores); `to_dict` omits the key when `None` so **flag-off receipts stay byte-identical**; `from_dict` round-trips; `from_debate_result` populates from metadata
- `aragora/gauntlet/odr_export.py`: ODR `cruxes` block now populated from `receipt.cruxes`; explicit `crux_set=` still wins; absent marker unchanged otherwise. **No schema change** — `cruxes` was already required in `odr_schema.json` with present/absent validation, so `aragora-verify` verifies pre-crux receipts unchanged.

## Acceptance (from the #9046 work order)

- ✅ `enable_crux_cards=True` → receipt `cruxes` block names the load-bearing disagreement(s) with agent attribution
- ✅ flag OFF → byte-identical receipts (tested: no `cruxes` key in `to_dict()`)
- ✅ additive-only schema; `aragora-verify` still verifies pre-crux receipts (jsonschema validation test included)

## Test plan

[...]

PR #9643, opened 2026-07-27
fix(debate): derive crux factor edges from critiques (#9581)

scarmani:
## Summary

Fixes the core of #9581. Crux cards produced **nothing** in any standard debate
— at any `min_score`, with any agents. The feature shipped in #9414 (receipt
schema) and #9506 (CLI flag) was inert.

## Root cause

`CruxDetector` scores a claim by how much authors disagree about it, which it
reads from the belief network's **factor edges**. Both paths that build that
network called `add_claim()` and never `add_factor()`. `node_factors` was empty
for every node, so `compute_disagreement_scores()` never entered its loop body
and no crux ever cleared a threshold.

## Fix

Each `Critique` already **records the disagreement**: which agent contested
whose proposal (`agent` → `target_agent`) and how hard (`severity`, 0-10). Those
become `CONTRADICTS` edges with strength `severity/10`. Derived from the
debate's own structured records — no prose parsing, no extra model call.

Supporting details, each from review:

- **`severity <= 0` is skipped.** The debate emits placeholder Critiques with
  `severity=0.0` when a critic times out; flooring those to a nonzero weight
  would manufacture crux cards for disagreements that never happened.
- **Each critique anchors to the proposal it actually addressed.** Revisions are
  recorded with `role="proposer"` too, so pinning to the first proposal
  attributed round-N critiques to round-1 text. Matched via
  `Critique.target_content`, **by prefix** — several agents record only
  `proposal[:200]`, so exact comparison would miss every longer proposal.
- **A supplied `belief_network` is not linked into.** Its claim ids are
  KM-derived, so message-derived ids match nothing: every edge would no-op while
  leaving orphan claims in shared state. An earlier revision of this PR claimed
  in a comment that it *did* cover that path; both reviewers caught it and I
  verified they were right (0 edges linked).

## Scope, deliberately narrowed

This PR now touches **`crux_cards.py` only** — `crux_detector.py` is identical
to `main`. Earlier revisions also seeded the claim's own author into the
disagreement map, excluded that author from `contesting_agents`, and raised the
asserted-claim prior. Every one of those drew findings across three rounds:

- the seed reads a **post-propagation** posterior (`detect_cruxes` propagates
  first), making disagreement non-monotone in severity;
- factors are indexed under both endpoints, so one critique inflated
  `total_disagreements` by ~2 and could surface the critique text itself as a
  card;
- the change shifted behaviour for four **un-gated** consumers (belief API,
  `crux_mode`, `cruxset_emission`, `winner_selector`).

They are all facets of the ordering defect in **#9644** and belong there,
designed together, rather than patched incrementally underneath an
edge-construction fix. This is the repo's own documented cure for a reviewer
flagging the same area round after round.

## What this delivers, and what it does not

**Delivers:** 0 crux cards before, **4 after**, on the same contested two-agent
debate — which unblocks the W3 exit criterion "crux cards in ≥1 published
receipt".

**Does not:** `total_disagreements` stays 0 (`CruxDetector` counts a claim as
disagreed-about only when ≥2 *other* authors relate to it, so a two-agent debate
cannot register one) and `contesting_agents` stays empty. **This PR therefore
does not meet the acceptance criterion I wrote into #9581** — that is #9644's
scope, and the tests assert the honest post-conditions rather than the ones the
dropped change produced.

98 crux tests pass, including `crux_mode` and `crux_detector`.

### DIFF
--- a/aragora/debate/crux_cards.py
+++ b/aragora/debate/crux_cards.py
@@ -69,6 +69,19 @@ def build_crux_cards(
     analysis = detector.detect_cruxes(top_k=top_k, min_score=min_score)
     if not analysis.cruxes:
         return None
+    if not analysis.total_disagreements:
+        # `crux_score` is a composite, so a claim can clear `min_score` on
+        # uncertainty and centrality alone. Emitting those as "crux cards" would
+        # publish load-bearing-disagreement claims into a DecisionReceipt with no
+        # disagreement actually detected behind them — an overclaim in exactly
+        # the surface auditors read. Absent is honest; mislabelled is not.
+        #
+        # `CruxDetector` registers a disagreement only when >=2 authors *other
+        # than the claim's own* relate to it, so a two-agent debate trading
+        # reciprocal critiques never reaches one. Making a two-agent debate
+        # attribute dissent is #9644's scope, not this edge-construction fix.
+        logger.info("crux_cards_suppressed_no_disagreement claims=%d", analysis.total_claims)
+        return None
 
     return {
         # CruxClaim.to_dict() carries per-crux dissent attribution:
@@ -150,7 +163,14 @@ def _target_claim_for_critique(messages: list[Any], target: str, critique: Any) 
 
     wanted = str(getattr(critique, "target_content", "") or "").strip()
     if wanted:
-        for i, msg in candidates:
+        # Newest first: a revision commonly keeps the earlier proposal's header
+        # or opening paragraph, so a prefix comparison can match round 1 as well
+        # as round N. Scanning backwards biases toward the current position —
+        # the same bias as the no-match fallback below — instead of silently
+        # re-introducing the stale-text attribution this function exists to
+        # avoid. A degenerate short early proposal ("Agreed.") would otherwise
+        # absorb every critique whose excerpt starts with it.
+        for i, msg in reversed(candidates):
             content = str(getattr(msg, "content", "") or "").strip()
             if not content:
                 continue
@@ -200,7 +220,10 @@ def _link_critiques(network: Any, messages: list[Any], critiques: list[Any]) -> 
             severity = float(getattr(critique, "severity", 0.0) or 0.0)
         except (TypeError, ValueError):
             severity = 0.0
-        if severity <= 0:
+        # `severity != severity` catches NaN: every NaN comparison is False, so a
+        # NaN would pass `<= 0` and then `min(1.0, nan/10.0)` returns 1.0 —
+        # a maximum-strength edge from an unusable value.
+        if severity != severity or severity <= 0:  # noqa: PLR0124 - NaN check
             continue
 
         target_claim = _target_claim_for_critique(messages, target, critique)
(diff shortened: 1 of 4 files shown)


scarmani:

Head: aea2195 (aea2195394573befc54d6c463dde665602ee1276), committed 2026-07-27T16:20:00Z.
PR: #9643.
Model family: claude

Verdict: PASS

Verified against the codebase at head aea2195, not just the diff: `BeliefNetwork.add_factor` does return `None` for unknown claim ids (so the "no-op on a KM-seeded network" claim holds — belief.py:563), `CruxDetector.compute_disagreement_scores` does require ≥2 distinct connected authors with nonzero variance (crux_detector.py:239), `Critique.severity` is the documented 0–10 scale (core_types.py:171), `DebateResult.critiques` exists (core_types.py:391), `_link_critiques` is only reachable from the message-fallback path so a supplied network truly gets no new claims/edges, and `_attach_crux_cards`'s except tuple covers the linker's plausible failure modes. All 24 tests in `tests/debate/test_crux_cards.py` pass at the PR head. The NaN guard, zero-severity placeholder skip, self-critique skip, and >10-severity clamp are all correct; flag-off behavior is untouched. The new `total_disagreements` suppression is a deliberate flag-on-only tightening that removes exactly the mislabelled-card class, and per #9581 no real card emission existed to regress.

- [P3] `aragora/debate/crux_cards.py:54` — the comment block begins mid-sentence ("# NOT linked into a supplied ``belief_network``…") with no subject; it reads as a truncated first line. Cosmetic, but this is a load-bearing explanatory comment and the missing opening ("Critiques are…") costs the next reader a beat.
- [P3] `aragora/debate/crux_cards.py` `_link_critiques` — the source-anchoring comment slightly overstates the degradation guarantee: if the unasserted 1:1 critiques↔critic-messages append-order invariant ever breaks such that a critic has more critiques than critic-role messages, the surplus critiques are silently dropped (`index >= len(source_claims)` → `continue`), which loses a real disagreement edge and can flip `total_disagreements` to 0, suppressing cards entirely — not merely "mis-anchored source prose." Invariant holds in today's producers per the comment's own audit, so non-blocking; a debug log on the drop path would make a future break diagnosable.
- [P3] `_target_claim_for_critique` — the bidirectional `longer.startswith(shorter)` prefix match means a very short `target_content` (or a short proposal like "Agreed.") can match the wrong revision; impact is attribution prose only and the newest-first bias equals the no-match fallback, so this is accepted-risk as documented in the docstring.

dogfood: yes

scarmani:

Head: aea2195 (aea2195394573befc54d6c463dde665602ee1276), committed 2026-07-27T16:20:00Z.
PR: #9643.
Model family: openai

Verdict: PASS

- [P3] `tests/debate/test_crux_cards.py`: tests now import and exercise private helper `_network_from_messages`; this locks tests to implementation details and may make future refactors noisier, but it is acceptable for regression coverage of edge construction.

No blocking findings.

dogfood: yes

PR #9647, opened 2026-07-27
fix(frontend): stop published frontend images dialing localhost for WebSockets

scarmani:
## Problem

Images published from `deploy/Dockerfile.frontend` inherit its localhost ARG defaults, and Next.js inlines `NEXT_PUBLIC_*` into the client bundle at build time. A no-build-args image carries **514** occurrences of `localhost:8080` and **554** of `ws://localhost:8765`.

Pre-existing, not a regression. Surfaced as a P3 during quorum review of #9597, which makes `deploy/Dockerfile.frontend` the single frontend image definition.

## What's actually broken (narrower than reported, and asymmetric)

`src/config.ts` already derives per-host at runtime, but only on one side:

- `resolveApiBaseUrl` **overrides** a localhost-baked value when served from a real hostname → HTTP self-heals to `api.<host>`.
- `resolveWsUrl` short-circuited on `if (envValue) return envValue` → **no override**.

Executing both resolvers verbatim:

| Scenario | API resolves to | WS resolves to |
|---|---|---|
| published image @ `aragora.ai` | `https://api.aragora.ai` ✅ | `ws://localhost:8765/ws` ❌ |
| self-hosted @ `aragora.acme.com` | `https://api.aragora.acme.com` ✅ | `ws://localhost:8765/ws` ❌ |

So HTTP works; every WebSocket feature (~12 hooks — spectate, live debate, oracle, nomic, gauntlet, control-plane) silently dials localhost. Unlike HTTP there is no same-origin rewrite to fall back on.

## Fix

**`resolveWsUrl` made symmetric with `resolveApiBaseUrl`.** A localhost WS URL is ignored when the page is served from a non-local hostname, deriving `wss://api.<host>/ws` instead. The rescue is keyed on the *serving hostname*, so a bundle whose `NEXT_PUBLIC_API_URL` made `_isProductionBuild` true still keeps a localhost WS endpoint when the browser is itself on localhost.

This makes the published image **host-agnostic**: one build works on aragora.ai, on a self-hoster's domain, and locally. `deploy/Dockerfile.frontend` and `docker.yml` gain comments recording why no production build-args belong there.

## What this PR originally did, and why that was wrong

The first revision added production build-args to `docker.yml`. Quorum review raised two P2s, both correct:

1. **It vendor-pins the published image.** `deploy/helm/aragora/values.yaml` pulls a published frontend image, and a self-hoster's `NEXT_PUBLIC_*` *runtime* env cannot reach an already-inlined bundle — their browser traffic would have gone to Aragora's backend. The decisive fact I had missed: **aragora.ai does not run this image.** `deploy-frontend.yml`'s Vercel lane builds the site directly with explicit production values, so the ghcr image's audience *is* self-hosters.

2. **The value had no `/ws` path.** Copied from `deploy-frontend.yml`. `useDebateWebSocketStore` reads `NEXT_PUBLIC_WS_URL` directly and passes it unmodified to `new WebSocket()`, and the server only registers `/ws` (`servers_route_registration.py:344`), so it would connect to `/`.

Also corrected from the first revision: I claimed the missing-`NEXT_PUBLIC_WS_URL` error banner fires on a localhost-baked image. It does not — the var is set (to localhost), so `isProd && !_WS_URL` is false.

Equally rejected: **flipping the Dockerfile ARG defaults to production**, for the same reason as (1) — `config.ts` honours a non-local baked URL verbatim. The localhost default is precisely what enables per-host derivation.

## Verification

Built the image three ways and grepped the emitted `.next` bundle. Raw `grep localhost` is **not** a clean signal — the source has many hardcoded devDefault fallback literals (409 remain even with production args). Two checks that are:

- In the config chunk, a production-baked build contains `"wss://api.aragora.ai"` (bare), a string that exists nowhere in the source. Duplicate localhost literals collapse 3→1 and 2→1, leaving only the devDefaults.
- A **sentinel** build-arg lands in **105** chunks (API) and **99** (WS), exactly matching the 514→409 (−105) and 554→455 (−99) drops. That accounts for every inlined site.

Tests: 9 new cases across three host-pinned suites (jsdom's `window.location` is non-configurable, so the serving hostname is pinned per file via `@jest-environment-options`). 4 failed before the fix, all pass after. Frontend suite: **1097 passed** across config + hooks; full run 4019 passed with one unrelated suite failing on a missing `react-is` transitive dep from a local `npm install --legacy-peer-deps` — reproduced identically with this change stashed; CI uses `npm ci`. `tsc --noEmit` and eslint clean.

## Adjacent findings (filed separately, not fixed here)

- #9648 — `runtimeBackend.ts` hardcodes `https://api.aragora.ai` and ignores `NEXT_PUBLIC_*`; `apiFetch` (23 call sites) uses it in the browser, so self-hosted deployments already route those calls to Aragora's production API.
- #9651 — `deploy-frontend.yml` sets `NEXT_PUBLIC_WS_URL=wss://api.aragora.ai` with no `/ws`, which breaks `useDebateWebSocketStore` on the live Vercel deployment.

## Review note

⚠️ Touches `.github/workflows/docker.yml` — a protected surface per `docs/AGENT_OPERATING_CONTRACT.md`. The change there is now **comment-only** (no functional change); the operator approved a workflow edit before this PR was opened, and the functional part has since been reverted.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -186,14 +186,15 @@ jobs:
             type=raw,value=latest,enable={{is_default_branch}}
             type=raw,value=${{ steps.version.outputs.version }},enable={{is_default_branch}}
 
-      # Next.js inlines NEXT_PUBLIC_* at build time, so the published image needs
-      # them passed here. Without build-args the image inherits the Dockerfile's
-      # localhost ARG defaults; src/config.ts then derives api.<serving-host> at
-      # runtime, so the image still works, but it is built as a development
-      # bundle: NEXT_PUBLIC_API_URL contains "localhost", so _isProductionBuild is
-      # false, SSR renders dev defaults while the client renders production ones
-      # (hydration mismatch), and the "NEXT_PUBLIC_WS_URL not set" banner fires.
-      # Values match deploy-frontend.yml so both publishing lanes agree.
+      # Deliberately passes no NEXT_PUBLIC_* build-args. This is a general-purpose
+      # published image: self-hosters pull it (see deploy/helm values.yaml) and
+      # serve it from their own domain, and NEXT_PUBLIC_* is inlined at build
+      # time, so baking https://api.aragora.ai here would make every self-hosted
+      # deployment send browser API and WebSocket traffic to Aragora's backend.
+      # The image is host-agnostic instead: it inherits the Dockerfile's localhost
+      # defaults, which src/config.ts recognises as unusable and replaces with a
+      # derivation from the serving host. aragora.ai itself is built with explicit
+      # production values by deploy-frontend.yml's Vercel lane, not from this image.
       - name: Build and push frontend
         uses: docker/build-push-action@v5
         with:
@@ -205,9 +206,6 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
           platforms: linux/amd64,linux/arm64
-          build-args: |
-            NEXT_PUBLIC_API_URL=https://api.aragora.ai
-            NEXT_PUBLIC_WS_URL=wss://api.aragora.ai
 
       - name: Build frontend for scanning
         run: |
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -186,6 +186,14 @@ jobs:
             type=raw,value=latest,enable={{is_default_branch}}
             type=raw,value=${{ steps.version.outputs.version }},enable={{is_default_branch}}
 
+      # Next.js inlines NEXT_PUBLIC_* at build time, so the published image needs
+      # them passed here. Without build-args the image inherits the Dockerfile's
+      # localhost ARG defaults; src/config.ts then derives api.<serving-host> at
+      # runtime, so the image still works, but it is built as a development
+      # bundle: NEXT_PUBLIC_API_URL contains "localhost", so _isProductionBuild is
+      # false, SSR renders dev defaults while the client renders production ones
+      # (hydration mismatch), and the "NEXT_PUBLIC_WS_URL not set" banner fires.
+      # Values match deploy-frontend.yml so both publishing lanes agree.
       - name: Build and push frontend
         uses: docker/build-push-action@v5
         with:
@@ -197,6 +205,9 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
           platforms: linux/amd64,linux/arm64
+          build-args: |
+            NEXT_PUBLIC_API_URL=https://api.aragora.ai
+            NEXT_PUBLIC_WS_URL=wss://api.aragora.ai
 
       - name: Build frontend for scanning
         run: |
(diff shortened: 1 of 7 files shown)


scarmani:

Head: 7e67e0d (7e67e0dd2c7a2093eaf4797ebb78edc3415d04c7), committed 2026-07-27T17:42:07Z.
PR: #9647.
Model family: claude

Verdict: PASS

Grounded the review by reading the full `config.ts` at head: the rescue predicate `!isLocalDevHostname(hostname)` is the exact complement of `isProductionEnvironment()`'s hostname branch (same five patterns: `localhost`, `127.0.0.1`, `192.168.*`, `10.*`, `*.local`), so whenever the rescue fires in-browser, `isProd` is guaranteed true and the fall-through always reaches `prodDefault(hostname)` — no path where a rescued value degrades to the localhost `devDefault`. SSR behavior is unchanged (rescue is gated on `inBrowser`), the mixed-bake guard correctly keys on serving host rather than `_isProductionBuild`, and the derived `ORACLE_WS_URL`/`PROMPT_ENGINE_WS_URL` inherit the correction. Dockerfile and workflow changes are comment-only with unchanged localhost ARG defaults. Test matrix covers prod-host rescue, self-hosted domain, mixed bake on localhost, local-dev verbatim, explicit-prod verbatim, and env restoration is handled in every file.

- [P3] `aragora/live/src/config.ts:160-166` — the rescue silently discards an explicitly configured localhost WS URL with no console.warn and no opt-out, so the niche-but-legitimate case (frontend served from a real domain, backend reached via local SSH tunnel) breaks undiagnosably. `resolveApiBaseUrl` has the same silent-rewrite behavior, so this is consistent, but a one-line `console.warn` on rescue would make the override observable.
- [P3] `aragora/live/src/config.ts:133-140` — `isLocalWsEndpoint` matches only `localhost`/`127.0.0.1`; a bundle baked with `ws://[::1]:8765/ws` or `ws://0.0.0.0:8765/ws` is not rescued. Consistent with the `isLocal` check in `resolveApiBaseUrl:109` and unreachable via the published image's defaults, so cosmetic.
- [P3] `aragora/live/src/config.ts:166-179` — for the rescued case SSR now returns the baked localhost value while the client returns `wss://api.<host>/ws`, a divergence the `_isProductionBuild` mechanism (config.ts:19-20) was explicitly built to avoid. Inert unless a component renders `WS_URL` into markup (sockets are client-only), and unavoidable for a host-agnostic image whose module-level constants can't see the request Host, but worth knowing if a hydration warning ever surfaces.

dogfood: yes

scarmani:

Head: 7e67e0d (7e67e0dd2c7a2093eaf4797ebb78edc3415d04c7), committed 2026-07-27T17:42:07Z.
PR: #9647.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

PR: #9647
Exact head: 7e67e0dd2c7a2093eaf4797ebb78edc3415d04c7
Authorized action: admin_squash_merge and branch_protection_reconcile, only if #9647 is non-draft and live exact-head checks/merge-packet remain otherwise green.

Issue #9640, opened 2026-07-27
merge-gate: 'blocked_by_live_gate' is unhandled by aragora-merge-quorum.yml — correct outcome, misleading 'Unrecognized status' error


scarmani:
> **Corrected 2026-07-27.** The original report claimed a self-sustaining deadlock permanently
> blocking Tier 0-2 PRs whose quorum was satisfied. **That was wrong.** See "Correction" below.
> The underlying defect is real but far narrower: the gate reaches the *correct* decision and
> reports it with a misleading error. Do not weaken the gate to "fix" this.

## Summary

When a merge-packet evaluates to `blocked_by_live_gate`, `.github/workflows/aragora-merge-quorum.yml`
has no branch for that status and hits its catch-all:

```
##[error]Unrecognized merge-quorum status 'blocked_by_live_gate' — failing closed.
```

Failing closed is the **right outcome**. The bug is purely diagnostic: the operator sees
"Unrecognized status" instead of the actual reason, which is available and specific.

## Correction — there is no deadlock

The original write-up asserted a circular failure (gate red → merge state non-CLEAN → live-gate
blocker → gate red). That is **not possible**:
`aragora/cli/commands/review_queue_unstable.py:446-455` treats **`BLOCKED` as an allowed**
merge state (`merge_state_status not in {"CLEAN", "BLOCKED"}`). A failing required check produces
`BLOCKED`, which is permitted — so the gate's own failure cannot feed back as a live-gate blocker.

On **#9012**, the actual blocker is the first branch of `admin_squash_live_gate_blockers`:

```python
if backend.OPERATOR_REVIEW_REQUIRED_LABEL in labels:
    blockers.append("operator-review-required label present")
```

#9012 carries the **`operator-review-required`** label. A human deliberately held it. The gate
correctly refuses to authorize the merge; it just fails to *say so*. Removing the label clears it.

## Scope

One ready PR is affected:

```bash
gh pr list --state open --draft=false --limit 200 --json number,labels \
  --jq '[.[]|select(.labels|any(.name=="operator-review-required"))|.number]'
# => [9012]
```

This is **not** a backlog-wide blocker. (The separate, real backlog constraint is that ~23 ready
PRs have green required checks and are MERGEABLE but sit at `signals=0` — no quorum evidence
collected. That is unrelated to this issue.)

## Recommended fix

**Add an explicit `blocked_by_live_gate` branch that still fails, but prints the blockers.**
[...]

PR #9012, opened 2026-07-08
docs(m6): fix scrutiny blockers - stale STATUS.md claim, missing archive stubs, positioning drift

an0mium:
## Summary

Fixes the 3 blocking findings from the m6-docs scrutiny round-1 synthesis
(`validation/m6-docs/scrutiny/synthesis.json`). Composed against a fresh
`origin/main` — **PR #9001 (archive collapse) merged mid-session** (squash
commit `2caed5d61f`, auto-merge fired at `2026-07-08T14:45:50Z`), so this PR
picks up from that state rather than redoing it. **PR #9003 (canonical
consistency sweep) was already merged** before this session started
(`2543ae8b42`) and is unrelated to the remaining findings below.

## Findings fixed

**1. VAL-DOCS-010 — undated stale-version claims in `docs/status/STATUS.md`**
`origin/main` already had line 184 (`v2.8.0` → `v2.9.0`) fixed by #9001's
last commit. The one remaining gap was line 389, inside the dated
`## Phase 10 ... (February 15, 2026)` historical section:
`- **aragora-debate package**: v0.2.0 built and validated via twine check`.
Since this sits inside genuine dated history (the twine check really did run
against 0.2.0 in that sprint), I did **not** rewrite it to claim a version it
never ran against — I appended a `(historical, current: 0.2.3)` clarifier,
per the feature contract's guidance for the dated-history case. Swept the
whole file afterward; no other undated `2.8.0`/`0.2.0` current-claims remain.

**2. Missing redirect stubs at the 3 paths #9001 archived**
#9001's merged diff moved 3 docs into `docs/archive/` (with correct disposition
rows already added to `docs/archive/README.md`) but left **no stub** at any of
the 3 original paths, reasoning "zero inbound references" in each case. That's
exactly the blocking pattern the scrutiny round flagged (stubs are for external
bookmarks/search-engine links too, not just internal doc-graph integrity).
Added ~15-line redirect stubs at all 3 original paths, mirroring the
`docs/guides/GETTING_STARTED.md` pattern, each linking to (a) the maintained
live replacement and (b) the `docs/archive/` snapshot (using the exact archive
filenames/replacement targets #9001 already recorded in `docs/archive/README.md`):

| Original path | Archive snapshot | Live replacement |
|---|---|---|
[...]

PR #9650, opened 2026-07-27
fix(merge-gate): handle blocked_by_live_gate + settled instead of failing as unrecognized

scarmani:
## Summary

The merge-quorum evaluator emits status strings the gate workflow never learned, so they fall
through to the catch-all:

```
##[error]Unrecognized merge-quorum status 'blocked_by_live_gate' — failing closed.
```

`aragora-merge-quorum` is a **required** check, so the operator sees a parser error instead of the
real reason — which the packet already knows. Two statuses were affected:

| Status | Emitted when | Now |
|---|---|---|
| `blocked_by_live_gate` (`review_queue.py:3057`) | quorum **is** satisfied but the live gate withheld admin squash — `operator-review-required` label, or mergeStateStatus outside `{CLEAN, BLOCKED}` | **still fails**, now naming the blockers |
| `settled` (`review_queue.py:3697`) | PR already merged, settlement receipt recorded | exits 0 — nothing left to gate |

Observed on **#9012**, which carries `operator-review-required`. `settled` was found while
enumerating the vocabulary for the drift test; not yet seen in the wild.

## `blocked_by_live_gate` still fails — deliberately

This is the whole point of the fix's shape. That status is emitted **only when the model quorum is
satisfied**, so the tempting repair — "the quorum is fine, let it pass" — would merge away a hold
an operator applied on purpose. Only the message changes; the branch renders
`admin_squash_gate_blockers`, which is already on the entry.

The test suite pins this: `test_blocked_by_live_gate_still_fails_closed` asserts the branch
contains `fail(` and does not `sys.exit(0)` before it, so a later "simplification" cannot quietly
convert an operator hold into a pass.

## Not a deadlock — correcting the record

An earlier draft of #9640 (mine) claimed a self-sustaining circular failure permanently blocking
Tier 0-2 PRs whose quorum was satisfied. **That was wrong**, and the "preferred fix" it proposed
would have weakened the gate. `review_queue_unstable.py:451` treats `BLOCKED` as an *allowed*
merge state:

```python
elif merge_state_status not in {"CLEAN", "BLOCKED"}:
```

A failing required check produces `BLOCKED`, which is permitted — so the gate's own red cannot feed
back as its own blocker. Exactly **one** ready PR is affected, and it is blocked because a human
held it. #9640 has been corrected.

## The drift guard

`tests/ci/test_merge_quorum_status_vocabulary.py` extracts the statuses the evaluator can assign
and the statuses the workflow dispatches on, then asserts the first set is a subset of the second.

Nothing coupled these two files before — that is why the drift was invisible until a PR happened to
hit it, and why enumerating the vocabulary immediately turned up a *second* unhandled status. Any
future status now fails a cheap unit test instead of a required check with unactionable text.

## Reviewed design tradeoffs

- **Rejected: keep `status = "satisfied"` and treat the live gate as advisory.** Removes an
  intentional `operator-review-required` hold. This was my original recommendation and it was wrong.
- **Rejected: exit 0 when the underlying quorum is satisfied.** Same defect, same reason.
- **Rejected: import a shared constant into the workflow.** The workflow's embedded Python is
  deliberately dependency-light (`os`, `sys`, `subprocess`, `json`); importing `aragora` there would
  couple the gate to package import health. A source-extracting test gets the same protection with
  no runtime coupling.
- **Accepted brittleness:** the guard regex-extracts from source, so a large refactor of either file
  could make it vacuously pass. Mitigated by `test_producer_and_workflow_are_both_readable`, which
  fails if either extraction returns an empty set.

## Verification

```
4/4 new tests pass
drift genuinely caught: removing the `settled` branch fails with
  "status(es) emitted by review_queue.py but not handled ... ['settled']"
  restoring it passes again
workflow YAML parses; embedded python block (123 lines) passes ast.parse
```

No production code changed — workflow + new test only.

Refs #9640.

### DIFF
--- a/.github/workflows/aragora-merge-quorum.yml
+++ b/.github/workflows/aragora-merge-quorum.yml
@@ -265,10 +265,10 @@ jobs:
                   "status. See docs/governance/MERGE_GATE_RECONCILIATION.md."
               )
 
-          # The PR is already merged and its settlement receipt is recorded. There is
-          # nothing left to authorize, so do not fail closed on it.
-          if status == "settled":
-              print(f"PR #{pr_number} is already merged with settlement recorded; nothing to gate.")
+          # The PR is already merged ('settled' additionally has a settlement receipt
+          # recorded). There is nothing left to authorize, so do not fail closed on it.
+          if status in ("settled", "already_merged"):
+              print(f"PR #{pr_number} is already merged (status={status}); nothing left to gate.")
               sys.exit(0)
 
           # The live gate withheld admin-squash authorization even though the model
--- a/.github/workflows/aragora-merge-quorum.yml
+++ b/.github/workflows/aragora-merge-quorum.yml
@@ -265,6 +265,29 @@ jobs:
                   "status. See docs/governance/MERGE_GATE_RECONCILIATION.md."
               )
 
+          # The PR is already merged and its settlement receipt is recorded. There is
+          # nothing left to authorize, so do not fail closed on it.
+          if status == "settled":
+              print(f"PR #{pr_number} is already merged with settlement recorded; nothing to gate.")
+              sys.exit(0)
+
+          # The live gate withheld admin-squash authorization even though the model
+          # quorum itself is satisfied — e.g. an 'operator-review-required' label, or
+          # a mergeStateStatus outside {CLEAN, BLOCKED}. Failing closed is correct: an
+          # operator hold must never be merged away. Only the reason needs surfacing,
+          # and the packet already carries it.
+          if status == "blocked_by_live_gate":
+              gate_blockers = entry.get("admin_squash_gate_blockers") or []
+              for blocker in gate_blockers:
+                  print(f"  - {blocker}")
+              detail = "; ".join(str(b) for b in gate_blockers) or "no blocker detail in packet"
+              fail(
+                  f"Tier {tier}: model quorum is satisfied, but the live gate withheld "
+                  f"admin-squash authorization ({detail}). Clear the blockers above — for "
+                  "an 'operator-review-required' label, that means an operator removing it "
+                  "— then re-run this check."
+              )
+
           fail(f"Unrecognized merge-quorum status '{status}' — failing closed.")
           PY
 
--- a/.github/workflows/aragora-merge-quorum.yml
+++ b/.github/workflows/aragora-merge-quorum.yml
@@ -228,6 +228,15 @@ jobs:
               # The statuses API returns most-recent first.
               return bool(states) and states[0] == "success"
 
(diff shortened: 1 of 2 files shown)


scarmani:

Head: c6d585a (c6d585a47863d416611aebed00e7fb1d39196d05), committed 2026-07-28T03:00:03Z.
PR: #9650.
Model family: claude

Verdict: PASS

Verified against the actual producer (`aragora/cli/commands/review_queue.py`) and base workflow, not just the diff text:

- The exit-0 short-circuit is safe: `settled` requires `pr_state == "MERGED"` (review_queue.py:2807) and the synthetic `already_merged` entry is only built when GitHub reports the PR merged (review_queue.py:3180-3182), so neither status can pass the required check on an open PR.
- Placement is correct: the merged check lands before the dissent guard, which tests the `unresolved_dissent` *field* — and the producer's `settled` branch (review_queue.py:3697) indeed does not clear that field, so the ordering matters and is pinned by a test.
- `blocked_by_live_gate` still fails closed; it can only be emitted when quorum was satisfied (review_queue.py:3052-3058), so it can't shadow the dissent guard. `admin_squash_gate_blockers` exists in every entry.
- Produced-vocabulary sweep (all 13 emission sites) matches the handled set exactly; `tests/ci/` is an existing collected package.

Findings:

- [P3] `tests/ci/test_merge_quorum_status_vocabulary.py:33-38` — `_PRODUCED` scans the entire 3,700-line producer for *any* `status = "literal"` assignment. A future unrelated local variable named `status` (e.g., an HTTP or git status string) enters the produced set and spuriously fails the vocabulary test. It fails in the safe/loud direction, but consider anchoring the scan to the packet-builder function or a shared constant.
- [P3] `tests/ci/test_merge_quorum_status_vocabulary.py:110-118` — in `test_blocked_by_live_gate_still_fails_closed`, the `"fail(" in branch` assertion is trivially satisfied by the catch-all `fail(` inside the 1200-char slice even if the branch's own `fail` were removed; only the second assertion does real work. Narrow the slice to the branch body (e.g., up to the next dedent) to make the first assertion meaningful.
- [P3] `.github/workflows/aragora-merge-quorum.yml` (new `blocked_by_live_gate` branch) — packet-derived blocker strings are interpolated into `::error::` via `fail(detail)`. Content is producer-generated but embeds label names; a crafted label could at worst forge log annotations (no code execution, collaborator-gated). Marginal; sanitizing newlines out of `detail` would close it.

dogfood: yes

scarmani:

Head: c6d585a (c6d585a47863d416611aebed00e7fb1d39196d05), committed 2026-07-28T03:00:03Z.
PR: #9650.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9652, opened 2026-07-27
fix(reasoning): measure crux disagreement from priors so dissent is attributable (#9644)

### LINKED ISSUE #9644: Crux cards never attribute dissent: detect_cruxes propagates before measuring disagreement
## Summary

`CruxAnalysisResult` items come back with `contesting_agents: []` even when a
crux is genuinely contested and `total_disagreements > 0`. Found while fixing
#9581 (PR #9643); deliberately left out of that PR because the fix changes what
"disagreement" means in the detector.

## Root cause

`CruxDetector.detect_cruxes()` runs:

```python
self.network.propagate()              # ~line 335
...
disagreement_data = self.compute_disagreement_scores()   # ~line 339
```

Belief propagation exists to converge the network toward a consistent posterior.
`compute_disagreement_scores()` then measures **variance across authors' implied
beliefs** — but propagation has already flattened exactly that variance. The
threshold for naming a contesting agent is `abs(author_mean - mean) > 0.2`, and
post-convergence no author clears it.

Net effect: a disagreement is *counted* (the ≥2-authors branch is entered) but
never *attributed*, so a crux card cannot say who contests what — which is most
of its value in a decision receipt.

## Reproduction

With PR #9643 applied, on a real 2-agent contested debate:

- `compute_disagreement_scores()` called directly on the pre-propagation network
  returns `disagreement=0.600, contesting=['claude', 'codex']`.
- The same network through `detect_cruxes()` returns items with
  `contesting_agents: []`.

The only difference is the intervening `propagate()`.

## Fix direction

Measure author disagreement from what each author **asserted** (priors), not
from the network's **converged** posteriors — either by computing disagreement
before `propagate()`, or by having `compute_disagreement_scores()` read priors.
Post-convergence residual is a different quantity and is not what a crux card
reports.

Note this also affects `consensus="crux_finder"` (`crux_mode.py`), so it needs
its own review rather than riding along with an edge-construction fix.

## Acceptance

A standard 2-agent debate on a contested question yields a receipt whose
`cruxes` block has ≥1 item with **non-empty `contesting_agents`** — the
criterion originally stated in #9581 and not met by #9643.

scarmani:
## Summary

Fixes #9644 — and with it, the acceptance criterion originally written into
#9581 that #9643 could not meet.

Crux cards counted disagreements but never named **who** disagreed:
`contesting_agents` came back empty on every card. In a `DecisionReceipt` that
field *is* the dissent attribution, so a card without it says a disagreement
existed while withholding the part an auditor needs.

## Three causes, fixed together

None of them works alone, which is why they were deferred out of #9643 rather
than patched under it — incremental patches there drew findings in three
consecutive review rounds.

**1. Disagreement was read from post-propagation posteriors.** `detect_cruxes`
runs `network.propagate()` first, and must — influence scoring needs a converged
network. But propagation exists to pull the network toward a consistent joint
belief, so measuring variance afterwards measures whatever survived that
reconciliation, which for a genuinely contested claim tends toward zero. Priors
are what each author put on the table and do not move.

**2. The claim's own author was absent from the comparison.** The map held only
*neighbours*, so the score measured how far contesters sat from each other
rather than whether anyone contested the author at all — two agents both
contradicting a claim scored zero variance. The author is now seeded, and
**excluded from the reported `contesting_agents`**: naming a proposer as a
dissenter in receipt attribution would be a false attribution.

**3. Asserted claims carried a 0.5 prior.** `add_claim` defaults to maximum
entropy — "no opinion" — recording an author as neutral about their own claim,
leaving no variance to find. Debate claims are now asserted at 0.8: a position,
not a proof.

## Result

A contested **two-agent** debate now produces cards with named contesters:

```
items=4  disagreements=4
  contested_by=['codex']   :: Counting overstates proof.
  contested_by=['claude']  :: Failing discards evidence.
```

**This removes the 3+ agent requirement** that #9643 documented and that the W3
bundle currently states. The regression test pinning the old limitation is
updated to pin the new capability, and the bundle's crux section needs a
corresponding correction — I'll follow up rather than let that text go stale.

## Testing

- 1257 crux/reasoning tests pass.
- The un-gated consumers this scoring change reaches are covered too: **170**
  KM-adapter/belief tests and **89** belief-handler tests, all passing.
- `test_disagreement_reads_priors_not_post_propagation_posteriors` **fails
  against `main`** and passes here — it propagates first, exactly as
  `detect_cruxes` does, then asserts the contester is still named.

## Reviewed design tradeoffs

- **Priors vs. reordering the pipeline:** computing disagreement before
  `propagate()` would also work, but influence and resolution-impact scoring
  genuinely need the converged network, so reordering would split the pipeline
  for one consumer. Reading priors is local and says what it means.
- **Seeding the author vs. lowering the ≥2-author threshold:** the threshold is
  the right check; the bug was that one side of every disagreement was missing
  from what it counted.
- **0.8 rather than higher:** an asserted debate claim is a position, not a
  proof, and a near-1.0 prior would let a single critique dominate scoring.

Closes #9644.

## Behaviour change worth noting for anyone comparing stored scores

Raised in review and recorded here rather than left to be discovered:

- **`crux_score` levels shift in the message-built (fallback) path.** Asserted
  claims now carry a 0.8 prior instead of `add_claim`'s 0.5 default, which lowers
  the pre-propagation uncertainty component (entropy 1.0 → ≈0.72). Rankings are
  unaffected in shape, but absolute scores stored by earlier releases are not
  directly comparable. This also slightly strengthens the
  no-disagreement suppression guard, which is intended.
- **Repeat objections from the *same* author take the max, not the sum.**
  Cross-author objections sum (so a wider dispute scores higher); a single critic
  escalating across rounds does not inflate their own weight. `_link_critiques`
  deliberately builds one edge per round, so this divergence is intentional —
  the docstring's "summed contest strength" is imprecise on that point and is
  tracked for a follow-up along with whether the contested-share denominator
  should include the claim's own author as a stance-taker.

### DIFF
--- a/aragora/debate/crux_cards.py
+++ b/aragora/debate/crux_cards.py
@@ -84,10 +84,9 @@ def build_crux_cards(
         # disagreement actually detected behind them — an overclaim in exactly
         # the surface auditors read. Absent is honest; mislabelled is not.
         #
-        # `CruxDetector` registers a disagreement only when >=2 authors *other
-        # than the claim's own* relate to it, so a two-agent debate trading
-        # reciprocal critiques never reaches one. Making a two-agent debate
-        # attribute dissent is #9644's scope, not this edge-construction fix.
+        # Reached when nothing was contested at all: since #9644 a single
+        # CONTRADICTS edge registers a disagreement and names its contester, so
+        # any real critique clears this guard.
         logger.info("crux_cards_suppressed_no_disagreement claims=%d", analysis.total_claims)
         return None
 
--- a/aragora/debate/crux_cards.py
+++ b/aragora/debate/crux_cards.py
@@ -21,12 +21,12 @@ logger = logging.getLogger(__name__)
 
 CRUX_CARDS_METADATA_KEY = "crux_cards"
 
-# An agent that states a claim in a debate is asserting it. ``add_claim`` defaults
-# to 0.5 — maximum entropy, i.e. "no opinion" — which recorded every author as
-# neutral about their own claim, leaving no variance for
-# ``CruxDetector.compute_disagreement_scores`` to measure and an empty
-# ``contesting_agents`` on every card (#9644). Deliberately short of certainty: an
-# asserted debate claim is a position, not a proof.
+# An agent that states a claim in a debate is asserting it, so it is not recorded
+# at ``add_claim``'s 0.5 default — maximum entropy, i.e. "no opinion" about its
+# own claim. Dissent attribution no longer reads belief values at all (#9644
+# derives it from edge polarity), so this now feeds only the uncertainty/entropy
+# component of ``crux_score`` and propagation dynamics. Deliberately short of
+# certainty: an asserted debate claim is a position, not a proof.
 _ASSERTED_CONFIDENCE = 0.8
 
 
--- a/aragora/debate/crux_cards.py
+++ b/aragora/debate/crux_cards.py
@@ -21,6 +21,14 @@ logger = logging.getLogger(__name__)
 
 CRUX_CARDS_METADATA_KEY = "crux_cards"
 
+# An agent that states a claim in a debate is asserting it. ``add_claim`` defaults
+# to 0.5 — maximum entropy, i.e. "no opinion" — which recorded every author as
+# neutral about their own claim, leaving no variance for
+# ``CruxDetector.compute_disagreement_scores`` to measure and an empty
+# ``contesting_agents`` on every card (#9644). Deliberately short of certainty: an
+# asserted debate claim is a position, not a proof.
+_ASSERTED_CONFIDENCE = 0.8
+
 
 def build_crux_cards(
     *,
@@ -108,6 +116,7 @@ def _network_from_messages(messages: list[Any], critiques: list[Any] | None = No
                 claim_id=f"msg_{i}_{getattr(msg, 'agent', 'unknown')}",
                 statement=str(getattr(msg, "content", ""))[:500],
                 author=str(getattr(msg, "agent", "unknown")),
+                initial_confidence=_ASSERTED_CONFIDENCE,
             )
             added += 1
     if not added:
--- a/aragora/reasoning/crux_detector.py
+++ b/aragora/reasoning/crux_detector.py
@@ -215,10 +215,17 @@ class CruxDetector:
         their own claim: this list becomes ``DecisionReceipt`` dissent
         attribution, where either would be a false statement about a person.
 
-        Magnitude is the mean contest strength (``factor.strength``, carrying
-        critique severity) scaled by the contested share of related authors, so a
-        severity-9 objection outweighs a nitpick and a claim with three
+        Magnitude is the summed contest strength (``factor.strength``, carrying
+        critique severity), capped at 1.0 and scaled by the contested share of
+        the authors who took a *stance* on the claim. So a severity-9 objection
+        outweighs a nitpick, two objections outweigh one, and a claim with three
         supporters and one contester scores below one contested by three.
+
+        Only ``SUPPORTS``/``CONTRADICTS`` authors count toward that share.
+        Non-stance relations (``REFINES``, ``QUALIFIES``, ``DEPENDS_ON``) are not
+        treated as either agreement or dissent — silence about a claim is not
+        support for it. Whether a qualification is *partial* dissent is a
+        modelling question worth its own change.
         """
         from aragora.reasoning.claims import RelationType
 
@@ -254,10 +261,15 @@ class CruxDetector:
                 disagreement_scores[node_id] = (0.0, [])
                 continue
 
-            related = len(contesters) + len(supporters - set(contesters))
-            contested_share = len(contesters) / related if related else 1.0
-            mean_strength = sum(contesters.values()) / len(contesters)
-            disagreement = min(1.0, max(0.0, mean_strength * contested_share))
+            # Capped sum, not mean: averaging made a second, weaker contester
+            # LOWER the score (0.7 alone -> 0.7; 0.7 plus a 0.1 nitpick -> 0.4),
+            # so a more widely contested claim could rank as less contested.
+            # This is monotonic in both the strength of objections and the
+            # number of distinct agents raising them.
+            contest_weight = min(1.0, sum(contesters.values()))
+            stance_authors = len(contesters) + len(supporters - set(contesters))
+            contested_share = len(contesters) / stance_authors if stance_authors else 1.0
+            disagreement = min(1.0, max(0.0, contest_weight * contested_share))
             disagreement_scores[node_id] = (disagreement, sorted(contesters))
 
         return disagreement_scores
--- a/aragora/reasoning/crux_detector.py
+++ b/aragora/reasoning/crux_detector.py
@@ -240,6 +240,12 @@ class CruxDetector:
 
                 if factor.relation_type == RelationType.CONTRADICTS:
                     strength = float(getattr(factor, "strength", 1.0) or 0.0)
+                    if strength <= 0:
+                        # A zero-weight contradiction would name a contester on a
+                        # claim scoring 0.0 — listed as contested by someone while
+                        # the same result excludes it from `total_disagreements`.
+                        # Keep the list and the count telling one story.
+                        continue
                     contesters[source.author] = max(contesters.get(source.author, 0.0), strength)
                 elif factor.relation_type == RelationType.SUPPORTS:
                     supporters.add(source.author)
--- a/aragora/reasoning/crux_detector.py
+++ b/aragora/reasoning/crux_detector.py
@@ -196,16 +196,36 @@ class CruxDetector:
         """
         Compute disagreement scores for all claims.
 
-        Disagreement = variance in how different agents' claims affect this belief.
-        Also returns list of contesting agents.
+        Disagreement = variance across what each author *asserted* about a claim.
+        Also returns the list of contesting agents.
+
+        Read from **priors, not posteriors** (#9644). ``detect_cruxes`` runs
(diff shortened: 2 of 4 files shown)


scarmani:
Implementation handoff from an independently validated #9644 worktree (not quorum evidence or settlement). Two current-head correctness blockers remain at `a873ab5877b69b268299c9502e133b699280c0a0`:

- [P1] `compute_disagreement_scores()` still treats factors as undirected because it selects the opposite endpoint for every `node_factors` entry. A single `critique -> proposal` edge therefore reports both the proposal *and the critique text itself* as contested. Exact-head reproduction: proposal `(0.6000000000000001, ['critic'])`; critique `(0.6000000000000001, ['proposer'])`. Only factors whose `target_node_id == node_id` describe incoming positions about the current claim; use `source_node_id` for those and skip outgoing factors. This is the double-counting/source-card defect already identified during #9643.

- [P1] `contesting_agents` is still compared with the all-author mean. With one proposer at 0.8 and two agreeing contradictors at implied 0.2, the mean is 0.4 and each critic is exactly 0.2 away, so strict `> 0.2` returns no contesters: `(0.5656854249492381, [])`. Compare each non-author position with the claim author's assertion-time prior instead; otherwise a three-agent contested debate can again emit disagreement without attribution.

Focused regressions that caught these: (1) an outgoing contradiction factor must leave the source claim at `(0.0, [])`; (2) two agreeing critics must both remain attributed; (3) scores/attribution are identical before and after propagation. The overlapping local branch will not be published while this PR owns #9644.

scarmani:

Head: 8c93de0 (8c93de06e9f31af83fc2541f742317545a062c24), committed 2026-07-27T17:59:00Z.
PR: #9652.
Model family: claude

Verdict: PASS

Grounded verification performed at head `8c93de0` (fetched `pull/9652/head`, ran in an isolated worktree, since removed): all 117 tests in the three touched/adjacent test files pass, plus 204 tests across downstream consumers (`test_crux_mode.py`, `test_cruxset_emission.py`, `test_crux_handler.py`, `test_belief.py`, `test_winner_selector.py`). Core claims of the diff check out against the unchanged code: `node_factors` indexes every factor under both source and target (`belief.py:579-580`), so filtering to `factor.target_node_id == node_id` correctly captures all incoming edges; `add_claim` accepts `initial_confidence` (`belief.py:535`); `math` remains used (`log2` at `crux_detector.py:338,356,423`) so no dead import; the source-contests-target direction convention is confirmed repo-wide (`claims.py:498,531`, `consensus_phase.py:1481-1487`, `crux_cards.py:250`); self-authored and unattributed sources are excluded; output is deterministic via `sorted()`. Pre-existing permissive tests of the old variance semantics (`test_crux_detection.py:146`, `test_crux_detector.py:405-445`) still pass unmodified.

- [P3] `aragora/debate/crux_cards.py:86-90` (head) — the suppression guard's second comment paragraph is now false: it states "`CruxDetector` registers a disagreement only when >=2 authors other than the claim's own relate to it, so a two-agent debate... never reaches one. Making a two-agent debate attribute dissent is #9644's scope, not this edge-construction fix." This PR *is* the #9644 fix and a single CONTRADICTS edge now registers disagreement — the very behavior `test_two_agent_debate_now_attributes_dissent` asserts. Comment rot in a load-bearing explanation touched by this PR; trim the stale paragraph here.
- [P3] `aragora/reasoning/crux_detector.py` (new `compute_disagreement_scores`) — non-stance relations (REFINES, QUALIFIES, ELABORATES, DEPENDS_ON) no longer count toward `related`, so 1 contester + 3 QUALIFIES authors yields `contested_share=1.0` where the old code treated them as 0.5-neutral participants. Defensible (silence ≠ support), but QUALIFIES is "adds conditions/exceptions" per its own docstring — arguably partial dissent now rendered invisible. Observation only.
- [P3] `aragora/reasoning/crux_detector.py` (new loop) — a CONTRADICTS factor with `strength` 0/None produces a named contester with `disagreement=0.0`: the author appears in `contesting_agents` while the claim is excluded from `total_disagreements` (`disagreement > 0` gate at line 341). Unreachable via `build_crux_cards` (`_link_critiques` skips `severity <= 0`) but possible for direct `BeliefNetwork` users; a `strength > 0` guard would make list and count consistent.

dogfood: yes

scarmani:

Head: 8c93de0 (8c93de06e9f31af83fc2541f742317545a062c24), committed 2026-07-27T17:59:00Z.
PR: #9652.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

Head: 31373a6 (31373a68a9366cba2ba6a990b9995eae68497064), committed 2026-07-27T18:09:36Z.
PR: #9652.
Model family: claude

Verdict: PASS

Verified against head 31373a6 (fetched PR ref, ran in isolated worktree): all 118 tests pass across `tests/debate/test_crux_cards.py`, `tests/reasoning/test_crux_detector.py`, and the **untouched** legacy `tests/reasoning/test_crux_detection.py` (its assertions are type-level, so the semantic rewrite doesn't regress it). Cross-checked the API surface the diff relies on: `add_claim(initial_confidence=...)` (belief.py:535), `Factor.strength` (belief.py:278), and `node_factors` indexing each factor under **both** endpoints (belief.py:579-580) — which makes the new `factor.target_node_id != node_id` incoming-edge filter both necessary and correct; the old symmetric traversal really did invert attribution. `math` remains used (`log2`), sole other caller `detect_cruxes` consumes the node-keyed dict via `.get` with a safe default, and the zero-strength skip keeps `contesting_agents` consistent with `total_disagreements`. No security surface touched.

- [P3] `aragora/reasoning/crux_detector.py` (~line 245, `mean_strength`): averaging contest strengths means adding a *weak* second contester lowers the score below the strong contester alone (0.7 alone → 0.7; 0.7 + 0.1 nitpick → mean 0.4, share 1.0 → 0.4). More dissenting agents can rank a claim *less* contested — `max` or a capped sum would be monotonic in contesters; the tests only compare single-contester pairs so this shape is unpinned.
- [P3] `aragora/debate/crux_cards.py:24-31`: `_ASSERTED_CONFIDENCE`'s justification is the variance mechanism this same PR deletes — post-merge, 0.8 is load-bearing only for uncertainty/entropy and propagation dynamics, not for `contesting_agents`. Comment is past-tense so technically accurate, but it will read as the constant's live purpose.
- [P3] `aragora/reasoning/crux_detector.py` (~line 243, `related`): docstring says "contested share of related authors," but the denominator counts only SUPPORTS/CONTRADICTS authors — REFINES/DEPENDS_ON/QUALIFIES neighbors (present on the supplied KM-seeded network path, where they previously contributed a neutral 0.5 vote to the variance) are now silently invisible to disagreement. Reasonable semantics for dissent attribution, but the wording overstates coverage.

dogfood: yes

scarmani:

Head: 31373a6 (31373a68a9366cba2ba6a990b9995eae68497064), committed 2026-07-27T18:09:36Z.
PR: #9652.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

scarmani:

Head: 01aa5dd (01aa5ddbc0a118c57e47ec95c837ceb854bbc4d5), committed 2026-07-27T18:18:39Z.
PR: #9652.
Model family: claude

Verdict: PASS

Grounding performed (not just diff-read): fetched head 01aa5dd, verified `BeliefNetwork.node_factors` registers each factor under **both** endpoints (`belief.py:579-580`) so the new incoming-edge filter sees all incoming edges; verified `add_claim(initial_confidence=...)` and the positional `add_factor` signature match test usage; confirmed `math` is still used elsewhere in `crux_detector.py` (no dead import); confirmed `compute_disagreement_scores` has exactly one caller (`detect_cruxes`), so the rewrite is contained and the return shape is unchanged. Ran both changed test suites at head (100 passed) plus 7 downstream crux-dependent suites — `test_crux_detection`, `test_crux_mode`, `test_winner_selector`, `test_crux_handler`, `test_belief`, `test_cruxset_emission`, `test_missing_emissions` (236 passed). Checked the algebra: capped-sum × contested-share is monotone in contester count/strength, supporters damp, zero-strength skip keeps the contester list consistent with `total_disagreements` (counts `disagreement > 0`), edge direction excludes inverted and self-attribution — matching the critic→proposal edge orientation built in `_link_critiques`.

- [P3] `aragora/reasoning/crux_detector.py` (~L246): repeat objections from the *same* author take `max()`, but the docstring says "summed contest strength" — a critic escalating severity-3 across three rounds registers 0.3, not 0.9, while `crux_cards._link_critiques` deliberately builds those parallel per-round edges. The max is a defensible anti-inflation choice, but the docstring wording overstates and the same-author multi-round case deserves a one-line note so the divergence reads as intended.
- [P3] `aragora/debate/crux_cards.py:29`: `_ASSERTED_CONFIDENCE = 0.8` uniformly lowers the pre-propagation uncertainty component (entropy 1.0 → ≈0.72) for every message-built network, silently shifting `crux_score` rankings versus prior releases in the fallback path. Intended per the comment (and it strengthens the suppression guard), but any consumer comparing stored crux scores across versions will see an unexplained level shift — worth a line in the PR description/changelog.

dogfood: yes

scarmani:

Head: 01aa5dd (01aa5ddbc0a118c57e47ec95c837ceb854bbc4d5), committed 2026-07-27T18:18:39Z.
PR: #9652.
Model family: openai

Verdict: PASS
- [P3] `aragora/reasoning/crux_detector.py:265` The contested-share denominator omits the target claim’s own author, even though the PR now treats authored claims as asserted positions. A two-agent claim with one contradiction is scored as 100% contested rather than 1 of 2 stance-taking authors, inflating disagreement/convergence-barrier ranking for lightly contested claims.

dogfood: yes

Issue #6474, opened 2026-04-23
runners: fleet census — 2 Mac runners missing vs memory record (10/12)


an0mium:
## Summary

Live GitHub runner API shows **10 self-hosted runners online, 0 offline**. Memory record (``~/.claude/projects/-Users-armand-Development-aragora/memory/MEMORY.md``) states "Runner Fleet: 12 total (3 Hetzner + 6 EC2 + 3 Mac)". Two Mac runners are missing — they aren't simply offline (which would show in the API), so either their registrations were removed, their installations were wiped, or the memory record was wrong.

## Live fleet (2026-04-23T16:43Z)

| Pool | Expected (memory) | Actual | Delta |
|---|---|---|---|
| Hetzner (Linux x64) | 3 | 3 | ✓ |
| EC2 (Linux x64) | 6 | 6 | ✓ |
| Mac (macOS ARM64) | 3 | **1** | **−2** |
| **Total** | 12 | 10 | **−2** |

Currently registered Macs: ``mac-studio-m3ultra`` (labels ``self-hosted, aragora, macOS, ARM64, mac-studio``).

## LAN-scan findings

mDNS/Bonjour browse of ``_companion-link._tcp`` + ``_ssh._tcp`` and reverse mDNS against the ARP table identified **5 Macs on 10.0.0.0/24**:

| IP | mDNS name | GH runner status | Notes |
|---|---|---|---|
| 10.0.0.50 | ``MacBook-Pro-5.local`` | (none) | Founder laptop — confirmed no actions-runner install |
| 10.0.0.62 | ``Mac-84`` | (none) | Unknown Mac model |
| 10.0.0.90 | ``Mac-Studio.local`` | **``mac-studio-m3ultra``** | Online, idle |
| 10.0.0.170 | ``MacBook-Pro16GB.local`` | (none) | Named suggests a 16 GB MacBook Pro — possibly a previously-registered runner |
| 10.0.0.193 | ``MacBook-Pro-3.local`` | (none) | Possibly a previously-registered runner |

So **3 additional Macs are reachable on the LAN** (Mac-84, MacBook-Pro16GB, MacBook-Pro-3) that are NOT running registered actions-runners. Two of them are likely the missing fleet members per the memory record.

## Possible causes

1. **Actions-runner services failed** on the two Macs — LaunchAgents stopped or crashed; would still show ``offline`` in GH API (they don't, so likely not this).
[...]

PR #9654, opened 2026-07-27
ci(runners): make the Mac runner health poll actually run, and check disk

scarmani:
## Summary

Two defects in `mac-runner-health-poll.yml`, found while root-causing why every CI job on `mac-studio-m3ultra` had been failing since ~2026-07-09.

**1. The poll has never run.** It is gated on an SSH key built from `MAC_RUNNER_SSH_KEY`. That secret has never been set, so the step was skipped and the job exited 0. Verified on every scheduled run back through 2026-07-06 (runs `30278119126`, `29752249342`, `29260594691`, `28801732723`) — all four log `MAC_RUNNER_SSH_KEY secret not set; poll will be skipped` and report success.

**2. It never checked disk.** Even once running, it only looked at TIME_WAIT and health-log freshness.

The host was found 100% full — **132 MiB free of 926 GiB** — after an unrotated third-party log grew to **474 GB**. The weekly poll was green throughout.

## Why this was hard to see

A full disk also stops the runner uploading its own job logs, so affected jobs went red with `BlobNotFound` / HTTP 404 and no error text. And `npm ci` failed *only* on branches that changed a lockfile: an unchanged lockfile re-extracts bytes it just freed from the warm machine-local `~/.npm/_cacache`, while a changed one needs net-new bytes. That made a disk problem present as a registry problem.

Audit of the 9 mac-studio reds since Jul 9: **5 disk** (Jul 12, 15, 24, and two on Jul 27 — both re-run green after the disk was freed), **4 Codex-runner** (Jul 9, 10, 26, 27), which is a separate open issue not addressed here.

## Changes

- Fail with `::error::` and exit 1 when `MAC_RUNNER_SSH_KEY` is absent, instead of warning and passing.
- Drop the now-redundant `KEY_READY` guards so a skipped poll cannot leave the job green.
- Report "healthy" only on an explicit `drift == '0'` — the old `!= '1'` was also satisfied by an empty output from a step that never ran.
- Add a free-space check with `DISK_MIN_FREE_GB` (default 50), alerting per host.
- Alert body carries recovery steps, including reading missing job output from `~/actions-runner/_diag/pages/` and truncating in place.

## Heads up: this will go red

The weekly run will fail until `MAC_RUNNER_SSH_KEY` is added or the workflow is disabled. That is the point — a monitor that cannot run should be visibly broken rather than quietly green. **Merging this without adding the secret buys a red weekly job, not working monitoring.**

## Reviewed design tradeoffs

- **Absolute GB, not a percentage.** npm/pip/node need bytes, and all fleet Macs have ~926 GB disks. Observed healthy free space was 441 GB (mac-studio) and 232 GB (macbook-intel), so 50 GB is quiet in steady state while leaving lead time at a weekly cadence. A percentage would be simultaneously too lax on large disks and too strict on small ones.
- **A single `df /`, despite APFS.** On APFS `/` is the read-only system snapshot, but it shares container free space with `/System/Volumes/Data`. Verified identical on both reachable hosts, so one `df` is accurate and avoids branching on volume layout.
- **Disk check placed before the health-log branches.** Those `continue` on missing/stale logs, so checking disk after them would skip exactly the host that is most likely in trouble.
- **Fails closed on unreadable output.** Empty or non-numeric `df` output alerts rather than being treated as healthy.
- **Rejected: a second SSH multiplexed into the existing call.** One extra round trip per host per week is negligible, and leaving the working TIME_WAIT parsing untouched keeps the blast radius on a live alerting path small.
- **Rejected: keeping a soft "skip if no secret" mode.** That mode is the whole reason this went unnoticed for months.

## Verification

Dispatch-verified on this branch, since `schedule`/`workflow_dispatch`-only workflows are never exercised by PR CI:

- Run `30290651643` (before the hard-fail commit) — green, and the log confirms the poll was skipped.
- Run `30290862705` (after) — **fails** with the actionable error.
- Threshold logic unit-tested against real `df -Pk /` output plus the actual full-disk value (132 MiB → alert), empty, and non-numeric.


scarmani:
## Exact-head Tier-4 implementation/review preapproval request

**PR:** #9654  
**Exact head:** `7cf31b46bc6e5cc66e5d4ea1f9cd98ba64661c20`  
**Observed main:** `01fa6ca33161df20b53b1418bc0a3d7009180bee`

Live revalidation shows this is the sole open PR touching `.github/workflows/mac-runner-health-poll.yml`. It remains open, non-draft, `MERGEABLE/BLOCKED`; all five non-quorum required checks are green, `aragora-merge-quorum` is the only required failure, no reviews or dissent exist, and no lane, lease, process, steering, or duplicate file owner was found.

Current main still permits the scheduled monitor to succeed without polling when `MAC_RUNNER_SSH_KEY` is absent and does not check free disk. This exact one-file change makes missing credentials fail closed, requires an explicit `drift == '0'` before reporting health, and adds a 50 GB free-space floor. Because it changes a workflow and deliberately turns an inert green schedule red until the SSH secret is provisioned, the live packet correctly classifies it as Tier 4.

### Operator decision needed

Before advancing, confirm that a provider-neutral `MAC_RUNNER_SSH_KEY` can be provisioned for all listed hosts, or explicitly accept a red weekly monitor until it is. If the design is accepted, post:

```text
Tier-4 Implementation/Review Preapproval

PR: #9654
Exact head: 7cf31b46bc6e5cc66e5d4ea1f9cd98ba64661c20
Observed main: 01fa6ca33161df20b53b1418bc0a3d7009180bee

I authorize treating this exact one-file workflow implementation as the
preapproved Tier-4 design and authorize one exact-head Claude/OpenAI evidence
round. I confirm that MAC_RUNNER_SSH_KEY will be provisioned for all configured
hosts before the first expected healthy scheduled run, or I explicitly accept
the monitor remaining red until that prerequisite is met.

This does not authorize settlement, merge, manual CI reruns, branch-protection
mutation, a head change, or any other workflow change. Stop on head, ownership,
required-check, scope, or reviewer drift.
```

This comment is a decision request only. It is not preapproval, settlement, or merge authorization, and it does not change the branch, checks, labels, readiness, secrets, workflows, or protection state.

scarmani:

Head: 7cf31b4 (7cf31b46bc6e5cc66e5d4ea1f9cd98ba64661c20), committed 2026-07-29T04:48:17Z.
PR: #9654.
Model family: claude

Verdict: PASS

- [P3] `.github/workflows/mac-runner-health-poll.yml:91-177` — the header at `:22-24` promises "never green without having checked anything", but the step still emits `drift=0` if the `while read` loop body executes **zero** times (loop under `set +e`, process-substitution failure invisible), producing "All Mac runners healthy". Not currently reachable — I read `:85-90` and HOSTS_JSON is a valid static literal, and `jq` is preinstalled on `ubuntu-latest` — but note the step explicitly `apt-get install`s `openssh-client` while silently assuming `jq`. Fix: count iterations and fail if `polled != $(jq 'length')`. This is the same vacuous-green class as the `ssh -n` bug in #9671.
- [P3] `:105` and `:122` — no `timeout-minutes` on the job (default 360 min), and this PR adds a **second** unbounded SSH session per host. `ConnectTimeout=10` bounds only the handshake, not the session; a 100%-full Mac is exactly the host likely to hang *after* connect (login writes to the Data volume). So the disk monitor can stall for hours on the condition it monitors. Verified the gap is real: `timeout-minutes` appears 215× across 77 other workflows in `.github/workflows`, but zero times here. Fails red rather than silently green, hence non-blocking. Fix: `timeout-minutes: 10` on the job plus `-o ServerAliveInterval=10 -o ServerAliveCountMax=3`.
- [P3] Docs drift introduced by this PR. I read both files: `scripts/runners/README.md:43` still states the criteria as "alerts if `timewait > 5000` or log is > 48h stale" — the third criterion and `DISK_MIN_FREE_GB` are undocumented. And `grep -rn "disk\|ENOSPC\|free space" docs/runners/FLEET.md` returns **nothing**, so a 3-week fleet-wide ENOSPC incident is recorded only in a YAML comment, while the alert body at `:201` points operators to FLEET.md "Past incidents" for the #6474 playbook.
- [P3] `:80` — `KEY_READY=1` is now written but never read. `grep -rn KEY_READY` across the whole repo returns exactly one hit (that write); both consumer conditions were removed at `:82` and `:223`. Harmless, but it reads like a live gate to the next editor. Delete it or keep it consumed.
- [P3] `:122-131` — `2>/dev/null` plus `|| echo ""` collapses "SSH died", "`df` not found", and "empty output" into one opaque `could not read free disk space` alert with no captured stderr. Given the first SSH already succeeded, this alert means something genuinely odd happened and the operator gets nothing to act on.
- [P3] Inherited, not introduced, but adjacent: `:170-172` writes `alerts<<EOF` to `$GITHUB_OUTPUT` where one alert embeds remote-controlled `$LAST_LINE` (`:150`). A host log line of `EOF` would close the heredoc early and let following lines forge outputs such as `drift=0`. This PR does **not** widen it — the new disk strings interpolate only `$name` and a value the `case` at `:128-131` has already proven numeric. Worth a random delimiter when next touched.
- [P3] Landing precondition, disclosed in the PR and not a code defect: the fail-closed `exit 1` at `:76` will turn every Monday run red until `MAC_RUNNER_SSH_KEY` is provisioned. I checked the blast radius rather than assuming it: the workflow is `schedule` + `workflow_dispatch` only (no PR trigger, so not a required check), `docs/AGENT_OPERATING_CONTRACT.md:69` scopes main-red incident mode to *required* checks, and `main-required-checks-auto-revert.yml:5-10` is scoped to Lint / SDK Parity Check / OpenAPI Spec / SDK Tests. No auto-revert or merge-gate cascade. Still needs operator provisioning or a written risk disposition before landing.

Correctness of the change itself holds up: disk check correctly sequenced after the `ssh-failed` `continue` and before the `no-log` `continue`; `-n` present on the new SSH and pinned by `test_mac_poll_ssh_is_pinned`; KiB→GiB divisor and `-lt` boundary verified exact; `drift != '1'` → `drift == '0'` is a strict fail-closed improvement.

Limitations: actionlint was denied in this session, so I did not verify it personally. All evidence above is local and mocked — **fleet reachability, real SSH behavior, and current Mac disk state are UNMEASURED**; no host was contacted, no workflow run, no GitHub mutation, no edits.

dogfood: yes

scarmani:

Head: 7cf31b4 (7cf31b46bc6e5cc66e5d4ea1f9cd98ba64661c20), committed 2026-07-29T04:48:17Z.
PR: #9654.
Model family: openai

Verdict: PASS

No findings.

Head reviewed: `7cf31b46bc6e5cc66e5d4ea1f9cd98ba64661c20`.

Validation:
- Read full changed workflow and compared against `origin/main`.
- `actionlint -shellcheck= .github/workflows/mac-runner-health-poll.yml`: passed.
- Requested pytest command: blocked by read-only sandbox before collection, `No usable temporary directory`.
- Requested `validate_workflow.py`: blocked by read-only sandbox creating its temp case directory, `Operation not permitted`.

dogfood: yes

scarmani:

PR: #9654
Exact head: 7cf31b46bc6e5cc66e5d4ea1f9cd98ba64661c20
Authorized action: one normal protected squash merge with --match-head-commit, only if every live exact-head gate permits.

Human-risk settlement: I accept the Tier 4 risk for the existing fail-closed monitor design, the seven unchanged Cycle368 P3 caveats, and the optional weekly monitor remaining red until its SSH configuration works. Missing or failed polling must not be represented as fleet health. Live fleet, network access, and secret provisioning remain unmeasured and outside this authorization.

The original prepared Claude/OpenAI reviews remain unchanged. Claude ran five SSH tests and nine mocked cases, while actionlint was denied. OpenAI ran actionlint, while its sandbox prevented behavioral tests from creating temporary files. These limitations and the existing attempt budget are preserved.

Recorded by Codex on the operator's direct 2026-09-15 approval to execute this pending decision after #10085 and #9688. This records human risk acceptance, not an independent human technical review.

No source changes, new reviews, administrative bypass, forced push, branch deletion, secret provisioning, network changes, or protection mutation. No manual CI rerun is inferred from the decision's conditional rerun option. Stop on any other head, custody, required-check, dissent, or merge-state blocker.

PR #9663, opened 2026-07-28
docs(runners): correct m1 runner IP, roster counts, and record the 2026-07-27 incident

scarmani:
## Summary

`FLEET.md` recorded `macbook-m1-16gb` at `10.0.0.170`. It isn't there — DHCP moved it to
`10.0.0.108`. Because the recorded IP was used as a *liveness test*, the host was written off as
powered down while it was up the entire time. Corrects the roster against the live API and records
the incident.

## Roster corrections (verified 2026-07-27)

| Row | Change |
|---|---|
| `macbook-m1-16gb` | IP `.170` → `.108` (DHCP); marked **online**, re-registered today; data volume ~95% full |
| `macbook-intel-64gb` | marked **not registered** — GC'd server-side, host off-network |
| 6 × AWS EC2 | marked **offline** (account suspension) |
| `**Total online: 12**` | stale and unverifiable → Registered/Online counts, the gap explained against the monitor's baseline, and the one-liner that regenerates them |

Added a note that the IP column is neither an identity nor a liveness check — identify by
`scutil --get LocalHostName` over SSH, judge liveness by the runner API.

## The incident entry

Both Mac incidents present an **identical surface symptom**: runner absent from the API, local
install present, LaunchAgent loaded. The entry leads with the discriminator, because that is what
would actually have saved time:

| | 2026-04-23 | 2026-07-27 |
|---|---|---|
| Log line | `Can't assign requested address` | `√ Connected to GitHub` |
| Cause | TCP ephemeral-port exhaustion (93d uptime) | GitHub deleted the registration for inactivity |
| Fix | reboot | re-register (reboot cannot fix server-side state) |

**One line — `√ Connected to GitHub` — rules out the entire network-exhaustion family.**

Second half of the failure: this host ran a non-standard
`~/Library/LaunchAgents/com.github.actions-runner.plist` with `KeepAlive=true`. The runner exits
deliberately on a deleted registration (`no retry needed`), and launchd restarted it forever —
`Runner.Listener` at **100% CPU, load average 84** on a 16 GB M1, doing zero work. That non-canonical
`Label` also made `./svc.sh status` report `not installed` on a host that was very much running one.

## Reviewed design tradeoffs

- **Recorded the misdiagnoses, not just the fix.** Both wrong turns here (IP-as-liveness, and
  assuming a recurrence of the April incident) are ones the *existing* doc actively encouraged — it
  published an IP table and a single Mac-runner incident whose fix was "reboot". Documenting only
  the resolution would leave both traps armed.
- **Left the monitor baseline at 12 rather than lowering it to 11.** The gap is real and should keep
  alerting; silencing it by editing the baseline would hide the suspended EC2 fleet and the missing
  MacBook. The doc now explains the gap instead.
- **Did not add the `KeepAlive` throttle as a code change here.** The right fix is preferring
  `svc.sh install` over hand-rolled plists, which the doc already prescribes; a `ThrottleInterval`
  patch to a plist that should not exist would entrench it. Called out in Lessons instead.
- **Marked rows in the existing table rather than splitting into online/offline sections.** Keeps
  one canonical roster to diff against `gh api` output; two lists drift.

## Verification

Roster figures taken directly from:

```bash
gh api repos/synaptent/aragora/actions/runners --jq '.runners[]|"\(.status)\t\(.name)"' | sort
# 11 registered, 5 online
```

Docs-only. `node docs-site/scripts/sync-docs.js` run — 276 files synced, no mirror changes for this
file, so no `build` red.

### DIFF
--- a/docs/runners/FLEET.md
+++ b/docs/runners/FLEET.md
@@ -5,28 +5,93 @@ Canonical roster of self-hosted GitHub Actions runners registered on
 `gh api repos/synaptent/aragora/actions/runners` endpoint cross-checked
 against on-host `~/actions-runner/.runner` configs.
 
-Last confirmed: **2026-04-23**.
+Last confirmed: **2026-07-27**.
 
-## Registered and online
+> **LAN IPs are DHCP and drift.** Do not treat the IP column as an identity or a
+> liveness test. On 2026-07-27 `macbook-m1-16gb` was assumed dead because it stopped
+> answering at its recorded address; it had simply moved (`.170` → `.108`) and was up
+> the whole time. Identify a host by `scutil --get LocalHostName` over SSH, and judge
+> liveness by the runner API, never by pinging a value from this table.
+
+## Roster
 
 | Runner name (GH) | Host | IP | OS | Labels | Notes |
 |---|---|---|---|---|---|
 | `aragora-hetzner-cpu1` | Hetzner CX52 cloud | — | Linux x64 | `self-hosted, Linux, X64, aragora, hetzner` | Background CPU work |
 | `aragora-hetzner-cpu2` | Hetzner CX52 cloud | — | Linux x64 | `self-hosted, Linux, X64, aragora, hetzner` | Background CPU work |
 | `aragora-hetzner-cpu3` | Hetzner CX52 cloud | — | Linux x64 | `self-hosted, Linux, X64, aragora, hetzner` | Background CPU work |
-| `i-07e538fafbe61696d` | AWS EC2 (production) | — | Linux x64 | `self-hosted, Linux, X64, aragora` | Production canary — also hosts the Aragora service |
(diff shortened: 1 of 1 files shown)


scarmani:

Head: 1fb1c9e (1fb1c9e7677c546e24c509bc7b26e743779e1bb9), committed 2026-07-28T03:07:02Z.
PR: #9663.
Model family: openai

Verdict: PASS

- [P3] `docs/runners/FLEET.md`: The new text exposes current internal runner details: live LAN IP, AWS account suspension, storage pressure, service labels, and recovery commands. I could verify this disclosure directly in the diff, but not whether the repo is public or access-controlled; repo visibility would determine whether this should be blocking.
- [P3] `docs/runners/FLEET.md`: “Baseline ... is 12; the 7-runner gap...” is ambiguous because the table says `Registered: 11 · Online: 5`. The math only works if the baseline is for online runners, not registered runners; I could not verify `runner-headcount-monitor.yml` from the shown files.

dogfood: yes

Issue #9661, opened 2026-07-28
DecisionReceipt fidelity: mid-debate responses recorded twice, and input_hash not reproducible


scarmani:
## Summary

Two defects in what a `DecisionReceipt` records about the debate it describes.
Both found reviewing the crux-cards artifact added to the July EU AI Act bundle
(#9657), and both matter more than usual because that receipt is published
evidence an external auditor is invited to check.

## 1. Every mid-debate agent response is recorded twice

`agent_responses` contains each round-1 and round-2 response **exactly twice**.
Confirmed across four independent debates run today:

| receipt | responses | duplicated | rounds |
|---|---|---|---|
| `06cfcbef` | 5 | 2 | 0,1 |
| `ea8b86d2` | 7 | 3 | 0,1,2 |
| `6741923c` | 5 | 2 | 0,1,2 |
| `5925f542` | 9 | 4 | 0,1,2 |

Only the round-0 proposal appears once. This is systemic, not run-specific.

**Downstream effects:**

- `agent_contributions` overcounts: in `5925f542` codex reports `messages: 4`
  for two actual messages. Critique counts appear unaffected, which suggests the
  duplication is specific to the message/response recording path.
- Duplicated messages enter the crux belief network as byte-identical claims
  (`msg_1_codex` and `msg_2_codex` in that receipt), so they occupy two of the
  top-k crux slots with the same statement.

Crux *scoring* is not inflated — `compute_disagreement_scores` takes the max per
author, so duplicate edges from one agent do not accumulate — but the published
record misrepresents how many times each agent spoke.

## 2. `input_hash` is not reproducible from the recorded input

`input_hash` on `5925f542` is
`4629a019e105f1d09a7c460601fa1c5ceef32bd8a70d4d846119c7e448642a10`, which is
sha256 of neither `task` nor `input_summary` (identical strings here), under
plain, stripped, trailing-newline or NFC encodings.

`receipt_models.py:1577` derives `sha256(task)` only when no hash is supplied,
so this one came from an upstream path hashing something else — likely a fuller
input payload that is not itself recorded.

No current bundle claim depends on it: the published verification chain names
the artifact-hash check, which passes. But the field is documented as "SHA-256
for integrity verification" and an auditor will try to reproduce it and fail.

**Fix direction:** either record the pre-image the hash is taken over, or derive
it from a field the receipt actually contains.

## Acceptance

- A debate of N rounds records each agent response once, and
  `agent_contributions` message counts match the transcript.
[...]

PR #9665, opened 2026-07-28
fix(debate): record each message once; the mock was hiding the duplication (#9661)

scarmani:
## Summary

Fixes the first and most consequential defect in #9661: **every mid-debate
response was recorded twice** in DecisionReceipts.

## Root cause

`DebateContext.add_message` already appends to `context_messages`,
`partial_messages` **and** `result.messages` (`debate_state.py:377-382`). Five
call sites then appended to `result.messages` again — `revision_phase`,
`debate_rounds` (×2), `critique_generator`, `debate_rounds_helpers`.

Effects on published receipts:

- `agent_responses` held each round-1/round-2 message twice
- `agent_contributions` overcounted — codex showed `messages: 4` for two actual
  messages
- duplicated messages entered the crux belief network as **byte-identical
  claims** that occupied top-k slots, distorting `crux_score` ranking

Dissent attribution was unaffected: `disagreement_score` takes the max per
author, so duplicate edges from one agent do not accumulate. That distinction is
why the July bundle's crux gate is currently recorded as *"met for dissent
attribution; ranking provisional"* rather than closed — this PR is what lets it
close.

## Why the tests did not catch it

`MockDebateContext.add_message` appended only to `context_messages`, unlike the
real method. The mock **cancelled the production bug out**: phases
double-appended, the mock single-appended, and
`test_execute_appends_to_result_messages` asserted `== 1` and passed.

A test that actively protected the defect. The mock now mirrors the real
contract on all three destinations, and a new test asserts each message is
recorded **exactly once** in all three — "exactly once" rather than "at least
one" is the entire point. It fails against the previous implementation.

`self._partial_messages` in `debate_rounds` is a genuinely separate list exposed
via `get_partial_messages()`, so those appends are deliberately untouched.

## Unplanned scope — flagged, not buried

The pre-push mypy hook gates on changed files, so touching these four pulled in
**8 pre-existing type errors** that had to be cleared to land a 4-line fix.

Those changes are mechanical narrowing (binding optional callables to locals so
mypy can narrow them; one implicit-Optional default), **with one caveat worth a
reviewer's attention**: two sites now `return None` early where a `None`
callable would previously have raised `TypeError`. That path should not occur —
the callbacks are always wired by the Arena, and there is already a
`if not self._critique_with_agent: skip` guard upstream — but it is a real
behaviour change on a branch that is unreachable in practice.

## Testing

**1624 debate tests pass.** mypy clean on all four touched files (was 8 errors
on `main`).

## Still open in #9661

Untouched here and deliberately not bundled: non-reproducible `input_hash` (and
the `evidence_hash` fields that reuse its pre-image), `rounds_used: 2` against
responses spanning rounds 0–2, and the receipt recording no generating commit.

### DIFF
--- a/aragora/debate/debate_state.py
+++ b/aragora/debate/debate_state.py
@@ -375,10 +375,10 @@ class DebateContext:
         return self.proposals.get(agent_name, "")
 
     def add_message(self, msg: Message) -> None:
-        """Add a message to both context and partial tracking."""
+        """Record one message in context, recovery, and final-result tracking."""
         self.context_messages.append(msg)
         self.partial_messages.append(msg)
-        if self.result:
+        if self.result is not None:
             self.result.messages.append(msg)
 
     def record_agent_failure(
--- a/aragora/debate/phases/critique_generator.py
+++ b/aragora/debate/phases/critique_generator.py
@@ -166,15 +166,21 @@ class CritiqueGenerator:
         async def generate_critique(critic: Agent, proposal_agent: str, proposal: str):
             """Generate critique and return CritiqueResult."""
             logger.debug("critique_generating critic=%s target=%s", critic.name, proposal_agent)
+            # Bound once so the optional-callable narrowing holds at every call
+            # site below (the attribute form defeats it).
+            critique_with_agent = self._critique_with_agent
+            if critique_with_agent is None:
+                return None
             base_timeout = getattr(critic, "timeout", AGENT_TIMEOUT_SECONDS)
             timeout = get_complexity_governor().get_scaled_timeout(float(base_timeout))
             task_id = f"{critic.name}:critique:{proposal_agent}"
 
             try:
                 with streaming_task_context(task_id):
-                    if self._with_timeout:
-                        crit_result = await self._with_timeout(
-                            self._critique_with_agent(
+                    with_timeout = self._with_timeout
+                    if with_timeout is not None:
+                        crit_result = await with_timeout(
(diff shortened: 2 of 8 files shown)


scarmani:

Head: 5d4575c (5d4575c5867443e924a1bc6ab572d900f65a876b), committed 2026-07-28T06:28:39Z.
PR: #9665.
Model family: claude

Verdict: PASS

- [P3] The inline `_critique_phase`/`_revision_phase` in `debate_rounds.py` received the dedup fix (lines 1241/1503) but *not* the callback-binding hardening — they still read `self._critique_with_agent` (line 988/1018) and `self._generate_with_agent` live, unlike the extracted `CritiqueGenerator`/`RevisionGenerator`. Pre-existing and theoretical, not a regression; noting for consistency.
- [P3] Unverified: I did not run the full `tests/debate/` suite. A static scan found no existing test outside the diff asserting a *doubled* `result.messages` count (the count assertions I found are on `context_messages`/`partial_messages`, which are unchanged, or on checkpoint-constructed results). Risk is low; `pytest tests/debate/` would confirm nothing regressed on the changed message-count behavior.

dogfood: yes

scarmani:

Head: 5d4575c (5d4575c5867443e924a1bc6ab572d900f65a876b), committed 2026-07-28T06:28:39Z.
PR: #9665.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9668, opened 2026-07-28
fix(deps): bump the whole tailwind family to 4.3.3, not just the top-level package

scarmani:
## Summary

**Supersedes #9624**, which bumps only top-level `tailwindcss` to 4.3.3 and leaves the engine on
4.3.2 — a state that is arguably worse than not bumping at all, because it looks applied and isn't.

`@tailwindcss/postcss`, `@tailwindcss/node` and `@tailwindcss/oxide` exact-pin `tailwindcss` to
their own version. Moving the top-level package alone makes npm materialize **nested 4.3.2 copies**:

```
node_modules/@tailwindcss/postcss/node_modules/tailwindcss   4.3.2
node_modules/@tailwindcss/node/node_modules/tailwindcss      4.3.2
node_modules/tailwindcss                                     4.3.3   ← hoisted
```

Source `@import "tailwindcss"` resolves the hoisted **4.3.3**, but the PostCSS pipeline compiles
with the **4.3.2** engine. Upstream exact-pins siblings precisely to prevent that pairing, and it
can silently change generated CSS.

## Provenance

Found by the merge-quorum **claude** reviewer on #9624 during a queue drain. I re-checked after
Dependabot rebased that PR to `62cf538` — **the split is still present**, so the rebase did not
resolve it.

## The fix

`tailwindcss` 4.3.2 → 4.3.3 (exact-pinned, so it needs an explicit edit) and
`@tailwindcss/postcss` `^4.3.2` → `^4.3.3` so node/oxide follow. Result:

```
4.3.3  node_modules/tailwindcss
4.3.3  node_modules/@tailwindcss/postcss
4.3.3  node_modules/@tailwindcss/node
4.3.3  node_modules/@tailwindcss/oxide
nested tailwindcss copies: NONE
```

## Reviewed design tradeoffs

- **Rejected: merging #9624 and fixing forward.** The intermediate state ships 4.3.3 CSS compiled by
  a 4.3.2 engine. If it changes generated CSS, the regression lands looking like a routine patch
  bump — the expensive failure mode. Supersede is cleaner than repair-after-merge.
- **Rejected: an `overrides` entry forcing `tailwindcss` 4.3.3 into the siblings.** It would flatten
  the tree without moving the engine packages, recreating a cross-version pairing upstream
  deliberately prevents — the same class of mistake as #9589's `brace-expansion` override.
- **Chose `^4.3.3` over exact-pinning `@tailwindcss/postcss`.** Keeps Dependabot able to move the
  family normally; only `tailwindcss` itself stays exact-pinned, matching existing convention.
- **Not adding a lockfile-uniformity CI check.** Tempting after two split-version bugs in one week
  (this and `brace-expansion`), but the general "sibling packages must agree" rule needs a
  per-family allowlist to avoid false positives. Worth its own issue rather than smuggling a new
  gate into a dependency PR.

## Verification

```
npm ci && npm audit --omit=dev --audit-level=high   -> 0 vulnerabilities, exit 0
npm run lint       -> clean (--max-warnings 0)
npm test           -> 260 suites / 4035 tests pass
npm run build:ci   -> success; CSS emitted (316KB)
```

Engine actually moved, checked from the built tree rather than inferred from the lockfile:

```
@tailwindcss/postcss 4.3.3 / @tailwindcss/node 4.3.3
```

**Flake disclosed:** one run out of four reported a single test failure that did not reproduce
across three subsequent full runs; the name was not captured before it went green. Not investigated
and not attributed to this change.

Closes #9624 once merged.

### DIFF
--- a/aragora/live/package-lock.json
+++ b/aragora/live/package-lock.json
@@ -40,7 +40,7 @@
         "@eslint/js": "^9.39.2",
         "@next/bundle-analyzer": "^16.2.9",
         "@playwright/test": "^1.61.1",
-        "@tailwindcss/postcss": "^4.3.2",
+        "@tailwindcss/postcss": "^4.3.3",
         "@testing-library/dom": "^10.4.1",
         "@testing-library/jest-dom": "^7.0.0",
         "@testing-library/react": "^16.3.2",
@@ -59,7 +59,7 @@
         "jest-environment-jsdom": "^30.4.1",
         "openapi-typescript": "^7.13.0",
         "postcss": "^8.5.22",
-        "tailwindcss": "4.3.2",
+        "tailwindcss": "4.3.3",
         "ts-jest": "^29.4.11",
         "typescript": "^5",
         "typescript-eslint": "^8.54.0"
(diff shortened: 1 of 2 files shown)


scarmani:

Head: ad4863c (ad4863ce52b51fe32cc27e4585b2aad9c9f9b740), committed 2026-07-28T18:03:59Z.
PR: #9668.
Model family: claude

Verdict: PASS

Verified locally (worktree at ad4863c): the entire tailwind family (`tailwindcss`, `@tailwindcss/postcss`, `@tailwindcss/node`, `@tailwindcss/oxide` + all 12 platform packages) is uniformly at 4.3.3 with no stale 4.3.2 pins (the one remaining "4.3.2" string is the unrelated pre-existing `ansi-escapes@4.3.2`); the root `packages[""]` block matches package.json exactly; `enhanced-resolve@5.24.4` satisfies `@tailwindcss/node`'s new `^5.24.1` spec, is that spec's only dependent in the lockfile, and its own deps are satisfied by the existing `tapable@2.3.3` and `graceful-fs@4.2.11` entries; every changed `resolved` URL points at registry.npmjs.org; and no other package.json in the repo consumes tailwindcss.

- [P3] Unverified: I could not confirm the 4.3.3 / enhanced-resolve 5.24.4 tarballs' integrity hashes or that these versions exist upstream — both `npm view` and WebFetch were permission-denied in this session, and 4.3.3 post-dates my knowledge. `npm ci` (exercised by the PR-time frontend build gate) or `npm view <pkg>@<ver> dist.integrity` would settle it.
- [P3] `aragora/live/package-lock.json:3044-3073` — the nested `inBundle` sub-entries under `@tailwindcss/oxide-wasm32-wasi` were dropped, the signature of regenerating the lock on a platform where this optional wasm32 dep isn't installed. Benign for `npm ci` (bundled deps ship inside the tarball; the entry keeps `bundleDependencies` and the package is optional/wasm32-only, and independent top-level `@emnapi/*` entries still exist for their other dependents), but expect these lines to churn back if the lock is next regenerated where the package does install.

dogfood: yes

scarmani:

Head: ad4863c (ad4863ce52b51fe32cc27e4585b2aad9c9f9b740), committed 2026-07-28T18:03:59Z.
PR: #9668.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9671, opened 2026-07-29
fix(ci): ssh -n in the Mac health poll so it does not stop after the first host

scarmani:
## Summary

`mac-runner-health-poll.yml` iterates hosts with
`while read -r entry; do ... done < <(echo "$HOSTS_JSON" | jq -c '.[]')` and calls `ssh` inside the
loop. ssh forwards its **inherited stdin** to the remote — and that stdin is the loop's own host
stream. The first ssh drains it, the loop exits after host 1, `alerts` stays empty, `drift=0` is
written, and the workflow reports **every Mac healthy having contacted exactly one**. ShellCheck
SC2095.

Mechanism reproduced locally, substituting a stdin-consuming command for ssh:

```
ssh WITHOUT -n (current main) -> hosts polled: 1 of 3
ssh WITH    -n (the fix)     -> hosts polled: 3 of 3
```

## This one already bit us

On 2026-07-27/28 `macbook-m1-16gb` sat in a launchd crash-loop — `Runner.Listener` at 100% CPU,
load average 84 — after GitHub garbage-collected its registration, and `macbook-intel-64gb` was
deregistered entirely. Recovering them took hours.

`mac-studio-m3ultra` is **first** in `HOSTS_JSON` and was healthy. So the monitor built to catch
exactly that outage would have polled one host, found it fine, and printed "All Mac runners
healthy" for two machines it never contacted.

## Provenance and scope

Found by the merge-quorum **claude** reviewer on #9654, which adds a *second* ssh call with the same
defect. This PR fixes the **pre-existing** call on main, separately and deliberately: the
pre-existing ssh runs first and drains stdin before #9654's new call is ever reached, so #9654
cannot deliver its stated guarantee until this lands. **#9654 still needs `-n` on its own call** —
that remains its author's change to make.

## The guard needed a second pass — worth reading

`tests/ci/test_workflow_ssh_stdin.py` fails any workflow `run:` block containing a `while read` loop
plus an ssh invocation lacking `-n` (or an explicit `< /dev/null`).

The first version tested `"-n" in invocation.split()`. That matched the `-n` belonging to the
**remote command** — `tail -n 1 ~/Library/Logs/...` — so it **passed against the un-fixed
workflow**. A guard that is vacuous against the single bug it was written for is worse than no
guard, because it launders the gap as coverage. I only caught it by removing `-n` and checking the
guard actually went red.

It now parses ssh's own flags: scan tokens after `ssh`, skip option arguments (`-o X`, `-i X`),
stop at the `[user@]host` destination. So neither a remote-command flag nor an option value
(`-o SendEnv=-n`) can mask a missing ssh `-n`, and `2>/dev/null` correctly does not count as stdin
detachment. Those three cases are pinned as explicit regression tests.

## Reviewed design tradeoffs

- **Rejected: fixing this inside #9654.** It's another author's branch, and the pre-existing call is
  independently broken on main today. Landing the base fix separately means #9654 reduces to a
  one-flag change and this protection doesn't wait on that PR's review cycle.
- **Rejected: `< /dev/null` instead of `-n`.** Equivalent, but `-n` is the documented ssh idiom and
  reads as intent rather than incidental redirection. The guard accepts both.
- **Scoped the guard to `.github/workflows/*.yml`.** The same hazard exists in `scripts/*.sh`, but
  shell scripts have far more ssh-shaped false positives (heredocs, comments, `ssh-keygen`). Better
  a precise guard where the bug actually lives than a noisy one that gets muted.
- **Kept the `while read` precondition.** An ssh outside a read loop consuming stdin is usually
  harmless; flagging every ssh repo-wide would produce noise and train people to ignore it.

## Verification

```
5/5 tests pass with the fix in place
removing `-n` from the workflow → 2 tests fail, naming the exact offending invocation
restoring it → passes again
workflow YAML parses
```

Refs #9654.

### DIFF
--- a/.github/workflows/mac-runner-health-poll.yml
+++ b/.github/workflows/mac-runner-health-poll.yml
@@ -76,7 +76,11 @@ jobs:
 
             echo "=== Polling $name ($host) ==="
 
-            LAST_LINE=$(ssh -o StrictHostKeyChecking=no -o ConnectTimeout=10 \
+            # -n is load-bearing: without it ssh inherits this loop's stdin (the
+            # jq host stream) and forwards it to the remote, draining the remaining
+            # hosts. The loop would then exit after the FIRST host with alerts empty
+            # and report every Mac healthy. See tests/ci/test_workflow_ssh_stdin.py.
+            LAST_LINE=$(ssh -n -o StrictHostKeyChecking=no -o ConnectTimeout=10 \
                            -i ~/.ssh/mac_runner \
                            armand@"$host" \
                            'tail -n 1 ~/Library/Logs/aragora-runner-health.log 2>/dev/null || echo no-log' \
--- a/tests/ci/test_workflow_ssh_stdin.py
+++ b/tests/ci/test_workflow_ssh_stdin.py
@@ -1,0 +1,177 @@
+"""Guard against ssh draining a `while read` loop's stdin in workflow run blocks.
+
+`ssh` forwards its inherited stdin to the remote command. Inside a
+``while read ...; do ... done < <(producer)`` loop, that inherited stdin *is* the
+producer stream, so the first ssh consumes every remaining item and the loop exits
+after one iteration — silently. This is the ShellCheck SC2095 class.
+
+The failure mode is specifically nasty for health checks: the loop finishes with an
+empty alert list and reports everything healthy, having contacted exactly one host.
+That is what `mac-runner-health-poll.yml` did — a monitor that would have reported
+"All Mac runners healthy" while `macbook-m1-16gb` was crash-looping and
+`macbook-intel-64gb` was deregistered entirely.
+
+The fix is `ssh -n` (or an explicit `< /dev/null`) on every ssh call inside such a loop.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+import pytest
+import yaml
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+WORKFLOW_DIR = REPO_ROOT / ".github" / "workflows"
+
+_LOOP = re.compile(r"\bwhile\b[^\n]*\bread\b")
+_SSH_START = re.compile(r"(?:^|[|&;(]|\$\()\s*ssh\s")
+
+
+def _run_blocks(path: Path) -> list[str]:
+    """Every `run:` script in a workflow, including inside composite step lists."""
+    try:
+        doc = yaml.safe_load(path.read_text(encoding="utf-8"))
+    except yaml.YAMLError:  # pragma: no cover - surfaced by the yaml lint hook
+        return []
+    blocks: list[str] = []
+
+    def walk(node: object) -> None:
+        if isinstance(node, dict):
+            run = node.get("run")
+            if isinstance(run, str):
+                blocks.append(run)
+            for value in node.values():
+                walk(value)
+        elif isinstance(node, list):
+            for item in node:
+                walk(item)
+
+    walk(doc)
+    return blocks
+
+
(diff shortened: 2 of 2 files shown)


scarmani:

Head: a10f70b (a10f70bbc68333501c3e0ecde4f11afce8c834ab), committed 2026-07-29T01:18:06Z.
PR: #9671.
Model family: claude

Verdict: PASS

Verified: ran `python -m pytest tests/ci/test_workflow_ssh_stdin.py -v` at head in this worktree — 5 passed. Read the full `mac-runner-health-poll.yml`: the loop feeds from `done < <(echo "$HOSTS_JSON" | jq -c '.[]')` (line 121) and the single ssh call now carries `-n` — the diagnosis (SC2095 stdin drain) and fix are correct, and key-based auth means `-n` cannot break authentication. Listed `.github/workflows/`: every workflow is `.yml`.

- [P3] `tests/ci/test_workflow_ssh_stdin.py:27` — `WORKFLOW_DIR.glob("*.yml")` skips `.yaml` workflows and the `templates/` subdirectory. Verified no `.yaml` file exists today (directory listing), so it's currently vacuous, but a future `.yaml` workflow silently escapes both guard tests. Glob both extensions.
- [P3] `tests/ci/test_workflow_ssh_stdin.py:100` — attached-value options are char-mined for flags: `-oSendEnv=-n` (or any attached `-oXxx=no…` value containing `n`) is not in `_OPTS_WITH_ARG` (exact-match only), so `flags.update` per-char yields a spurious `-n` → false-safe. The regression test only covers the spaced form `-o SendEnv=-n`. Verified by tracing `_ssh_own_flags`; no current workflow uses the attached form. Relatedly `-B` (bind_interface, takes an arg) is missing from `_OPTS_WITH_ARG`, though that failure mode is a loud false positive rather than vacuity.
- [P3] `tests/ci/test_workflow_ssh_stdin.py:29` — `_SSH_START` only matches ssh after line-start or `|&;($( `, so `if ssh …`, `timeout 30 ssh …`, `env FOO=1 ssh …` inside a read loop escape the guard entirely. Verified by regex inspection; no such pattern exists in current workflows (guard test passed repo-wide).
- [P3] `tests/ci/test_workflow_ssh_stdin.py:110` — `_stdin_is_safe` scans the whole joined invocation including the quoted remote command, so `ssh host 'cmd < /dev/null'` (remote-side redirect) counts as locally safe → false-safe potential.
- [P3] `tests/ci/test_workflow_ssh_stdin.py:152` — `test_guard_detects_a_synthetic_offender` takes `tmp_path` but never uses it, and despite the docstring it never exercises `_run_blocks`/`_offenders` on an actual synthetic workflow file — only the string helpers. Writing the sample to `tmp_path` and running the YAML path end-to-end would make the self-test honest.

All findings are completeness gaps in a guard heuristic, none reachable by current repo content (verified by the passing repo-wide scan); the workflow fix itself is correct.

dogfood: yes

scarmani:

Head: a10f70b (a10f70bbc68333501c3e0ecde4f11afce8c834ab), committed 2026-07-29T01:18:06Z.
PR: #9671.
Model family: openai

Verdict: PASS

- [P3] `tests/ci/test_workflow_ssh_stdin.py:91` The ssh flag parser only skips option arguments when the option is a separate token, so accepted forms like `ssh -oSendEnv=-n ...` are misclassified as having ssh-level `-n`. I verified this by importing the helper locally: `_stdin_is_safe("ssh -oSendEnv=-n user@host 'cmd'")` returns `True`. This weakens the future workflow guard but does not affect the fixed workflow invocation in this PR.

dogfood: yes

Issue #9391, opened 2026-07-17
P0 migration ledger: retire AWS production and restore provider-neutral API


scarmani:
## Operator decision — 2026-08-19: AWS is retired infrastructure

AWS recovery is no longer the remediation path for this incident. Do not spend additional cycles on MFA, account recovery, billing reinstatement, EC2, SSM, or origin diagnosis. The production replacement must remain provider-neutral at the application boundary.

### Migration ledger and exit gate

- [ ] Provider-neutral secret loading is implemented and validated without raw production-secret fallback.
- [ ] Non-AWS ODR receipt signing is fail-closed and independently verifiable.
- [ ] A canary API runs through the existing container path with standard PostgreSQL persistence suitable for Supabase.
- [ ] Cloudflare routes a canary hostname to an authorized non-AWS target.
- [ ] External `/healthz`, `/api/health`, WebSocket, persistence, signing-key discovery, signed-receipt, and offline verification checks pass.
- [ ] The same persistence and signing proof passes again after one canary restart.
- [ ] Exact deployed SHA/image digest, key fingerprint, verification times, and rollback proof are recorded here without secret material.
- [ ] Only after the above: a separately authorized Tier-4 PR disables AWS deploy/monitor noise and corrects AWS-only governance and operational docs.

### Active implementation lane

- Draft PR: https://github.com/synaptent/aragora/pull/9800
- Staged exact head: `a9b5de83502426ff83415e64b9b27dc562161e1d`
- Plan: `docs/plans/aws-retirement-migration-9391.md`
- Hard boundary: no workflow, protected-governance, settlement, merge, AWS recovery, or paid-host creation in the migration lane.
- Provider choices such as Hetzner, Cloudflare, and Supabase belong in deployment configuration and secret adapters, not application branches.

### Current access constraints

- Supabase CLI authentication is available.
- Local Docker daemon, Hetzner authentication, and Cloudflare tunnel authentication are not currently available.
- These constraints do not block provider-neutral implementation and dry-run proof, but they block the live canary until an existing authorized target and credentials are available.

---

## Incident

`https://api.aragora.ai/api/health` is currently unreachable. GitHub Actions consistently receives Cloudflare 522, and an independent local probe times out with zero response bytes. The public frontend remains reachable.

## Timeline

- Last successful Uptime Monitor: 2026-07-16 14:40:54 UTC — https://github.com/synaptent/aragora/actions/runs/29507686041
[...]

PR #9680, opened 2026-07-29
docs(plans): record the founder's W4 Variant B decision

scarmani:
## Summary

Records the founder's decision of 2026-07-28: **the W4 demo ships Variant B.**

W4's external outcome specified "signed dissent-preserving receipts", written
assuming production signing. The production Ed25519 key is in AWS Secrets
Manager in the suspended account (#9391), and the operator has stated
reinstatement cannot be resolved on the required timeline — so a
production-signed demo receipt is not reachable before Aug 9.

Variant B is the contingency the W3 bundle already uses: a real receipt signed
with a locally generated Ed25519 key, verifiable offline by the public
`aragora-verify` package, carrying a dated statement that the local key is not
the pinned production identity.

## Why the amendment spells out the claim boundary

So the demo cannot quietly overclaim once it is being shown to people:

- **Proven** — the mechanism end to end: question → adversarial debate →
  dissent-preserving receipt with attributed cruxes → third-party offline
  verification, under 10 minutes. None of it depends on which key signed the
  bytes, and it is the whole of the Crucible-orphan pitch.
- **Not proven** — issuer authenticity. A local key attests the bytes are
  intact, not that Aragora vouched for them.
- **Upgrade path** — signatures are detached, so on reinstatement the same ODR
  bytes can be re-signed with the production key without changing the content
  digest, and the published artifact stays valid.

## What this decision does not unblock

Recorded in the amendment so W4 readiness is not overstated:

- **Trailing-week product share must reach ≥40%.** Measured **28%** on
  2026-07-28 — down from 33% earlier the same day, because the merge-tooling and
  crux-internals fixes that landed classify as substrate, not product.
- **The W3 bundle must publish**, which needs the founder's earned-claim review.

Plan document only. No code, no bundle artifacts.

### DIFF
--- a/docs/plans/2026-07-09-thirty-day-external-proof-month.md
+++ b/docs/plans/2026-07-09-thirty-day-external-proof-month.md
@@ -139,6 +139,43 @@ at 0%.
 - **Founder:** demo distribution decision (Crucible-orphan outreach), pentest
   kickoff, monthly budget re-tune.
 
+**AMENDMENT 2026-07-28 — founder decision: the demo ships Variant B.**
+
+"signed dissent-preserving receipts" above was written assuming production
+signing. The production Ed25519 key lives in AWS Secrets Manager in the
+suspended account (#9391), and the operator has stated reinstatement cannot be
+resolved on the required timeline, so a production-signed demo receipt is not
+reachable before Aug 9.
+
+The founder has decided the W4 demo ships **Variant B** — the same contingency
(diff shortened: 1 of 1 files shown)


scarmani:

Head: e41ca09 (e41ca09b795d94c0b3459a916897e6081960e136), committed 2026-07-29T19:47:02Z.
PR: #9680.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #9320, opened 2026-07-15
fix(sharing): warn on legacy share model access

scarmani:
## Summary

- warn on each legacy `aragora.server.handlers.social.sharing.DebateVisibility` and `.ShareSettings` access with the fully qualified deprecated symbol
- preserve identity with `aragora.storage.share_models` and keep handler global lookups patchable through the legacy paths
- import the model symbols canonically from the social package initializer so importing `SharingHandler` alone emits no new sharing compatibility warning

## Validation

- `python3 -m pytest tests/handlers/social/test_sharing.py -q -k 'test_debate_visibility_legacy_symbol_warns_and_preserves_identity or test_share_settings_legacy_symbol_warns_and_preserves_identity or test_legacy_share_symbol_monkeypatch_reaches_handler'` -> 3 passed
- `python3 -m pytest tests/handlers/social/test_sharing.py -q --timeout=120` -> 115 passed
- `python3 -m mypy --ignore-missing-imports --follow-imports=skip --show-error-codes aragora/server/handlers/social/sharing.py aragora/server/handlers/social/__init__.py` -> no issues
- `make lint` -> passed
- `ruff format --check aragora/ tests/ scripts/` -> passed
- `bash $MISSION/scripts/typecheck_differential.sh` -> pass, `new=103 <= 108`
- `python3 scripts/ci/check_import_contracts.py --layers foundation,infrastructure` -> no new violations
- `make test-smoke` -> Core/Server/Memory imports OK
- `PYTEST_BIN="python3 -m pytest" bash scripts/test_tiers.sh smoke` -> 138 passed
- `bash scripts/automation_pr_preflight.sh origin/main HEAD` -> passed

## Risk

Tier 3 because the diff touches `aragora/server/handlers/**`. This PR changes only compatibility access semantics and focused tests. It does not alter sharing permissions, visibility decisions, storage behavior, or persistence behavior. Human settlement remains required after a clean two-family exact-head review packet.

## V1 scope expansion

`ARAGORA_ALLOW_SCOPE_EXPANSION=1`

Rationale: `VAL-P4A-025` explicitly requires symbol-level deprecation compatibility and monkeypatch propagation on `aragora.server.handlers.social.sharing`, so the two handler-package files in this bounded diff are the required implementation surface. No endpoint, permission, visibility, persistence, or sharing semantics are expanded.

## Acceptance

Implements `VAL-P4A-025`.

PR #9729, opened 2026-08-09
test(governance): close contract drift sequence regressions

scarmani:
## Summary

- add the 15 exact VAL-CDG-017 regression selectors
- pin PR #9320 historical-nonconforming, backfill capsule, provenance, and first-parent semantic facts
- enforce the constants → matcher → Stage 1 → Stage 2 → corrective → backfill → route-core → OpenAPI-rearm chronology
- keep the suite hermetic, with full-clone Git recomputation and shallow-clone checked-in fact assertions

## Validation

- exact collection: 15/27 collected, 12 deselected
- exact execution: 15 passed, 12 deselected
- `make lint`
- `ruff format --check aragora/ tests/ scripts/`
- `bash scripts/preflight_mypy.sh --diff-base origin/main`
- `make test-smoke`
- `PYTEST_BIN='python3 -B -m pytest' bash scripts/test_tiers.sh smoke` (138 passed)

## Risk

Tests-only Tier 0 change. No workflow or production-source changes.

### DIFF
--- a/tests/scripts/test_contract_drift_sequence.py
+++ b/tests/scripts/test_contract_drift_sequence.py
@@ -1,0 +1,645 @@
+"""VAL-CDG-017 historical backfill and route-truth sequence regressions.
+
+The tests deliberately use checked-in records and immutable local Git objects,
+not live GitHub API calls. The later ``route_truth`` sealer owns live capsule
+re-authentication; this file pins the facts it must rediscover.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import subprocess
+from functools import lru_cache
+from pathlib import Path
+from typing import TypedDict
+
+
+class AssetBinding(TypedDict):
+    api_id: int
+    byte_length: int
+    name: str
+    sha256: str
+
+
+class PatchBinding(TypedDict):
+    byte_length: int
+    sha256: str
+
+
+class CorrectiveCapsuleBinding(TypedDict):
+    end_sha: str
+    release_id: int
+    tag: str
+
+
+class SequenceEntry(TypedDict):
+    fulfills: tuple[str, ...]
+    merge_sha: str
+    pr: int
+    role: str
+
+
+ROOT = Path(__file__).resolve().parents[2]
+BACKFILL_RECORD = ROOT / "docs/audits/2026-08-01-pr9320-historical-first-parent-backfill.md"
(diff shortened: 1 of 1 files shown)


scarmani:

Head: ef095cf (ef095cfc75a754f8c15f5be715580d4b455ca80e), committed 2026-08-09T05:12:33Z.
PR: #9729.
Model family: claude

Verdict: PASS

Verification performed: worktree is at head `ef095cfc75`; all four referenced data files exist (`docs/audits/2026-08-01-pr9320-historical-first-parent-backfill.md`, `scripts/baselines/contract_drift_inventory.json`, `.github/workflows/openapi.yml`, `scripts/validate_openapi_routes.py`); ran the full file — **15/15 tests pass in 1.24s** in a full-history clone, meaning the git-guarded branches (merge parentage, tree SHAs, first-parent patch digest, `rev-list --first-parent` ordering) executed live, not just the checked-in-record assertions.

- [P3] `tests/scripts/test_contract_drift_sequence.py:513` (and the analogous block in `test_route_core_merges_before_openapi_rearm`): the `_all_commits_present` guard checks object *existence* (`cat-file -e`), not first-parent *reachability* from HEAD, but the body then calls `first_parent.index(sha)`. Worktrees share one object store, so on a checkout where the pinned merges exist as objects but are not on HEAD's first-parent chain (e.g., this test file cherry-picked onto an older base), the test dies with an unhandled `ValueError` instead of skipping cleanly. Unlikely in normal flow (any branch containing this file descends from the pinned merges), hence non-blocking; a membership check before `.index()` would make the guard honest.
- [P3] Several assertions are tautologies comparing a module constant to a literal copy of itself in the same file (`BACKFILL_RECEIPT_EXECUTION`, `BACKFILL_RULE_SUITE`, `FIRST_PARENT_PATCH`, `SEMANTIC_PATHS`, `CORRECTIVE_CAPSULE["release_id"]`, and `BACKFILL_TAG == f"backfill-{PR_9320_MERGE}"` which restates the definition verbatim). They add edit-friction but zero verification power; the real checks are the ones recomputing hashes from `contract_drift_inventory.json`, the audit record, and git objects — which do hold at head.
- Security posture is clean: all subprocess calls use fixed argv lists (no `shell=True`), `check=True` where output is consumed, and the file makes no network or GitHub API calls, matching its stated design of pinning to checked-in records and immutable local git objects.

dogfood: yes

PR #9756, opened 2026-08-14
feat(nomic): add generic repository profiles

scarmani:
## Summary
- add typed repository identity, revision, evaluation criterion, evidence reference, and context pack models
- load and hash the `nomic` section from repository-owned or external `.aragora.yaml` files
- fail closed on invalid paths, duplicate criteria, dirty repositories, missing/untracked files, and escaping symlinks

## Risk
Tier 3: additive public planning interfaces and provenance semantics. Requires explicit human risk settlement before landing.

## Verification
- `pre-commit run --files aragora/nomic/repository_profile.py aragora/nomic/__init__.py tests/nomic/test_repository_profile.py`
- `pytest -q tests/nomic/test_repository_profile.py tests/nomic/test_context_builder.py` (25 passed)
- `mypy --follow-imports=skip aragora/nomic/repository_profile.py`

Part 1 of the stacked Generalize Nomic Repository Planning implementation. No legacy execution, governance, workflow, auth, or merge-authority behavior changes.

PR #9757, opened 2026-08-14
feat(nomic): emit commit-addressed context packs

scarmani:
## Summary
- add `NomicContextBuilder.build_context_pack()` while preserving legacy context APIs
- source all evidence from exact Git blobs and enumerate only tracked files
- emit deterministic, atomic packs with portable evidence IDs, blob/SHA-256 metadata, manifests, optional corpus, and artifact verification
- recheck clean HEAD immediately before publication and reject drift

## Risk
Tier 3 stacked planning/provenance API change. Requires explicit human risk settlement before landing.

## Verification
- `pre-commit run --files aragora/nomic/context_builder.py tests/nomic/test_context_pack.py`
- `pytest -q tests/nomic/test_context_pack.py tests/nomic/test_context_builder.py tests/nomic/test_nomic_phases.py` (50 passed)
- `mypy --follow-imports=skip aragora/nomic/context_builder.py`

Stacked on #9756. No legacy execution, governance, workflow, auth, or merge-authority behavior changes.


scarmani:

Head: 3465518 (3465518b55a5c1aae7273c4552a21060afa8934d), committed 2026-08-18T19:12:52Z.
PR: #9757.
Model family: claude

Verdict: PASS

Verified: read the full `context_builder.py` at head 3465518 (the truncated `_render_pack_context` hunk and surrounding code) plus `repository_profile.py` (from #9756) that it depends on; ran `tests/nomic/test_context_pack.py` + `tests/nomic/test_context_builder.py` → 47 passed; confirmed the aragora `.gitignore` ignores `.nomic/`. No blocking findings.

- [P3] `aragora/nomic/context_builder.py:268` — `SKIP_DIRS` is still matched against absolute `path.parts` while the test-dir filter on the next line was switched to `relative_path.parts`. A checkout living under a directory named `build`/`dist`/`venv` (e.g. `~/build/aragora`) yields an empty index. Pre-existing, but the PR touched the adjacent line and left the inconsistency; use `relative_path.parts` for both.
- [P3] `aragora/nomic/context_builder.py` `_collect_commit_evidence` (`git ls-tree … check=True`) and `_read_commit_blobs` (`cat-file --batch … check=True`) — `verify_context_pack` on a pack whose commit isn't in the local object store (other clone, gc'd, or forged `revision`) escapes as `subprocess.CalledProcessError`, not `RepositoryStateError`, breaking the error contract every other failure path in this module honours. Verified by reading; wrap in `RepositoryStateError`.
- [P3] `verify_context_pack` mutates builder state (`self._index = self._index_from_evidence(...)`) because `_render_pack_rlm_summary` reads `self._index` instead of taking the index as a parameter. Verifying a pack with a different `include_tests` policy silently replaces the builder's cached index for later `build_index()`/`build_rlm_context()` calls. Pass the index explicitly.
- [P3] Cost/blocking: `_collect_commit_evidence` reads every eligible blob for the whole tree into memory and hashes it — measured on this repo via `git ls-tree -l` with the PR's filters: 15,787 files / ~266 MB (tests ~138 MB), plus a ~100 MB corpus string; and the reuse path (`destination.exists()`) re-does the full read on every call. All via synchronous `subprocess.run` inside `async def`. Fine for the nomic loop process; will stall a shared event loop if this is later wired into a server handler.
- [P3] Behaviour change in `build_index()`: when the path is a git top-level, `git ls-files` is authoritative and the `codebase_manifest.tsv` cache / `ARAGORA_CONTEXT_USE_MANIFEST` become effectively dead (only used when git enumeration fails); untracked working-tree files are no longer indexed. Tests show this is intentional, but it should be stated in the PR body since `scripts/nomic_loop.py`, `aragora/debate/codebase_context.py`, and `aragora/rlm/codebase_context.py` all consume this index.

dogfood: yes

scarmani:

Head: 3465518 (3465518b55a5c1aae7273c4552a21060afa8934d), committed 2026-08-18T19:12:52Z.
PR: #9757.
Model family: openai

Verdict: PASS

- [P3] `aragora/nomic/context_builder.py:428` The symlink safety checks are TOCTOU-sensitive: `parent.mkdir()` / `mkdtemp(dir=parent)` can still follow a `.nomic` component swapped to a symlink after `_assert_pack_destination_safe()` runs, creating directories or temp artifacts outside the repo before the next check fails. This requires concurrent local mutation, so non-blocking, but fully closing it would require fd-based/no-follow directory creation or revalidating immediately around each filesystem operation.

dogfood: yes

PR #10107, opened 2026-09-18
feat(odr): enforce v0.2 verifier consistency and signature policies

scarmani:
## Summary
- Add matching ODR v0.2 consistency checks to both offline verifiers. Invalid verdict issuers fail `verdicts_consistency`; supporting/dissenting membership fails `quorum_consistency`; severity/blocking summaries fail `dissent_consistency`. These checks precede cryptographic failures.
- Warn on rule/reached mismatch and verified-signature expiry. Add `--now`, `--strict-expiry`, and signer-committed `--require-issuer` to the standalone CLI. Foreign signature entries emit `key_id_mismatch`; a matching-key failure still wins over another matching-key success.
- Add JSON `dissent_trail`/`key_id`, the human dissent/adjudication section and keyed footer, and accurate version labels. Normalize explicit UTC timestamps before signing. Update the specification's previously deferred expiry-policy paragraph.
- Preserve existing schema scoping, signature constructions, package version (0.1.2), v0.1 fixture bytes, and canonical digests.

## Exit metric moved
Row 3 dependency: the staged ODR v0.2 profile gains verifier consistency, expiry/issuer policy and a readable dissent trail. No row-3 scoreboard movement is claimed here; `m1-odr-spec-v02-ietf-mapping` owns that movement. Committed vectors remain a separate feature.

## Test commands
- Initial implementation RED: new verifier tests failed on missing policy flags, consistency precedence, mismatch warnings, and the v0.1-only label; the empty adjudicated-trail case also failed before its fix.
- Collection-1 repair RED: malformed package quorum shapes raised exceptions without jsonschema; appended unverified expired signatures failed strict expiry; unposted merge-quorum signals emitted a false mismatch warning. Focused GREEN: **119 passed**.
- On repair head `c73e36214277841419ab1b196bf7bf0f31d84c20`, `timeout 290 python3 -m pytest aragora-verify/tests tests/gauntlet tests/verification -q -n 8 -p no:cacheprovider --timeout=120`: **2985 passed, 3 skipped**.
- `bash "$MISSION_DIR/bin/gate.sh" lint|typecheck|contracts|test|guardrails` (each separately, bounded at 290 seconds): all **GATE PASSED** on the repair head. Curated tests: **4310 passed, 8 skipped**; publish shadow: **207 passed, 29 skipped** (optional Aragora/jsonschema tests), also run separately; skip markers **95**, unchanged from base; mypy count **1695 <= 1744**.
- Both schema files compare byte-identical; extension-walker prefixes identical; the 24-row ruling-9 matrix passes after repair. Dependency policy passes, with no runtime dependency added. Preflight passes; formatted whole-PR delta **800 LOC**.
- Retained real-file proof on initial head `14fc59b9d6`: all **16** v0.1 receipts keep exact exit codes and digests; signed golden receipt/three chain cases remain **0 / 0,1,1**. Real #9949 mutations fail under the pinned check names in both engines. Signed P1/P2 adjudicated receipts show one `Dissent trail`, severity labels and keyed footer; expiry **0/default, 1/strict, 0/before, 1/after**; missing/nonverifying issuer **1**.
- Retained offline demo exports: default unsigned **0.1**, requested unsigned **0.2**, mission-key signed **0.2** all verify; signed without key **3**, body tamper **1**, missing signing file **1/no output**, empty key env unsigned.
- Retained clean-venv proof: built and installed the exact initial-head **0.1.2** wheel without Aragora; its CLI verified the signed v0.2 fixture and issuer requirement. Docs sync/build passed before collection 1; the repair changes no docs.

## Reviewed design tradeoffs
- The default stays 0.1 until aragora-verify 0.2.0 is on PyPI; the flip is `m2-odr-default-v02` (invariant §5.2b). No version bump, default flip, ACTA implementation, or committed vector generation belongs here.
- Gate dissent and severity findings remain independent. A PASS review may carry advisory findings. The new severity check does not derive `present`, `dissenting_agents`, or verdict-level blocking.
- Rule checks use recorded `counted_families`, required count and frontier requirement, and recognize recorded gate dissent. For `method: merge-quorum`, the rule is only a necessary condition: posting state is not fully recorded, so a recorded false result does not generate a positive-inference warning. No live gate is imported and no recorded result is changed.
- Expiry is policy over signer-committed metadata in entries that verify under the supplied key. It warns by default and fails in strict mode, including an unparseable expiry; absence stays valid. Foreign/unverified entries cannot inject expiry warnings or strict-expiry failures. v0.1 metadata never establishes an issuer requirement.
- The standalone package remains stdlib plus cryptography. Mirrored checks avoid a runtime dependency on Aragora. Existing v0.2 positive test fixtures now use matching summaries and participant issuers; all v0.1 example files are unchanged.

## Risks
- Strict expiry is opt-in. Verification still does not establish whether the underlying decision is correct.
- PyPI 0.1.1 rejects requested v0.2 receipts until the separately owned 0.2.0 release; the local version is intentionally not bumped here.
- The non-required metrics-drift check is already red on main. New parametrized tests change source counts; no metrics or guardrail baseline is raised.
- Collection-1 advisory P3s (standalone `reached` typing without jsonschema and per-finding `blocking`/severity symmetry) were not expanded into the bounded repair. Collection 2 raised the latter as P2; the repeated observation is confirmed and now awaits an orchestrator ruling at the 800-LOC cap.

## Collection-1 repair
Both reviewers identified expiry applied to unverified entries. Claude also identified malformed package quorum crashes and false rule/reached warnings on valid merge-quorum output. One tests-first batch, `c73e36214277841419ab1b196bf7bf0f31d84c20`, fixes all three P2 issues: shape validation before consistency/trail processing, verified-entry-only expiry in both engines, and conservative rule warnings when posting state is absent. The two advisory P3s remain recorded above. Collection 1 used native Claude `max-02` with `ANTHROPIC_MODEL=claude-opus-5`; both bodies were grounded CHANGES-REQUESTED, no transport failure.

## Collection-2 disposition
Claude returned grounded countable PASS; OpenAI returned grounded CHANGES-REQUESTED P2 on per-finding `blocking`/severity consistency, repeating Claude's collection-1 P3. A signed P1 finding with `blocking: false` and consistent aggregate summaries passes both engines; this observation is not refuted. The PR is parked for an orchestrator ruling on a cap-respecting repair/split. Two of three collections are spent, with no transport failures. No ready transition, evidence apply or merge; no unchanged-head reroll. Claude again used `max-02` and `ANTHROPIC_MODEL=claude-opus-5`.

## Ruling-40 repair
Ruling 40 resolved the collection-2 park by authorizing one bounded repair under the same 800-LOC cap. `f543dce557953280b7efc2b13c642a056aa27717` makes both engines check each `quorum.dissent.findings[i].blocking` against that finding's own severity, inside the existing findings block and before the aggregate loop, byte-identically, under the already-pinned `dissent_consistency` name; no new check name enters either engine and v0.1 documents are unaffected. Three regressions cover it: `test_v02_per_finding_blocking_and_v01` and the parametrized `quorum.dissent.findings.0.blocking` mutant row (both run through the `engine` fixture, so both engines), plus `test_v02_per_finding_blocking_consistency` in the package suite. To stay inside the cap, `test_v02_signature_precedence_and_v01` (a test this PR itself adds; the equivalent core test already runs both engines) and the CHANGELOG note move to `m1-aragora-verify-0-2-0`, and the one-sentence spec statement of the enforced consistency set moves to `m1-odr-spec-v02-ietf-mapping`. Measured formatted delta: +754/-44 = 798.

## Ruling-41 repair
Ruling 41 unparked the feature as attempt 2 for one more bounded repair. `6f3b3db75c0c79f6f2e6b5167aa8cea20eb847a5` makes the package's dependency-free fallback walker (`aragora-verify/src/aragora_verify/schema.py::_check_quorum`) type `quorum.reached` as a boolean and `quorum.independence` as an object, with the two diagnostics byte-identical to the in-repo walker's, so the two bundled engines agree with or without the optional `jsonschema` extra — the configuration an outsider gets from a plain `pip install aragora-verify`. The in-repo walker is unchanged (it was already correct) and no new check name enters either engine; the regression is folded into the existing stubbed `test_versions_and_unknown_members_without_jsonschema`, which asserts both mutations now fail with those exact diagnostics. To stay inside the cap, `test_v02_expiry_clock_and_v01_compatibility` (another test this PR itself adds; the expiry clock is also covered by the CLI test and by the core suite) joins the earlier deferrals to `m1-aragora-verify-0-2-0`, which also still carries the CHANGELOG note and Claude's two documentation P3s; the one-sentence spec statement stays with `m1-odr-spec-v02-ietf-mapping`. Measured formatted delta: +746/-44 = 790.

## Ruling-42 repair, and the answer to the round-4 crash finding
Ruling 42 authorized one more bounded repair on the same parity axis ruling 41 opened. `37c9f016f317d23167cc755f17648bacfc24210b` makes the package's dependency-free fallback walker type `quorum.method` as a string, with the diagnostic `quorum.method: must be a string` byte-identical to the in-repo walker's at `aragora/gauntlet/odr_verify.py:302`, so the two bundled engines now agree on `method`, `reached` and `independence` with or without the optional `jsonschema` extra. The in-repo walker is unchanged and no new check name enters either engine; the regression is one more mutation row in the same stubbed `test_versions_and_unknown_members_without_jsonschema`. `quorum.supporting_agents` is the one known remaining divergence on this axis and goes to `m1-aragora-verify-0-2-0` with the same byte-identical-diagnostic requirement. Measured formatted delta: +752/-44 = 796.

OpenAI's round-4 P2 — that the v0.2 consistency pass can crash on schema-invalid-but-JSON-valid nested quorum data because it indexes `p["agent"]`, `v["issuer"]`, `finding["severity"]` and `finding["blocking"]` outside `_safe_check` — was refuted by execution on `6f3b3db75c`, and no repair was made for it. 22 mutations covering every indexing site the finding names, run through both engines on the dependency-free path, gave 44 runs and 0 crashes: each mutation is stopped by `schema_conformance` before the consistency pass ever runs, because both engines return early on any structure error (`verifier.py:568-577`, `odr_verify.py:1067-1070`) and only then call `_check_v02_consistency`. The boundary comment at `odr_verify.py:1073-1078` states that invariant verbatim and it holds at this head, so the reads are outside `_safe_check` by design. This is also the restatement of collection 1's Claude P2, which was repaired tests-first in `c73e36214277841419ab1b196bf7bf0f31d84c20`.


scarmani:
## Collection summary

Summary: 6 advisory findings. Severity labels only; not a merge decision.
Gate regime: severity-gated dissent ON; P2/P3 findings are advisory to the merge gate.
Merge-gate dissent: none.

### Family outcomes
- claude: changes_requested
- openai: changes_requested

### Adjudication

> none

### Collection action

> prepare
> dry-run; re-run with --apply to post

### Findings

- [P2] claude (advisory): `aragora-verify/src/aragora_verify/verifier.py:323-338` and `:397-400` (`_check_v02_consistency`, `_dissent_trail`) — Standalone verifier now crashes on malformed v0.2 input when the optional `jsonschema` extra is absent. The lightweight walker (`schema.py:_check_quorum`) never types `quorum.dissent` or `quorum.participants`; the new unguarded checks then do `"findings" in dissent` / `p["agent"]` / `dissent.get(...)`. Verified by stubbing `schema._jsonschema_errors` (same technique as `test_versions_and_unknown_members_without_jsonschema`): `dissent=5` → TypeError, `dissent="abc"` → AttributeError, `participants="abc"|{...}` → TypeError. `cli.main` only catches FileNotFound/OSError/JSON/VerificationError, so this is an uncaught traceback instead of exit 1/2. Regression: pre-PR `_check_quorum_consistency` isinstance-guards these. The core engine is unaffected (its `_validate_quorum` is stricter). Fix: type-check `dissent`/`participants` in the walker, or guard the new checks like the legacy ones.
- [P2] claude (advisory): `aragora-verify/src/aragora_verify/verifier.py` `_check_expiry` (~line 370) and `aragora/gauntlet/odr_verify.py` `_check_expiry` (~line 916) — Expiry is evaluated on every `signatures[]` entry, including ones whose `key_id` doesn't match the supplied key and whose signature never verified. Because `signatures` sit outside the content digest, anyone can append `{"alg":"Ed25519","key_id":"ed25519-…","signature":"c2ln","expires_at":"2000-…"}` to a valid receipt and flip `--strict-expiry` to FAIL (exit 1), or inject a misleading "expired" warning without any key. Verified on both engines: valid unexpired sig (expires 2999) + appended foreign expired entry → `signature=pass`, `signature_expiry=fail`, `ok=False`. Contradicts the spec sentence in this same diff ("only entries whose `key_id` matches the supplied key count"). Expiry should apply only to verified/matching entries (the `verified` list already exists in the standalone engine).
- [P2] claude (advisory): both `_check_v02_consistency` `quorum_rule` blocks — The inferred `reached` (`len(counted_families) &gt;= required_signals` + frontier) does not match how the reference producer sets it: `aragora/swarm/quorum_receipt.py:159` has `reached = action == "post" and posted_quorum and not dissenting`, and `counted_families` is built from `counting_families` (`would_count`, which includes dissenters; `quorum_evidence.py:1006`). A legitimately produced receipt with two counted families and one dissenter therefore warns `quorum.reached: recorded False, rule implies True` — a false tamper-flavoured signal on producer output. Verified by constructing such a doc on both engines. It also ignores `requires_at_least_one_western` (`quorum_evidence.py:314`), which isn't recorded in `rule`. Either factor in `dissent.present`/`dissenting_agents` (already in the doc) or downgrade the check to "rule not satisfied but reached=True" only.
- [P2] openai (advisory): `aragora-verify/src/aragora_verify/verifier.py:_check_expiry` and `aragora/gauntlet/odr_verify.py:_check_expiry`: expiry is evaluated over every `doc["signatures"]` entry, not only the signature(s) that verified with the supplied key. Because signatures are detached from the content digest, an attacker can append a bogus or mismatched expired signature and make `--strict-expiry` fail an otherwise valid receipt. This should be scoped to verified/key-matching v0.2 signatures, or at minimum ignore entries that do not verify under the provided key.
- [P3] claude (advisory): `schema.py` walker parity — Without `jsonschema`, standalone accepts `quorum.reached: "yes"` on v0.2 (emits `recorded yes, rule implies True`) while the core engine rejects it at schema_conformance. Verified via the same stubbed run.
- [P3] claude (advisory): `_dissent_trail` / `_check_v02_consistency` — Per-finding `blocking` is never cross-checked against `severity` (spec §: "`blocking` (P0/P1 true)"), and the trail label is derived from severity, ignoring `finding.blocking`. A P2 finding with `blocking: true` passes consistency while `dissent.blocking: true` on the same doc FAILs. Harmless with the reference producer (both derived from severity), but the check is asymmetric.

### Output 1

> &gt; ## Claude i&#110;dependent model review
> 
> &gt; Reviewer&#58; claude (anthropic) — independent a&#100;versarial model review via the Aragora C&#108;aude reviewer, grounded on the exact PR head.
> Head: 14fc59b (14fc59b9d6b310b7a72917348f559db55cc96eb6), committed 2026-09-18T01:47:58Z.
> PR: #10107.
> &gt; Model family&#58; claude
> 
> Verdict: CHANGES-REQUESTED
> 
> - (P2) `aragora-verify/src/aragora_verify/verifier.py:323-338` and `:397-400` (`_check_v02_consistency`, `_dissent_trail`) — Standalone verifier now crashes on malformed v0.2 input when the optional `jsonschema` extra is absent. The lightweight walker (`schema.py:_check_quorum`) never types `quorum.dissent` or `quorum.participants`; the new unguarded checks then do `"findings" in dissent` / `p["agent"]` / `dissent.get(...)`. Verified by stubbing `schema._jsonschema_errors` (same technique as `test_versions_and_unknown_members_without_jsonschema`): `dissent=5` → TypeError, `dissent="abc"` → AttributeError, `participants="abc"|{...}` → TypeError. `cli.main` only catches FileNotFound/OSError/JSON/VerificationError, so this is an uncaught traceback instead of exit 1/2. Regression: pre-PR `_check_quorum_consistency` isinstance-guards these. The core engine is unaffected (its `_validate_quorum` is stricter). Fix: type-check `dissent`/`participants` in the walker, or guard the new checks like the legacy ones.
> 
> - (P2) `aragora-verify/src/aragora_verify/verifier.py` `_check_expiry` (~line 370) and `aragora/gauntlet/odr_verify.py` `_check_expiry` (~line 916) — Expiry is evaluated on every `signatures[]` entry, including ones whose `key_id` doesn't match the supplied key and whose signature never verified. Because `signatures` sit outside the content digest, anyone can append `{"alg":"Ed25519","key_id":"ed25519-…","signature":"c2ln","expires_at":"2000-…"}` to a valid receipt and flip `--strict-expiry` to FAIL (exit 1), or inject a misleading "expired" warning without any key. Verified on both engines: valid unexpired sig (expires 2999) + appended foreign expired entry → `signature=pass`, `signature_expiry=fail`, `ok=False`. Contradicts the spec sentence in this same diff ("only entries whose `key_id` matches the supplied key count"). Expiry should apply only to verified/matching entries (the `verified` list already exists in the standalone engine).
> 
> - (P2) both `_check_v02_consistency` `quorum_rule` blocks — The inferred `reached` (`len(counted_families) &gt;= required_signals` + frontier) does not match how the reference producer sets it: `aragora/swarm/quorum_receipt.py:159` has `reached = action == "post" and posted_quorum and not dissenting`, and `counted_families` is built from `counting_families` (`would_count`, which includes dissenters; `quorum_evidence.py:1006`). A legitimately produced receipt with two counted families and one dissenter therefore warns `quorum.reached: recorded False, rule implies True` — a false tamper-flavoured signal on producer output. Verified by constructing such a doc on both engines. It also ignores `requires_at_least_one_western` (`quorum_evidence.py:314`), which isn't recorded in `rule`. Either factor in `dissent.present`/`dissenting_agents` (already in the doc) or downgrade the check to "rule not satisfied but reached=True" only.
> 
> - (P3) `schema.py` walker parity — Without `jsonschema`, standalone accepts `quorum.reached: "yes"` on v0.2 (emits `recorded yes, rule implies True`) while the core engine rejects it at schema_conformance. Verified via the same stubbed run.
> 
> - (P3) `_dissent_trail` / `_check_v02_consistency` — Per-finding `blocking` is never cross-checked against `severity` (spec §: "`blocking` (P0/P1 true)"), and the trail label is derived from severity, ignoring `finding.blocking`. A P2 finding with `blocking: true` passes consistency while `dissent.blocking: true` on the same doc FAILs. Harmless with the reference producer (both derived from severity), but the check is asymmetric.
> 
> Ran `aragora-verify/tests` + `tests/gauntlet/test_odr_*.py`: 699 passed — none of the above are covered by the new tests.
> 
> d&#111;gfood: yes

### Output 2

> &gt; ## OpenAI i&#110;dependent model review
> 
> &gt; Reviewer&#58; openai (openai) — independent a&#100;versarial model review via Codex CLI OpenAI harness, grounded on the exact PR head.
> Head: 14fc59b (14fc59b9d6b310b7a72917348f559db55cc96eb6), committed 2026-09-18T01:47:58Z.
> PR: #10107.
> &gt; Model family&#58; openai
> 
> Verdict: CHANGES-REQUESTED
> 
> - (P2) `aragora-verify/src/aragora_verify/verifier.py:_check_expiry` and `aragora/gauntlet/odr_verify.py:_check_expiry`: expiry is evaluated over every `doc["signatures"]` entry, not only the signature(s) that verified with the supplied key. Because signatures are detached from the content digest, an attacker can append a bogus or mismatched expired signature and make `--strict-expiry` fail an otherwise valid receipt. This should be scoped to verified/key-matching v0.2 signatures, or at minimum ignore entries that do not verify under the provided key.
> 
> d&#111;gfood: yes

scarmani:
## Collection summary

Summary: 5 advisory findings. Severity labels only; not a merge decision.
Gate regime: severity-gated dissent ON; P2/P3 findings are advisory to the merge gate.
Merge-gate dissent: none.

### Family outcomes
- claude: pass
- openai: changes_requested

### Adjudication

> none

### Collection action

> prepare
> dry-run; re-run with --apply to post

### Findings

- [P2] openai (advisory): `aragora-verify/src/aragora_verify/verifier.py:338` and `aragora/gauntlet/odr_verify.py:883` - v0.2 consistency checks recompute aggregate `dissent.blocking`, but never validate each `findings[].blocking` against `findings[].severity`. I verified both verifiers return `ok=True` for a receipt with a `P1` finding marked `blocking: false` as long as top-level `dissent.blocking` is `true`, contrary to the v0.2 spec text that finding `blocking` is true for P0/P1. This lets internally inconsistent/tampered dissent metadata verify successfully.
- [P3] claude (advisory): `aragora-verify/src/aragora_verify/verifier.py:379-395` (and `odr_verify.py:927`): `strict_expiry` FAILs if *any* verified entry is expired, so a receipt re-signed with a fresh entry (`replace=False` appends) still fails once the old entry lapses. Verified by reading `_check_expiry`; if the intent is "at least one unexpired verifying signature", the loop should aggregate rather than fail per entry. Design call, not a bug.
- [P3] claude (advisory): `verifier.py:359-370` / `odr_verify.py:898-914`: `quorum_rule` forces `reached=False` whenever `dissent.present`/`dissenting_agents` is set, so an `advisory_settle` receipt (`ARAGORA_ENABLE_ADVISORY_DISSENT_SETTLE`, `quorum_evidence.py:232-250`) that records `reached: true` would warn. Unverified whether the exporter actually emits `reached: true` with `present: true` on that path; an end-to-end export test through `merge_quorum_reconcile` would settle it. WARN-only either way.
- [P3] claude (advisory): `aragora-verify/src/aragora_verify/cli.py:53-54`: `_render` prints `Dissent trail` / `(no dissent recorded)` even for a schema-FAIL result where the trail was never computed — mildly misleading; gate the section on `result.odr_digest` or `ok`.
- [P3] claude (advisory): `aragora/gauntlet/odr_signing.py:375,380`: signer now emits `+00:00` instead of the caller's `Z`; both are RFC 3339 and tests were updated, but any external consumer that string-compared `Z` timestamps in signed entries would see a change. Nothing in-repo depends on it (grepped `signed_at"] ==` across `tests/`).

### Output 1

> &gt; ## Claude i&#110;dependent model review
> 
> &gt; Reviewer&#58; claude (anthropic) — independent a&#100;versarial model review via the Aragora C&#108;aude reviewer, grounded on the exact PR head.
> Head: c73e362 (c73e36214277841419ab1b196bf7bf0f31d84c20), committed 2026-09-18T07:38:51Z.
> PR: #10107.
> &gt; Model family&#58; claude
> 
> Verdict: PASS
> 
> Verification performed: read `schema.py`, `odr_schema.json`, `odr_verify.py` (`_validate_quorum`/`_validate_dissent`/`_validate_extensions`), `odr_signing.py` (`_parse_rfc3339`), `quorum_evidence.py` and `quorum_receipt.py`; ran `pytest aragora-verify/tests tests/gauntlet -k "odr or verif or cli or schema"` (761 passed) and `tests/scripts/test_build_disagreement_atlas.py` (38 passed, the only other `sign_odr_receipt` caller).
> 
> - Confirmed safe: every member the unguarded `_check_v02_consistency` / `_check_expiry` / `_dissent_trail` index (`verdicts[].issuer`, `findings[].severity/issuer/text`, `rule.*`, `adjudication.verdict/reason`, `participants[].agent`, `signatures[].expires_at`) is type-checked by the schema-driven `member()` walker in both verifiers even without `jsonschema`, so the "no crash on malformed input" boundary holds. `{"claude","openai"}` matches `WESTERN_FRONTIER_FAMILIES`; the `severity_max`/`blocking` derivation matches the producer in `quorum_receipt.py:101-104`.
> - (P3) `aragora-verify/src/aragora_verify/verifier.py:379-395` (and `odr_verify.py:927`): `strict_expiry` FAILs if *any* verified entry is expired, so a receipt re-signed with a fresh entry (`replace=False` appends) still fails once the old entry lapses. Verified by reading `_check_expiry`; if the intent is "at least one unexpired verifying signature", the loop should aggregate rather than fail per entry. Design call, not a bug.
> - (P3) `verifier.py:359-370` / `odr_verify.py:898-914`: `quorum_rule` forces `reached=False` whenever `dissent.present`/`dissenting_agents` is set, so an `advisory_settle` receipt (`ARAGORA_ENABLE_ADVISORY_DISSENT_SETTLE`, `quorum_evidence.py:232-250`) that records `reached: true` would warn. Unverified whether the exporter actually emits `reached: true` with `present: true` on that path; an end-to-end export test through `merge_quorum_reconcile` would settle it. WARN-only either way.
> - (P3) `aragora-verify/src/aragora_verify/cli.py:53-54`: `_render` prints `Dissent trail` / `(no dissent recorded)` even for a schema-FAIL result where the trail was never computed — mildly misleading; gate the section on `result.odr_digest` or `ok`.
> - (P3) `aragora/gauntlet/odr_signing.py:375,380`: signer now emits `+00:00` instead of the caller's `Z`; both are RFC 3339 and tests were updated, but any external consumer that string-compared `Z` timestamps in signed entries would see a change. Nothing in-repo depends on it (grepped `signed_at"] ==` across `tests/`).
> 
> d&#111;gfood: yes

### Output 2

> &gt; ## OpenAI i&#110;dependent model review
> 
> &gt; Reviewer&#58; openai (openai) — independent a&#100;versarial model review via Codex CLI OpenAI harness, grounded on the exact PR head.
> Head: c73e362 (c73e36214277841419ab1b196bf7bf0f31d84c20), committed 2026-09-18T07:38:51Z.
> PR: #10107.
> &gt; Model family&#58; openai
> 
> Verdict: CHANGES-REQUESTED
> 
> - (P2) `aragora-verify/src/aragora_verify/verifier.py:338` and `aragora/gauntlet/odr_verify.py:883` - v0.2 consistency checks recompute aggregate `dissent.blocking`, but never validate each `findings[].blocking` against `findings[].severity`. I verified both verifiers return `ok=True` for a receipt with a `P1` finding marked `blocking: false` as long as top-level `dissent.blocking` is `true`, contrary to the v0.2 spec text that finding `blocking` is true for P0/P1. This lets internally inconsistent/tampered dissent metadata verify successfully.
> 
> d&#111;gfood: yes

scarmani:
## Collection summary

Summary: 6 advisory findings. Severity labels only; not a merge decision.
Gate regime: severity-gated dissent ON; P2/P3 findings are advisory to the merge gate.
Merge-gate dissent: none.

### Family outcomes
- claude: pass
- openai: changes_requested

### Adjudication

> none

### Collection action

> prepare
> dry-run; re-run with --apply to post

### Findings

- [P2] openai (advisory): `aragora-verify/src/aragora_verify/schema.py:91` — The standalone verifier’s dependency-free schema path still does not validate core quorum types such as `quorum.reached` and `quorum.independence`, so malformed v0.2 receipts can get `schema_conformance pass` and overall `ok=True` when `jsonschema` is unavailable. I verified by monkeypatching `aragora_verify.schema._jsonschema_errors = lambda doc: []`, setting `valid_odr("0.2")["quorum"]["reached"] = "yes"` and `["independence"] = "not-an-object"`, and observing `validate_structure(doc) == []` plus `verify(doc).ok is True`. The in-tree gauntlet verifier already rejects these shapes, so the standalone mirror regresses/parity-breaks the offline verifier’s schema guarantees.
- [P3] claude (advisory): `aragora-verify/src/aragora_verify/verifier.py:250` (and `odr_verify.py:780`): every entry whose `key_id` differs from the supplied key now emits a `key_id_mismatch` warning into `result.warnings`. On a legitimate multi-signer v0.2 receipt (emitter + reviewer/notary roles are defined in the schema) verifying with one key, this produces one noisy warning per other signer and pollutes `--json` `warnings`, which previously held only weakening signals. Consider emitting it only when the signature also *verifies* under the supplied key (the genuine relabel case), or naming it distinctly.
- [P3] claude (advisory): `aragora-verify/src/aragora_verify/verifier.py:604-616`: `--require-issuer` is satisfied by a verified-but-expired signature unless `--strict-expiry` is also passed (expiry is only WARN by default). That's consistent with the documented default, but the docs sentence for `--require-issuer` doesn't say so; worth one clause so operators don't assume issuer-required implies unexpired.
- [P3] claude (advisory): `aragora/gauntlet/odr_signing.py:375,380`: explicit `signed_at`/`expires_at` are re-emitted via `isoformat()` with no `timespec`, so a caller-supplied fractional-second timestamp is written as `…00:00.500000+00:00` while the default path uses `timespec="seconds"`. Verifiers parse both fine; just an inconsistency in the committed metadata format.
- [P3] claude (advisory): `aragora-verify/src/aragora_verify/cli.py:54`: on a schema-conformance failure `dissent_trail` is the default `[]`, so the text renderer prints "(no dissent recorded)" under an unverifiable receipt. Cosmetic, but slightly misleading.
- [P3] claude (advisory): Unverified: I attempted a mutation probe (malformed `rule`, `verdicts`, `findings`, `adjudication`, absent-marker-with-members) against the unguarded standalone `verify()` to confirm no exception escapes, but the command was not permitted to run. Static reading of `_validate_extensions` indicates all such inputs return `schema_conformance` FAIL first; running that probe would close the gap.

### Output 1

> &gt; ## Claude i&#110;dependent model review
> 
> &gt; Reviewer&#58; claude (anthropic) — independent a&#100;versarial model review via the Aragora C&#108;aude reviewer, grounded on the exact PR head.
> Head: f543dce (f543dce557953280b7efc2b13c642a056aa27717), committed 2026-09-18T08:34:57Z.
> PR: #10107.
> &gt; Model family&#58; claude
> 
> Verdict: PASS
> 
> Verified: package suite (228 passed) and gauntlet ODR suites (307 passed) run green at f543dce. I read `aragora-verify/src/aragora_verify/schema.py` and the bundled `odr_schema.json` to confirm that every member the unguarded `_check_v02_consistency`/`_check_expiry`/`_dissent_trail` paths index (`participants[].agent`, `verdicts[].issuer`, `findings[].severity/blocking/issuer/text`, `rule.*`, `adjudication.verdict/reason`, `signatures[].expires_at`) is required+typed by `_validate_extensions`/`_check_quorum` before those functions run, on both engines. I also confirmed the producer (`aragora/swarm/quorum_receipt.py:60-105`) derives `blocking`, `severity_max`, verdict issuers, and `counted_families` vocabulary (`claude`/`openai`, `quorum_evidence.py:151`) exactly as the new consistency checks expect, so real merge-quorum receipts won't spuriously FAIL.
> 
> - (P3) `aragora-verify/src/aragora_verify/verifier.py:250` (and `odr_verify.py:780`): every entry whose `key_id` differs from the supplied key now emits a `key_id_mismatch` warning into `result.warnings`. On a legitimate multi-signer v0.2 receipt (emitter + reviewer/notary roles are defined in the schema) verifying with one key, this produces one noisy warning per other signer and pollutes `--json` `warnings`, which previously held only weakening signals. Consider emitting it only when the signature also *verifies* under the supplied key (the genuine relabel case), or naming it distinctly.
> - (P3) `aragora-verify/src/aragora_verify/verifier.py:604-616`: `--require-issuer` is satisfied by a verified-but-expired signature unless `--strict-expiry` is also passed (expiry is only WARN by default). That's consistent with the documented default, but the docs sentence for `--require-issuer` doesn't say so; worth one clause so operators don't assume issuer-required implies unexpired.
> - (P3) `aragora/gauntlet/odr_signing.py:375,380`: explicit `signed_at`/`expires_at` are re-emitted via `isoformat()` with no `timespec`, so a caller-supplied fractional-second timestamp is written as `…00:00.500000+00:00` while the default path uses `timespec="seconds"`. Verifiers parse both fine; just an inconsistency in the committed metadata format.
> - (P3) `aragora-verify/src/aragora_verify/cli.py:54`: on a schema-conformance failure `dissent_trail` is the default `[]`, so the text renderer prints "(no dissent recorded)" under an unverifiable receipt. Cosmetic, but slightly misleading.
> - (P3) Unverified: I attempted a mutation probe (malformed `rule`, `verdicts`, `findings`, `adjudication`, absent-marker-with-members) against the unguarded standalone `verify()` to confirm no exception escapes, but the command was not permitted to run. Static reading of `_validate_extensions` indicates all such inputs return `schema_conformance` FAIL first; running that probe would close the gap.
> 
> d&#111;gfood: yes

### Output 2

> &gt; ## OpenAI i&#110;dependent model review
> 
> &gt; Reviewer&#58; openai (openai) — independent a&#100;versarial model review via Codex CLI OpenAI harness, grounded on the exact PR head.
> Head: f543dce (f543dce557953280b7efc2b13c642a056aa27717), committed 2026-09-18T08:34:57Z.
> PR: #10107.
> &gt; Model family&#58; openai
> 
> Verdict: CHANGES-REQUESTED
> 
> - (P2) `aragora-verify/src/aragora_verify/schema.py:91` — The standalone verifier’s dependency-free schema path still does not validate core quorum types such as `quorum.reached` and `quorum.independence`, so malformed v0.2 receipts can get `schema_conformance pass` and overall `ok=True` when `jsonschema` is unavailable. I verified by monkeypatching `aragora_verify.schema._jsonschema_errors = lambda doc: []`, setting `valid_odr("0.2")["quorum"]["reached"] = "yes"` and `["independence"] = "not-an-object"`, and observing `validate_structure(doc) == []` plus `verify(doc).ok is True`. The in-tree gauntlet verifier already rejects these shapes, so the standalone mirror regresses/parity-breaks the offline verifier’s schema guarantees.
> 
> d&#111;gfood: yes

scarmani:
## Collection summary

Summary: 4 advisory findings. Severity labels only; not a merge decision.
Gate regime: severity-gated dissent ON; P2/P3 findings are advisory to the merge gate.
Merge-gate dissent: none.

### Family outcomes
- claude: pass
- openai: changes_requested

### Adjudication

> none

### Collection action

> prepare
> dry-run; re-run with --apply to post

### Findings

- [P2] openai (advisory): `aragora-verify/src/aragora_verify/verifier.py:_check_v02_consistency` and `aragora/gauntlet/odr_verify.py:_check_v02_consistency`: v0.2 verification can crash on schema-invalid-but-JSON-valid nested quorum data because the new consistency pass directly indexes `p["agent"]`, `v["issuer"]`, `finding["severity"]`, and `finding["blocking"]` before the existing guarded verification checks. The gauntlet verifier explicitly promises malformed receipts become FAIL, but this new pass is outside `_safe_check`; the standalone verifier has the same unguarded path. This is a regression/security boundary issue for untrusted receipts. Wrap the v0.2 consistency pass in the same safe-check pattern or make it fully defensive.
- [P3] claude (advisory): `docs/specs/OPEN_DECISION_RECEIPT.md:343-346` (and docs-site twin) — §8 still only describes the supporting/dissenting-agent cross-check. This PR adds two new hard-fail checks for v0.2 (`verdicts_consistency`, `dissent_consistency`) and a `quorum_rule` warning, but the spec text isn't updated; since §9.2(3) makes `aragora-verify` normative, third-party implementers can't learn these from the spec. Verified by reading the file.
- [P3] claude (advisory): `aragora-verify/src/aragora_verify/schema.py:75-106` — the stdlib-only path still doesn't type `quorum.method` (gauntlet `_validate_quorum` does, line 301). New code at `verifier.py` `quorum["method"] != "merge-quorum"` tolerates it, so no crash (verified via probe: standalone `ok=True`, gauntlet `schema_conformance` FAIL), but it's a cross-verifier divergence this PR widens by relying on `method`.
- [P3] claude (advisory): `aragora-verify/src/aragora_verify/verifier.py` `_check_expiry` — default policy leaves `ok=True` / exit 0 on an expired-but-valid signature (WARN unless `--strict-expiry`). Documented and tested, just flagging that the safe default for a signature-validity property is arguably the strict one; consider surfacing it in the `authenticity_unverified`-style flag so JSON consumers don't have to grep warnings.

### Output 1

> &gt; ## Claude i&#110;dependent model review
> 
> &gt; Reviewer&#58; claude (anthropic) — independent a&#100;versarial model review via the Aragora C&#108;aude reviewer, grounded on the exact PR head.
> Head: 6f3b3db (6f3b3db75c0c79f6f2e6b5167aa8cea20eb847a5), committed 2026-09-18T09:07:18Z.
> PR: #10107.
> &gt; Model family&#58; claude
> 
> Verdict: PASS
> 
> Verified against the worktree at 6f3b3db: full ODR suites (`aragora-verify/tests`, `tests/gauntlet/test_odr_*`) → 524 passed; all six `docs/specs/examples/*.odr.json` verify under both engines; a 24-case malformed-v0.2 fuzz (findings/verdicts/rule/adjudication/signatures/absent-marker mutants) through both `verify()` and `verify_odr_document()` on the stdlib-only schema path produced schema FAILs, never a crash — so the unguarded `_check_v02_consistency`/`_check_expiry`/`_dissent_trail` reads are indeed covered by validation. Rule logic matches `TierQuorumRule.is_satisfied_by` and `WESTERN_FRONTIER_FAMILIES = {claude, openai}`; dissent-summary logic matches the producer in `aragora/swarm/quorum_receipt.py:74-104` and spec §4.10. Signer's `+00:00` normalization is safe: bundled schema has no `format`/`pattern` on `signed_at`/`expires_at`, and both verifiers parse via `fromisoformat`.
> 
> - (P3) `docs/specs/OPEN_DECISION_RECEIPT.md:343-346` (and docs-site twin) — §8 still only describes the supporting/dissenting-agent cross-check. This PR adds two new hard-fail checks for v0.2 (`verdicts_consistency`, `dissent_consistency`) and a `quorum_rule` warning, but the spec text isn't updated; since §9.2(3) makes `aragora-verify` normative, third-party implementers can't learn these from the spec. Verified by reading the file.
> - (P3) `aragora-verify/src/aragora_verify/schema.py:75-106` — the stdlib-only path still doesn't type `quorum.method` (gauntlet `_validate_quorum` does, line 301). New code at `verifier.py` `quorum["method"] != "merge-quorum"` tolerates it, so no crash (verified via probe: standalone `ok=True`, gauntlet `schema_conformance` FAIL), but it's a cross-verifier divergence this PR widens by relying on `method`.
> - (P3) `aragora-verify/src/aragora_verify/verifier.py` `_check_expiry` — default policy leaves `ok=True` / exit 0 on an expired-but-valid signature (WARN unless `--strict-expiry`). Documented and tested, just flagging that the safe default for a signature-validity property is arguably the strict one; consider surfacing it in the `authenticity_unverified`-style flag so JSON consumers don't have to grep warnings.
> 
> d&#111;gfood: yes

### Output 2

> &gt; ## OpenAI i&#110;dependent model review
> 
> &gt; Reviewer&#58; openai (openai) — independent a&#100;versarial model review via Codex CLI OpenAI harness, grounded on the exact PR head.
> Head: 6f3b3db (6f3b3db75c0c79f6f2e6b5167aa8cea20eb847a5), committed 2026-09-18T09:07:18Z.
> PR: #10107.
> &gt; Model family&#58; openai
> 
> Verdict: CHANGES-REQUESTED
> 
> - (P2) `aragora-verify/src/aragora_verify/verifier.py:_check_v02_consistency` and `aragora/gauntlet/odr_verify.py:_check_v02_consistency`: v0.2 verification can crash on schema-invalid-but-JSON-valid nested quorum data because the new consistency pass directly indexes `p["agent"]`, `v["issuer"]`, `finding["severity"]`, and `finding["blocking"]` before the existing guarded verification checks. The gauntlet verifier explicitly promises malformed receipts become FAIL, but this new pass is outside `_safe_check`; the standalone verifier has the same unguarded path. This is a regression/security boundary issue for untrusted receipts. Wrap the v0.2 consistency pass in the same safe-check pattern or make it fully defensive.
> 
> d&#111;gfood: yes

scarmani:
## Collection summary

Summary: 5 advisory findings. Severity labels only; not a merge decision.
Gate regime: severity-gated dissent ON; P2/P3 findings are advisory to the merge gate.
Merge-gate dissent: none.

### Family outcomes
- claude: pass
- openai: pass

### Adjudication

> none

### Collection action

> post
> prepared exact-head evidence artifact; posting without reviewer regeneration

### Findings

- [P3] claude (advisory): `aragora-verify/src/aragora_verify/verifier.py:250`, `aragora/gauntlet/odr_verify.py:780` — the new per-entry `key_id_mismatch` warning fires for every co-signer whose key isn't the one supplied, so a legitimate multi-signer v0.2 receipt (emitter + reviewer) verified with one key emits tamper-flavoured warnings for each other signer. Warn-only, but the wording ("key_id_mismatch") will read as an anomaly to consumers; consider "not verified with supplied key".
- [P3] claude (advisory): `aragora-verify/src/aragora_verify/verifier.py:583-584` — `quorum_consistency` (and the v0.2 checks) now precede `canonical_digest`/`signature` in `checks[]`, changing the positional order in `--json` output for v0.1 receipts too. In-repo consumers look up by name, so no breakage found; external consumers indexing by position would shift.
- [P3] claude (advisory): `aragora-verify/src/aragora_verify/verifier.py:359`, `aragora/gauntlet/odr_verify.py:908` — western-frontier family set is a hard-coded literal in two verifiers (standalone can't import `aragora`, so duplication is forced); currently in sync with `quorum_evidence.py`, but there's no test pinning that equality, so drift would go unnoticed.
- [P3] claude (advisory): `aragora-verify/src/aragora_verify/verifier.py:378-381` — expiry is only evaluated for verified entries, so `--strict-expiry` without `--pubkey` is a silent no-op (no SKIP/note check emitted). Consistent with the "bind expiry to verified signatures" decision, but a `signature_expiry` SKIP would make the policy visible in output.
- [P3] claude (advisory): `aragora/gauntlet/odr_verify.py:1053-1057` — `verify_odr_document` gains `now`/`strict_expiry` but not `require_issuer`; the docs scope `--require-issuer` to the package CLI, so this is a documented parity gap rather than a defect.

### Output 1

> &gt; ## Claude i&#110;dependent model review
> 
> &gt; Reviewer&#58; claude (anthropic) — independent a&#100;versarial model review via the Aragora C&#108;aude reviewer, grounded on the exact PR head.
> Head: 37c9f01 (37c9f016f317d23167cc755f17648bacfc24210b), committed 2026-09-18T09:39:29Z.
> PR: #10107.
> &gt; Model family&#58; claude
> 
> Verdict: PASS
> 
> Verified by running both suites at 37c9f01 (`tests/gauntlet/test_odr_{verify_v02,export_v02,verify_schema}.py`: 273 passed; `aragora-verify/tests`: 227 passed) and a 5,760-case single-member mutation fuzz of a signed v0.2 receipt (type swaps / deletions at every path, both engines, with and without key+`strict_expiry`+`now`): 0 crashes. Confirmed the unguarded `_check_v02_consistency`/`_check_expiry`/`_dissent_trail` indexing is covered by the bundled-schema extension validator (`verdicts[].issuer`, `rule.*`, `findings[].severity/blocking/text/issuer`, `adjudication.verdict/reason` are typed/required; `signatures` is a required top-level array). Confirmed the producer (`aragora/swarm/quorum_receipt.py:56-104`) computes `blocking`/`severity_max`/`issuer` exactly as the new consistency checks expect, and the `{"claude","openai"}` literal matches `WESTERN_FRONTIER_FAMILIES` in `aragora/swarm/quorum_evidence.py:151`.
> 
> - (P3) `aragora-verify/src/aragora_verify/verifier.py:250`, `aragora/gauntlet/odr_verify.py:780` — the new per-entry `key_id_mismatch` warning fires for every co-signer whose key isn't the one supplied, so a legitimate multi-signer v0.2 receipt (emitter + reviewer) verified with one key emits tamper-flavoured warnings for each other signer. Warn-only, but the wording ("key_id_mismatch") will read as an anomaly to consumers; consider "not verified with supplied key".
> - (P3) `aragora-verify/src/aragora_verify/verifier.py:583-584` — `quorum_consistency` (and the v0.2 checks) now precede `canonical_digest`/`signature` in `checks[]`, changing the positional order in `--json` output for v0.1 receipts too. In-repo consumers look up by name, so no breakage found; external consumers indexing by position would shift.
> - (P3) `aragora-verify/src/aragora_verify/verifier.py:359`, `aragora/gauntlet/odr_verify.py:908` — western-frontier family set is a hard-coded literal in two verifiers (standalone can't import `aragora`, so duplication is forced); currently in sync with `quorum_evidence.py`, but there's no test pinning that equality, so drift would go unnoticed.
> - (P3) `aragora-verify/src/aragora_verify/verifier.py:378-381` — expiry is only evaluated for verified entries, so `--strict-expiry` without `--pubkey` is a silent no-op (no SKIP/note check emitted). Consistent with the "bind expiry to verified signatures" decision, but a `signature_expiry` SKIP would make the policy visible in output.
> - (P3) `aragora/gauntlet/odr_verify.py:1053-1057` — `verify_odr_document` gains `now`/`strict_expiry` but not `require_issuer`; the docs scope `--require-issuer` to the package CLI, so this is a documented parity gap rather than a defect.
> 
> d&#111;gfood: yes

### Output 2

> &gt; ## OpenAI i&#110;dependent model review
> 
> &gt; Reviewer&#58; openai (openai) — independent a&#100;versarial model review via Codex CLI OpenAI harness, grounded on the exact PR head.
> Head: 37c9f01 (37c9f016f317d23167cc755f17648bacfc24210b), committed 2026-09-18T09:39:29Z.
> PR: #10107.
> &gt; Model family&#58; openai
> 
> Verdict: PASS
> 
> No findings.
> 
> d&#111;gfood: yes

scarmani:

Head: 37c9f01 (37c9f016f317d23167cc755f17648bacfc24210b), committed 2026-09-18T09:39:29Z.
PR: #10107.
Model family: claude

Verdict: PASS

Verified by running both suites at 37c9f01 (`tests/gauntlet/test_odr_{verify_v02,export_v02,verify_schema}.py`: 273 passed; `aragora-verify/tests`: 227 passed) and a 5,760-case single-member mutation fuzz of a signed v0.2 receipt (type swaps / deletions at every path, both engines, with and without key+`strict_expiry`+`now`): 0 crashes. Confirmed the unguarded `_check_v02_consistency`/`_check_expiry`/`_dissent_trail` indexing is covered by the bundled-schema extension validator (`verdicts[].issuer`, `rule.*`, `findings[].severity/blocking/text/issuer`, `adjudication.verdict/reason` are typed/required; `signatures` is a required top-level array). Confirmed the producer (`aragora/swarm/quorum_receipt.py:56-104`) computes `blocking`/`severity_max`/`issuer` exactly as the new consistency checks expect, and the `{"claude","openai"}` literal matches `WESTERN_FRONTIER_FAMILIES` in `aragora/swarm/quorum_evidence.py:151`.

- [P3] `aragora-verify/src/aragora_verify/verifier.py:250`, `aragora/gauntlet/odr_verify.py:780` — the new per-entry `key_id_mismatch` warning fires for every co-signer whose key isn't the one supplied, so a legitimate multi-signer v0.2 receipt (emitter + reviewer) verified with one key emits tamper-flavoured warnings for each other signer. Warn-only, but the wording ("key_id_mismatch") will read as an anomaly to consumers; consider "not verified with supplied key".
- [P3] `aragora-verify/src/aragora_verify/verifier.py:583-584` — `quorum_consistency` (and the v0.2 checks) now precede `canonical_digest`/`signature` in `checks[]`, changing the positional order in `--json` output for v0.1 receipts too. In-repo consumers look up by name, so no breakage found; external consumers indexing by position would shift.
- [P3] `aragora-verify/src/aragora_verify/verifier.py:359`, `aragora/gauntlet/odr_verify.py:908` — western-frontier family set is a hard-coded literal in two verifiers (standalone can't import `aragora`, so duplication is forced); currently in sync with `quorum_evidence.py`, but there's no test pinning that equality, so drift would go unnoticed.
- [P3] `aragora-verify/src/aragora_verify/verifier.py:378-381` — expiry is only evaluated for verified entries, so `--strict-expiry` without `--pubkey` is a silent no-op (no SKIP/note check emitted). Consistent with the "bind expiry to verified signatures" decision, but a `signature_expiry` SKIP would make the policy visible in output.
- [P3] `aragora/gauntlet/odr_verify.py:1053-1057` — `verify_odr_document` gains `now`/`strict_expiry` but not `require_issuer`; the docs scope `--require-issuer` to the package CLI, so this is a documented parity gap rather than a defect.

dogfood: yes

scarmani:

Head: 37c9f01 (37c9f016f317d23167cc755f17648bacfc24210b), committed 2026-09-18T09:39:29Z.
PR: #10107.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

PR #10117, opened 2026-09-18
feat(scripts): first-hour receipt script from a clean venv [Tier 2]

scarmani:
## Summary

`scripts/receipt_first_hour.sh` takes an empty machine to a verified Open Decision Receipt, and `tests/scripts/test_receipt_first_hour.py` pins its contract. Nothing under `.github/` — the CI jobs are the separate Tier-4 workflow PR.

The script creates a throwaway virtualenv under `mktemp -d`, installs the published `aragora` and `aragora-verify` (positional `<pkg>==<ver>` pins, or local wheels via `--aragora-wheel`/`--verify-wheel` for pre-publish runs), then runs `aragora demo rate-limiter --offline --receipt r.json`, `aragora receipt export r.json --format odr --output r.odr.json` and `aragora-verify r.odr.json`.

Contract details it holds to (architecture §2.7):

- Transcript: one `step: <install|demo|export|verify>` line as each step starts, `venv: <abs path>` and `odr: <abs path>.odr.json` once they exist, and a final `total wall time: <n>s`.
- Exit-code discipline: only verifier exit 0 is success. A failing step prints the tool's code as `exit=<n>` and propagates it; no `|| true` and no `-eq 3` anywhere near the `aragora-verify` line. A successful run over 300 s exits 1.
- Env hygiene: the demo and export steps run with `ARAGORA_ODR_SIGNING_KEY_FILE`, `ARAGORA_ODR_SIGNING_KEY_SECRET`, `ARAGORA_ENV` and `ARAGORA_ENVIRONMENT` removed (`env -u`) and `ARAGORA_USE_SECRETS_MANAGER=false` set outright, so the first-hour document is unsigned whatever the caller exported and whichever machine it runs on; the install step drops `PIP_INDEX_URL`, `PIP_EXTRA_INDEX_URL`, `PIP_FIND_LINKS` and `PIP_CONSTRAINT` so the packages a stranger verifies really come from PyPI; `PYTHONPATH` is `unset` so a checkout cannot shadow the installed packages.
- RULING 4 / invariant §5.2b: the script never passes `--odr-version` and never sets `ARAGORA_ODR_PROFILE_VERSION`. It exercises the DEFAULT path of the published pair, so it stays green on 2.9.0 + 0.1.1 (v0.1 document) today and after 2.11.0 + 0.2.0 (v0.2 document).
- `--help`/`-h` exit 0 and list `--published-receipt` and `--pubkey-url`; the explicit `--published-receipt <url> --pubkey-url <url>` path is wired now (each flag requires the other), and the M4 self-resolve step lands later.

Round-1 review repair (`f6a352a`): relative wheel paths are resolved against the caller's cwd before the script moves into its scratch directory (and a missing wheel is a usage error), the unsigned-export guard was widened to the secrets-manager switches above, and `--help` now prints the header up to the first line of code instead of a hardcoded line range that already truncated one line.

Round-2 review repair (`9009707`): two holes in that same env-hygiene construct. `env -u ARAGORA_USE_SECRETS_MANAGER` did not close the AWS signing path, because with the flag absent `aragora/config/secrets.py:330-338` still opts in when `AWS_EXECUTION_ENV`/`AWS_LAMBDA_FUNCTION_NAME` is set or the environment name is production/staging; the script now sets the flag to `false` explicitly. And a caller-exported `PIP_INDEX_URL`/`PIP_EXTRA_INDEX_URL`/`PIP_FIND_LINKS`/`PIP_CONSTRAINT` could silently redirect the install to a mirror, which falsifies the one guarantee this script exists to give — that the artifacts under test are the *published* ones. The existing hygiene test was extended to cover both (it now runs with an AWS marker and a bogus index set, and asserts what the toolchain actually received).

## Exit metric moved

Row 8 prerequisite (first-hour path): a stranger can now go from nothing to a verified receipt with one command, in 12-15 s measured.

## Test commands

```
# unit suite (fake toolchain, no network)
python3 -m pytest tests/scripts/test_receipt_first_hour.py -q         -> 12 passed in 4.79s

# VAL-ODR-022.1 default run against published PyPI (aragora 2.9.0 / aragora-verify 0.1.1)
bash scripts/receipt_first_hour.sh; echo exit=$?                      -> exit=0
  step: install / step: demo / step: export / step: verify (in order)
  venv: /var/folders/.../T/receipt-first-hour.AruWpt/venv
  odr:  /var/folders/.../T/receipt-first-hour.AruWpt/work/r.odr.json
  => VERIFIED (ODR v0.1)
  total wall time: 13s

# VAL-ODR-022.1 help
bash scripts/receipt_first_hour.sh --help; echo exit=$?               -> exit=0, lists --published-receipt and --pubkey-url
bash scripts/receipt_first_hour.sh -h | grep -c -E '--published-receipt|--pubkey-url'  -> 2

# VAL-ODR-022.2 pins
head -30 scripts/receipt_first_hour.sh | grep -c 'aragora=='          -> 1
bash scripts/receipt_first_hour.sh aragora-verify==99.9.9; echo exit=$?   -> exit=1
  ERROR: No matching distribution found for aragora-verify==99.9.9
  receipt-first-hour: install step failed, exit=1 / total wall time: 2s

# VAL-ODR-022.1/.3/.4 static contract
grep -c -E 'odr-version|ODR_PROFILE_VERSION' scripts/receipt_first_hour.sh  -> 0
grep -c -E 'python3? -m venv' scripts/receipt_first_hour.sh                 -> 1 (under the mktemp -d root)
grep -nE 'PYTHONPATH' scripts/receipt_first_hour.sh | grep -v -E 'unset PYTHONPATH|env -u PYTHONPATH' | wc -l  -> 0
grep -n 'aragora-verify' scripts/receipt_first_hour.sh | grep -c -E '\|\| true|-eq 3'  -> 0

# VAL-ODR-029.1 signing-env hygiene, hostile caller environment (round-2 repair, both holes at once)
ARAGORA_ODR_SIGNING_KEY_FILE="$RF_ODR_SIGNING_KEY_FILE" ARAGORA_ODR_SIGNING_KEY_SECRET=mission-secret \
  ARAGORA_USE_SECRETS_MANAGER=true ARAGORA_ENV=production AWS_EXECUTION_ENV=AWS_ECS_FARGATE \
  PIP_INDEX_URL=https://pypi.invalid/simple PIP_EXTRA_INDEX_URL=https://extra.invalid/simple \
  bash scripts/receipt_first_hour.sh; echo exit=$?    -> exit=0, => VERIFIED, total wall time: 13s
python3 -c "import json,sys;print(json.load(open(sys.argv[1]))['signatures'])" <odr path>  -> []
  (install succeeded against real PyPI with both bogus indexes exported, and the AWS marker plus
   ARAGORA_USE_SECRETS_MANAGER=true did not produce a signature: both repairs exercised in one run)

# VAL-ODR-029.3 tamper on the first-hour document (corrected sub-case: exit 0 is the *contract*)
<venv>/bin/aragora-verify <tampered copy>; echo exit=$?                -> exit=0, => VERIFIED
python3 -c "import json;d=json.load(open(p));print(sorted(d));print('odr_digest' in d, d['signatures'])"
  ['attestation','claim','confidence','cruxes','issued_at','odr_version','profile','quorum',
   'reasoning','receipt_id','routing','signatures','source','subject']
  False []
  (no top-level odr_digest and no signatures: nothing in the unsigned v0.1 document binds its content,
   so exit 0 is correct. Tamper-evidence is proven on the signed v0.2 path by VAL-CROSS-001.5 below.)

# round-1 repair, with real pip and relative wheels, run from /tmp
bash <wt>/scripts/receipt_first_hour.sh --aragora-wheel rf-dist-cross/aragora-2.10.0-py3-none-any.whl \
  --verify-wheel rf-dist-cross/aragora_verify-0.2.0-py3-none-any.whl       -> exit=0, => VERIFIED, total wall time: 14s
bash <wt>/scripts/receipt_first_hour.sh --aragora-wheel rf-dist-cross/nope.whl  -> exit=2
  receipt-first-hour: no such wheel: rf-dist-cross/nope.whl

# VAL-CROSS-001 clean venv, both LOCAL wheels
pip install -q /tmp/rf-dist-cross/aragora-2.10.0-py3-none-any.whl /tmp/rf-dist-cross/aragora_verify-0.2.0-py3-none-any.whl
  1. unsigned export + verify            -> exit=0, `0.2 []`
  2. signed export (mission key inline)  -> exit=0, `1 ed25519-`
  3. verify --pubkey                     -> exit=0, VERIFIED, Dissent trail, => VERIFIED (key_id=ed25519-44c316618e9a0f58)
  4. verify without pubkey               -> exit=3, UNVERIFIED
  5. tampered receipt_id + --pubkey      -> exit=1, [FAIL] signature: ... INVALID
  6. elapsed=11s; then the script itself -> exit=0, total wall time: 12s

# repository gates (mission bin/gate.sh, from this worktree, on the final head)
lint / typecheck / contracts / test / guardrails  -> GATE PASSED x5 (test: 4578 passed, 8 skipped)
bash scripts/automation_pr_preflight.sh origin/main HEAD  -> preflight: ok (exit 0)
```

## Reviewed design tradeoffs

- **Pins as positional arguments, not flags.** `bash scripts/receipt_first_hour.sh aragora==2.9.0 aragora-verify==0.1.1` reads like the `pip install` line it becomes, keeps the CI job one token away from a replay, and leaves `--*-wheel` free for the pre-publish path. An unrecognised argument is a usage error (exit 2) rather than a silently ignored token.
- **`env -u` per invocation instead of a global `unset`.** The signing inputs are stripped on the exact commands that write the document, so the hygiene is visible on the line it protects and cannot drift if a later step needs a key (the published-receipt path deliberately keeps the caller's environment).
- **The venv and the document are left on disk.** A trap that deleted the temp root would make the transcript's `venv:`/`odr:` paths useless for the tamper and signature proofs that follow the run, which is precisely what the validation contract inspects. The tests point `TMPDIR` at pytest's `tmp_path` so the suite does not accumulate scratch roots.
- **The 300 s budget is enforced only on the success path.** A failing step already exits non-zero with its own code; re-mapping that to a budget failure would hide the real cause. The budget answers "was this still a first *hour* path", not "did it work".
- **No profile version anywhere.** Passing `--odr-version` would make the script test a pinned profile rather than the promise that the published emitter and the published verifier agree on their defaults, which is the invariant this script exists to keep executable.
- **Round-1 suggestion not taken: `python3 -m venv --upgrade-deps`.** It would make every run fetch a pip upgrade from PyPI before the install, and it turns venv creation into a network step that fails outright on an offline or mirror-restricted machine, which is exactly the pre-publish local-wheel case the flags exist for. An incompatible bundled pip already surfaces as a non-zero install step carrying pip's own message, which is the behaviour the exit-code contract asks for.

## Risks

- The run depends on PyPI being reachable; a resolution failure surfaces as a non-zero install step with the pip error, not a silent pass. The local-wheel flags cover the offline and pre-publish cases.
- The published pair emits an **unsigned v0.1** document with no top-level `odr_digest` and `signatures == []`, so a tampered copy of that document still verifies (`[WARN] signature: receipt is unsigned (v0.1)`) and correctly exits 0 — nothing in it binds the content. Both `aragora-verify` 0.1.1 and 0.2.0 behave this way. Tamper-evidence is proven on the v0.2 path instead (VAL-CROSS-001.5 above: exit 1, `[FAIL] signature`), and the published path becomes tamper-evident once `aragora` ≥2.11.0 ships the v0.2 default, which is the transition VAL-ODR-029.1 already anticipates.
- `mktemp -d` roots accumulate under `$TMPDIR` across runs by design; they are inside the OS temp sweep.

## Round-2 finding sort

Both accepted findings are repaired on this head (`9009707`, described above). The rest are recorded as advisory and deliberately not repaired here:

- **P3 — missing option value exits 1, not the script's exit-2 usage convention.** `--aragora-wheel` with no argument hits `${2:?...}`, which makes bash exit 1 with its own message instead of routing through the usage error path that every other malformed invocation uses. Real but cosmetic: the run still fails loudly with a message naming the option, and no contract assertion reads the code for this case. Not worth churning the head on this PR's last review pass.
- **P3 — the nightly Windows shard collects `tests/scripts/`.** A `.github/` concern, and this PR is required to stay free of `.github/` edits (the CI jobs are the separate Tier-4 workflow PR, `m1-receipt-first-hour-ci-jobs`, where it belongs). The affected job is gated `if: github.event_name != 'pull_request'`, so it cannot block this merge, and adding a `skipif` guard here would change the repository's skip-marker count, which the skip-marker audit gate pins.

## Round-3 finding sort (collection 3 of 3, not clean)

Both families returned CHANGES-REQUESTED on one shared P2: scrubbing `PIP_INDEX_URL`/`PIP_EXTRA_INDEX_URL`/`PIP_FIND_LINKS`/`PIP_CONSTRAINT` leaves pip's *config-file* channel open (`PIP_CONFIG_FILE`, `~/.config/pip/pip.conf`, `/etc/pip.conf`, `$VIRTUAL_ENV/pip.conf`), so a machine- or user-level `index-url` can still substitute the packages while the transcript claims the published ones. Reproduced on this head: a poisoned `PIP_CONFIG_FILE` makes the run exit 1 at the install step, and a valid mirror would have substituted silently. **Repaired on this head** (`9a62437`): `PIP_CONFIG_FILE=/dev/null` joins the same `INSTALL_ENV` array and the existing hygiene test is extended again — no new test and no other code change. Worth recording: `pip --isolated` alone does not close this, because `PIP_CONFIG_FILE` outranks isolation, while `PIP_CONFIG_FILE=/dev/null` disables user, site and environment-level config files together.

The four round-3 P3s stay advisory and unrepaired here: the 300 s budget is only evaluated after the run ends, so a hung step is unbounded; a host without `ensurepip` fails with `install step failed, exit=127` and no hint; the `--published-receipt` path emits no `step:` line, labels both curls `download` and follows redirects on the public key; and the suite could not be executed in the reviewer's sandbox (it verified the demo→export→verify chain and the AWS rationale by reading the repo instead).

## Attempt 2: the pip config-file channel, closed and proved

`pip --isolated` is not the lever — `PIP_CONFIG_FILE` outranks it — so the install step pins
`PIP_CONFIG_FILE=/dev/null`, which disables the user, site and environment-level config files
together. Measured on this head:

```
# the channel, and that isolation does not close it
PIP_CONFIG_FILE=<poisoned pip.conf> python3 -m pip --isolated config list
  global.index-url='https://worker-poison.invalid/simple'
PIP_CONFIG_FILE=/dev/null python3 -m pip config list
  :env:.config-file='/dev/null'          (no global.index-url)

# the script, with each half of the channel poisoned
PIP_CONFIG_FILE=<poisoned pip.conf> bash scripts/receipt_first_hour.sh   -> exit=0, total wall time: 13s
HOME=<fake home with .config/pip/pip.conf> bash scripts/receipt_first_hour.sh -> exit=0, total wall time: 17s
pytest tests/scripts/test_receipt_first_hour.py                          -> 12 passed
  (the extended hygiene test asserts the fake pip observes CONFIG_FILE=/dev/null alongside the
   four <unset> index variables)

# contract proofs re-run at this head
VAL-ODR-022.1  bash scripts/receipt_first_hour.sh        -> exit=0; four step: lines in order,
               venv: and odr: under a mktemp -d root, total wall time: 14s
VAL-ODR-022.1  --help / -h                               -> exit=0, byte-identical, both flags listed
VAL-ODR-022.2  aragora-verify==99.9.9                    -> exit=1, "install step failed, exit=1"
VAL-ODR-029.1  ARAGORA_ODR_SIGNING_KEY_FILE=<key> run    -> exit=0, signatures []
VAL-ODR-029.3  tampered copy                             -> exit=0; keys carry no odr_digest,
               'odr_digest' in d / d['signatures'] -> False []
VAL-CROSS-001  steps 1-6 in full, clean venv, local wheels aragora-2.10.0 + aragora_verify-0.2.0
  1. unsigned v0.2 export + verify   -> exit=0, `0.2 []`
  2. signed export (key inline)      -> exit=0, `1 ed25519-` (ed25519-44c316618e9a0f58)
  3. verify --pubkey                 -> exit=0, VERIFIED, Dissent trail, => VERIFIED (key_id=ed25519-44c316618e9a0f58)
  4. verify without --pubkey         -> exit=3, UNVERIFIED
  5. tampered receipt_id             -> exit=1, [FAIL] signature: ... INVALID
  6. elapsed=12s; the script itself  -> exit=0, total wall time: 13s

lint / typecheck / contracts / guardrails / test  -> GATE PASSED x5
  (test: 4578 passed, 8 skipped; skip markers 95; publish-shadow 230 passed, 29 skipped)
bash scripts/automation_pr_preflight.sh origin/main HEAD  -> preflight: ok (exit 0)
```

Diff size at this head: 538 added, 0 removed across the two files.

## Round-4 finding sort (attempt 2, collection 1 of 3)

Claude returned a grounded PASS on `9a62437`. OpenAI returned CHANGES-REQUESTED with two [P2]s,
sorted below; one is refuted by execution, the other is repaired as text on `34819e1`.

- **[P2] `${EXTRA_SPECS[@]+"${EXTRA_SPECS[@]}"}` "is not portable and can expand badly when
  empty" — REFUTED, and the proposed replacement is the bug.** The finding recommends
  `"${EXTRA_SPECS[@]}"` "directly; empty arrays expand to nothing safely in Bash". That is true
  from Bash 4.4 onward and false on the Bash that ships with macOS, which is what a stranger
  running this script on a Mac gets. Measured on both:

  ```
  /bin/bash 3.2.57:  set -u; a=(); printf "%s " "${a[@]}"        -> a[@]: unbound variable
  /bin/bash 3.2.57:  set -u; a=(); printf "%s " ${a[@]+"${a[@]}"} -> before  after   (ok)
  bash 5.3.9:        both forms ok
  /bin/bash 3.2.57 and 5.3.9: a=("x y" "z"); printf "[%s]" ${a[@]+"${a[@]}"} -> [x y][z]
  ```

  The current idiom is the portable one, it preserves quoting on non-empty arrays, and the
  default no-extra-spec run is exercised green in every proof above. Unchanged.
- **[P2] pip resolves transitive dependencies from the live index, so the transcript is not a
  byte-exact replay.** The observation is correct and the wording it lands on was mine: the
  header said a pinned run "can be replayed exactly", which claims more than the install
  delivers. Repaired as text on `34819e1` — the header now says a run names the exact `aragora`
  and `aragora-verify` under test and that their dependencies resolve at install time. The
  suggested remedy (a constraints/lock input, or `--no-deps`) is deliberately not taken here:
  `PIP_CONSTRAINT` is one of the variables this PR scrubs precisely so a caller cannot decide
  what gets installed, and `--no-deps` would leave `aragora` unable to run the demo the script
  verifies. A fully pinned artifact set is a different guarantee from "the published packages,
  as a stranger would get them", which is the one this script exists to keep executable.


scarmani:
## Collection summary

Summary: 6 advisory findings. Severity labels only; not a merge decision.
Gate regime: severity-gated dissent ON; P2/P3 findings are advisory to the merge gate.
Merge-gate dissent: none.

### Family outcomes
- claude: changes_requested
- openai: pass

### Adjudication

> none

### Collection action

> prepare
> dry-run; re-run with --apply to post

### Findings

- [P2] claude (advisory): `scripts/receipt_first_hour.sh` — `cd "$WORK"` (temp dir) runs *before* `pip install "$ARAGORA_SPEC" "$VERIFY_SPEC"`, so `--aragora-wheel` / `--verify-wheel` given as relative paths (the natural `dist/aragora-2.11.0-py3-none-any.whl` form) resolve against the scratch dir and pip fails with a "not a valid file" error. `test_local_wheels_replace_the_pypi_specs` only exercises absolute `tmp_path` wheels, so it can't catch this. Resolve to an absolute path at parse time (`realpath`/`$(cd "$(dirname ...)" &amp;&amp; pwd)/$(basename ...)`) or run pip from the original cwd.
- [P3] claude (advisory): `scripts/receipt_first_hour.sh` demo/export lines — unsetting only the two `ARAGORA_ODR_SIGNING_KEY_*` vars doesn't stop signing. Verified by reading `aragora/gauntlet/odr_export.py:316-339` → `odr_signing.load_signing_key_from_secrets` → `_load_pem_secret_from_aws` and `aragora/config/secrets.py:330-337`: with `ARAGORA_USE_SECRETS_MANAGER=true`, `ARAGORA_ENV=production|staging`, or `AWS_EXECUTION_ENV` in the caller's env, export consults the default AWS secret. If a key is found the receipt is signed and `aragora-verify` with no `--pubkey` returns 3 (`aragora-verify/src/aragora_verify/cli.py:102`), failing the run; if AWS is unreadable, export fails. Add `-u ARAGORA_USE_SECRETS_MANAGER -u ARAGORA_ENV -u ARAGORA_ENVIRONMENT` (or set `ARAGORA_USE_SECRETS_MANAGER=false`) to make the "unsigned receipt" assumption hold.
- [P3] claude (advisory): `scripts/receipt_first_hour.sh` `usage()` — `sed -n '2,23p'` truncates the header: line 24 ("The virtualenv and the exported document are left on disk…") is dropped, and the hardcoded range breaks silently on any header edit. Print up to the first non-`#` line instead.
- [P3] claude (advisory): `tests/scripts/test_receipt_first_hour.py` `_run` — the script is intentionally non-cleaning, so every test invocation (8+ per session) leaves a `receipt-first-hour.*` dir in the real `$TMPDIR`. Pass `TMPDIR=str(tmp_path)` in the env so pytest reclaims them.
- [P3] claude (advisory): `tests/scripts/test_receipt_first_hour.py` `FAKE_PYTHON` — the fallthrough `exec /usr/bin/env python3 "$@"` re-resolves to the fake itself (its dir is first on PATH), i.e. infinite recursion if the script ever calls `python3` with anything but `-m venv`. Latent today; substitute `sys.executable` when writing the shim.
- [P3] claude (advisory): (unverified) `scripts/receipt_first_hour.sh` install step — the venv's bundled pip is never upgraded. `pyproject.toml` builds with `setuptools&gt;=83` and PEP 639 `license = "MIT"` (Metadata-Version 2.4); stock pip on e.g. Ubuntu 22.04 (22.0.2) may reject the wheel, and `requires-python &gt;=3.10` will fail opaquely on a 3.9 system. `python3 -m venv --upgrade-deps` (3.9+) is cheap insurance. Verifying would mean building the wheel and installing it into a fresh py3.10 venv with stock pip.

### Output 1

> &gt; ## Claude i&#110;dependent model review
> 
> &gt; Reviewer&#58; claude (anthropic) — independent a&#100;versarial model review via the Aragora C&#108;aude reviewer, grounded on the exact PR head.
> Head: eeb6179 (eeb617954c789b781e83f2a947e2e0efffff7942), committed 2026-09-18T23:05:49Z.
> PR: #10117.
> &gt; Model family&#58; claude
> 
> Verdict: CHANGES-REQUESTED
> 
> - **(P2)** `scripts/receipt_first_hour.sh` — `cd "$WORK"` (temp dir) runs *before* `pip install "$ARAGORA_SPEC" "$VERIFY_SPEC"`, so `--aragora-wheel` / `--verify-wheel` given as relative paths (the natural `dist/aragora-2.11.0-py3-none-any.whl` form) resolve against the scratch dir and pip fails with a "not a valid file" error. `test_local_wheels_replace_the_pypi_specs` only exercises absolute `tmp_path` wheels, so it can't catch this. Resolve to an absolute path at parse time (`realpath`/`$(cd "$(dirname ...)" &amp;&amp; pwd)/$(basename ...)`) or run pip from the original cwd.
> - **(P3)** `scripts/receipt_first_hour.sh` demo/export lines — unsetting only the two `ARAGORA_ODR_SIGNING_KEY_*` vars doesn't stop signing. Verified by reading `aragora/gauntlet/odr_export.py:316-339` → `odr_signing.load_signing_key_from_secrets` → `_load_pem_secret_from_aws` and `aragora/config/secrets.py:330-337`: with `ARAGORA_USE_SECRETS_MANAGER=true`, `ARAGORA_ENV=production|staging`, or `AWS_EXECUTION_ENV` in the caller's env, export consults the default AWS secret. If a key is found the receipt is signed and `aragora-verify` with no `--pubkey` returns 3 (`aragora-verify/src/aragora_verify/cli.py:102`), failing the run; if AWS is unreadable, export fails. Add `-u ARAGORA_USE_SECRETS_MANAGER -u ARAGORA_ENV -u ARAGORA_ENVIRONMENT` (or set `ARAGORA_USE_SECRETS_MANAGER=false`) to make the "unsigned receipt" assumption hold.
> - **(P3)** `scripts/receipt_first_hour.sh` `usage()` — `sed -n '2,23p'` truncates the header: line 24 ("The virtualenv and the exported document are left on disk…") is dropped, and the hardcoded range breaks silently on any header edit. Print up to the first non-`#` line instead.
> - **(P3)** `tests/scripts/test_receipt_first_hour.py` `_run` — the script is intentionally non-cleaning, so every test invocation (8+ per session) leaves a `receipt-first-hour.*` dir in the real `$TMPDIR`. Pass `TMPDIR=str(tmp_path)` in the env so pytest reclaims them.
> - **(P3)** `tests/scripts/test_receipt_first_hour.py` `FAKE_PYTHON` — the fallthrough `exec /usr/bin/env python3 "$@"` re-resolves to the fake itself (its dir is first on PATH), i.e. infinite recursion if the script ever calls `python3` with anything but `-m venv`. Latent today; substitute `sys.executable` when writing the shim.
> - **(P3)** (unverified) `scripts/receipt_first_hour.sh` install step — the venv's bundled pip is never upgraded. `pyproject.toml` builds with `setuptools&gt;=83` and PEP 639 `license = "MIT"` (Metadata-Version 2.4); stock pip on e.g. Ubuntu 22.04 (22.0.2) may reject the wheel, and `requires-python &gt;=3.10` will fail opaquely on a 3.9 system. `python3 -m venv --upgrade-deps` (3.9+) is cheap insurance. Verifying would mean building the wheel and installing it into a fresh py3.10 venv with stock pip.
> 
> d&#111;gfood: yes

### Output 2

> &gt; ## OpenAI i&#110;dependent model review
> 
> &gt; Reviewer&#58; openai (openai) — independent a&#100;versarial model review via Codex CLI OpenAI harness, grounded on the exact PR head.
> Head: eeb6179 (eeb617954c789b781e83f2a947e2e0efffff7942), committed 2026-09-18T23:05:49Z.
> PR: #10117.
> &gt; Model family&#58; openai
> 
> Verdict: PASS
> 
> No findings.
> 
> d&#111;gfood: yes

scarmani:
## Collection summary

Summary: 4 advisory findings. Severity labels only; not a merge decision.
Gate regime: severity-gated dissent ON; P2/P3 findings are advisory to the merge gate.
Merge-gate dissent: none.

### Family outcomes
- claude: changes_requested
- openai: pass

### Adjudication

> none

### Collection action

> prepare
> dry-run; re-run with --apply to post

### Findings

- [P2] claude (advisory): `scripts/receipt_first_hour.sh:129-130` — `UNSIGNED_ENV` unsets `ARAGORA_USE_SECRETS_MANAGER`/`ARAGORA_ENV`/`ARAGORA_ENVIRONMENT`, but that does not close the secrets-manager channel the comment claims to close. Verified by reading `aragora/config/secrets.py:330-337`: when the flag is *unset*, `use_aws` is also true if `AWS_EXECUTION_ENV` or `AWS_LAMBDA_FUNCTION_NAME` is set (ECS/CodeBuild/Lambda-hosted runners). `aragora/gauntlet/odr_export.py:333` → `odr_signing.load_signing_key_from_secrets` → `_load_pem_secret_from_aws` (`odr_signing.py:171-243`): with `use_aws` true and no boto3, `_get_aws_client` returns `None` → `all_not_found=False` → `OdrSigningError` → the export step fails; with creds and a provisioned default secret, the receipt is silently signed and the unkeyed `aragora-verify` call still passes (signatures "NOT checked"). Fix: set `ARAGORA_USE_SECRETS_MANAGER=false` explicitly (an explicit flag short-circuits the AWS heuristics at `secrets.py:331`) instead of `-u`, and update `test_signing_key_env_is_unset_for_the_demo_and_export_steps` to assert `SECRETS_MANAGER=false` rather than `&lt;unset&gt;` (the current test locks in the weaker behavior).
- [P3] claude (advisory): `scripts/receipt_first_hour.sh:74-88` — `${2:?...}` on a missing option value exits 1 with bash's raw message, unlike every other usage error which exits 2 and prints usage.
- [P3] claude (advisory): `scripts/receipt_first_hour.sh:113` — only `PYTHONPATH` is scrubbed; `PIP_INDEX_URL`/`PIP_EXTRA_INDEX_URL`/`PIP_FIND_LINKS`/`PIP_CONSTRAINT` from the caller's environment can silently replace the "PUBLISHED" package with a mirror/local build, undermining the script's stated guarantee. Consider `-u` on those for the pip call.
- [P3] claude (advisory): `tests/scripts/test_receipt_first_hour.py` — unverified: shard 4 of `.github/workflows/test.yml` (`:1510`, `:1527`, `:1591`) runs `tests/` on `windows-latest` with no ignore for `tests/scripts`; this test spawns `bash`, relies on `#!/usr/bin/env bash` fakes found via PATH, and asserts POSIX paths (`venv: /`, `Path(...).is_dir()`), which is unlikely to hold under Git Bash/MSYS path translation. Sibling tests guard with `pytest.mark.skipif(sys.platform == "win32", ...)` (`tests/scripts/test_check_sdk_parity.py:626`); a Windows-shard run (or adding the same skipif) would settle it.

### Output 1

> &gt; ## Claude i&#110;dependent model review
> 
> &gt; Reviewer&#58; claude (anthropic) — independent a&#100;versarial model review via the Aragora C&#108;aude reviewer, grounded on the exact PR head.
> Head: f6a352a (f6a352a986815e4ab88f350cbbc52d322310c53a), committed 2026-09-18T23:17:27Z.
> PR: #10117.
> &gt; Model family&#58; claude
> 
> Verdict: CHANGES-REQUESTED
> 
> - (P2) `scripts/receipt_first_hour.sh:129-130` — `UNSIGNED_ENV` unsets `ARAGORA_USE_SECRETS_MANAGER`/`ARAGORA_ENV`/`ARAGORA_ENVIRONMENT`, but that does not close the secrets-manager channel the comment claims to close. Verified by reading `aragora/config/secrets.py:330-337`: when the flag is *unset*, `use_aws` is also true if `AWS_EXECUTION_ENV` or `AWS_LAMBDA_FUNCTION_NAME` is set (ECS/CodeBuild/Lambda-hosted runners). `aragora/gauntlet/odr_export.py:333` → `odr_signing.load_signing_key_from_secrets` → `_load_pem_secret_from_aws` (`odr_signing.py:171-243`): with `use_aws` true and no boto3, `_get_aws_client` returns `None` → `all_not_found=False` → `OdrSigningError` → the export step fails; with creds and a provisioned default secret, the receipt is silently signed and the unkeyed `aragora-verify` call still passes (signatures "NOT checked"). Fix: set `ARAGORA_USE_SECRETS_MANAGER=false` explicitly (an explicit flag short-circuits the AWS heuristics at `secrets.py:331`) instead of `-u`, and update `test_signing_key_env_is_unset_for_the_demo_and_export_steps` to assert `SECRETS_MANAGER=false` rather than `&lt;unset&gt;` (the current test locks in the weaker behavior).
> - (P3) `scripts/receipt_first_hour.sh:74-88` — `${2:?...}` on a missing option value exits 1 with bash's raw message, unlike every other usage error which exits 2 and prints usage.
> - (P3) `scripts/receipt_first_hour.sh:113` — only `PYTHONPATH` is scrubbed; `PIP_INDEX_URL`/`PIP_EXTRA_INDEX_URL`/`PIP_FIND_LINKS`/`PIP_CONSTRAINT` from the caller's environment can silently replace the "PUBLISHED" package with a mirror/local build, undermining the script's stated guarantee. Consider `-u` on those for the pip call.
> - (P3) `tests/scripts/test_receipt_first_hour.py` — unverified: shard 4 of `.github/workflows/test.yml` (`:1510`, `:1527`, `:1591`) runs `tests/` on `windows-latest` with no ignore for `tests/scripts`; this test spawns `bash`, relies on `#!/usr/bin/env bash` fakes found via PATH, and asserts POSIX paths (`venv: /`, `Path(...).is_dir()`), which is unlikely to hold under Git Bash/MSYS path translation. Sibling tests guard with `pytest.mark.skipif(sys.platform == "win32", ...)` (`tests/scripts/test_check_sdk_parity.py:626`); a Windows-shard run (or adding the same skipif) would settle it.
> 
> d&#111;gfood: yes

### Output 2

> &gt; ## OpenAI i&#110;dependent model review
> 
> &gt; Reviewer&#58; openai (openai) — independent a&#100;versarial model review via Codex CLI OpenAI harness, grounded on the exact PR head.
> Head: f6a352a (f6a352a986815e4ab88f350cbbc52d322310c53a), committed 2026-09-18T23:17:27Z.
> PR: #10117.
> &gt; Model family&#58; openai
> 
> Verdict: PASS
> 
> No findings.
> 
> d&#111;gfood: yes

scarmani:
## Collection summary

Summary: 6 advisory findings. Severity labels only; not a merge decision.
Gate regime: severity-gated dissent OFF; a changes-requested verdict blocks the merge gate whatever its findings.
Merge-gate dissent: claude, openai.

### Family outcomes
- claude: changes_requested
- openai: changes_requested

### Adjudication

> none

### Collection action

> prepare
> dry-run; re-run with --apply to post

### Findings

- [P2] claude (advisory): `scripts/receipt_first_hour.sh:127-129` — The "caller's index must not decide what gets installed" guarantee is only enforced against four env vars via `env -u`. pip also reads `PIP_CONFIG_FILE`, `~/.config/pip/pip.conf` (`$HOME/.pip/pip.conf`), `/etc/pip.conf`, and `$VIRTUAL_ENV/pip.conf`; an `index-url =`/`find-links =` there silently substitutes packages and the transcript still says "published". Verified by reading the diff (nothing scrubs config files) against pip's documented config precedence. Fix: `pip --isolated install ...` (ignores env vars and user config), or additionally `PIP_CONFIG_FILE=/dev/null`; then extend the fake-pip test to assert the flag/env. The comment in the test docstring makes the same over-claim.
- [P2] openai (advisory): `scripts/receipt_first_hour.sh:111` The install path still allows caller pip configuration to choose the package source. The script unsets `PIP_INDEX_URL`, `PIP_EXTRA_INDEX_URL`, `PIP_FIND_LINKS`, and `PIP_CONSTRAINT`, but pip also reads config files (`pip.conf` / `pip.ini`, `PIP_CONFIG_FILE`). That means a machine-level or user-level `index-url`/`find-links` can still install non-PyPI packages while the transcript claims evidence about the published packages. Add `PIP_CONFIG_FILE=/dev/null` or equivalent isolation, and ideally assert the resolved install source.
- [P3] claude (advisory): `scripts/receipt_first_hour.sh:54-61` — The 300 s budget is checked post hoc only; a hung `pip install` or demo runs indefinitely. Header says "a run over 300 s exits non-zero", which is true only if the run ends. Consider `timeout`-wrapping the pip/demo steps or documenting the limit as advisory.
- [P3] claude (advisory): `scripts/receipt_first_hour.sh:121-127` — On Debian/Ubuntu hosts without `python3-venv`/ensurepip, `python3 -m venv` succeeds but `$VENV/bin/pip` is absent; the run reports `install step failed, exit=127` with no hint. A `[ -x "$VENV/bin/pip" ]` check with a clear message would help the "empty machine" audience. Unverified on a live Debian box; reproducible with `python3 -m venv --without-pip`.
- [P3] claude (advisory): `scripts/receipt_first_hour.sh:156-160` — The published-receipt path emits no `step:` line and labels both curls `download`, so a failure doesn't say which URL failed; also `curl -L` follows redirects for the *public key*, so a redirect at the pubkey URL changes what's trusted without appearing in the transcript. Print the resolved URL (`-w '%{url_effective}'`) or drop `-L` for the key.
- [P3] claude (advisory): `tests/scripts/test_receipt_first_hour.py` — I could not execute the suite (sandbox denied the fetch/run); findings above are from reading. Verified by reading the repo instead: `aragora demo --offline --receipt` (`aragora/cli/parser.py:1435-1443`), `receipt export --format odr --output` (`aragora/cli/commands/receipt.py:171-179`), `aragora-verify --pubkey` with exit 0/1/2/3 semantics (`aragora-verify/src/aragora_verify/cli.py:65-103`), and that `ARAGORA_USE_SECRETS_MANAGER=false` with both key vars unset yields an unsigned export rather than an error (`odr_export.py:320-339`, `secrets.py:330-338`) — so the demo→export→verify chain and the AWS_EXECUTION_ENV rationale in the script are correct.

### Output 1

> &gt; ## Claude i&#110;dependent model review
> 
> &gt; Reviewer&#58; claude (anthropic) — independent a&#100;versarial model review via the Aragora C&#108;aude reviewer, grounded on the exact PR head.
> Head: 9009707 (9009707b2e7645de26837329eb804b52437ccd07), committed 2026-09-18T23:44:45Z.
> PR: #10117.
> &gt; Model family&#58; claude
> 
> Verdict: CHANGES-REQUESTED
> 
> - (P2) `scripts/receipt_first_hour.sh:127-129` — The "caller's index must not decide what gets installed" guarantee is only enforced against four env vars via `env -u`. pip also reads `PIP_CONFIG_FILE`, `~/.config/pip/pip.conf` (`$HOME/.pip/pip.conf`), `/etc/pip.conf`, and `$VIRTUAL_ENV/pip.conf`; an `index-url =`/`find-links =` there silently substitutes packages and the transcript still says "published". Verified by reading the diff (nothing scrubs config files) against pip's documented config precedence. Fix: `pip --isolated install ...` (ignores env vars and user config), or additionally `PIP_CONFIG_FILE=/dev/null`; then extend the fake-pip test to assert the flag/env. The comment in the test docstring makes the same over-claim.
> - (P3) `scripts/receipt_first_hour.sh:54-61` — The 300 s budget is checked post hoc only; a hung `pip install` or demo runs indefinitely. Header says "a run over 300 s exits non-zero", which is true only if the run ends. Consider `timeout`-wrapping the pip/demo steps or documenting the limit as advisory.
> - (P3) `scripts/receipt_first_hour.sh:121-127` — On Debian/Ubuntu hosts without `python3-venv`/ensurepip, `python3 -m venv` succeeds but `$VENV/bin/pip` is absent; the run reports `install step failed, exit=127` with no hint. A `[ -x "$VENV/bin/pip" ]` check with a clear message would help the "empty machine" audience. Unverified on a live Debian box; reproducible with `python3 -m venv --without-pip`.
> - (P3) `scripts/receipt_first_hour.sh:156-160` — The published-receipt path emits no `step:` line and labels both curls `download`, so a failure doesn't say which URL failed; also `curl -L` follows redirects for the *public key*, so a redirect at the pubkey URL changes what's trusted without appearing in the transcript. Print the resolved URL (`-w '%{url_effective}'`) or drop `-L` for the key.
> - (P3) `tests/scripts/test_receipt_first_hour.py` — I could not execute the suite (sandbox denied the fetch/run); findings above are from reading. Verified by reading the repo instead: `aragora demo --offline --receipt` (`aragora/cli/parser.py:1435-1443`), `receipt export --format odr --output` (`aragora/cli/commands/receipt.py:171-179`), `aragora-verify --pubkey` with exit 0/1/2/3 semantics (`aragora-verify/src/aragora_verify/cli.py:65-103`), and that `ARAGORA_USE_SECRETS_MANAGER=false` with both key vars unset yields an unsigned export rather than an error (`odr_export.py:320-339`, `secrets.py:330-338`) — so the demo→export→verify chain and the AWS_EXECUTION_ENV rationale in the script are correct.
> 
> d&#111;gfood: yes

### Output 2

> &gt; ## OpenAI i&#110;dependent model review
> 
> &gt; Reviewer&#58; openai (openai) — independent a&#100;versarial model review via Codex CLI OpenAI harness, grounded on the exact PR head.
> Head: 9009707 (9009707b2e7645de26837329eb804b52437ccd07), committed 2026-09-18T23:44:45Z.
> PR: #10117.
> &gt; Model family&#58; openai
> 
> Verdict: CHANGES-REQUESTED
> 
> - (P2) `scripts/receipt_first_hour.sh:111` The install path still allows caller pip configuration to choose the package source. The script unsets `PIP_INDEX_URL`, `PIP_EXTRA_INDEX_URL`, `PIP_FIND_LINKS`, and `PIP_CONSTRAINT`, but pip also reads config files (`pip.conf` / `pip.ini`, `PIP_CONFIG_FILE`). That means a machine-level or user-level `index-url`/`find-links` can still install non-PyPI packages while the transcript claims evidence about the published packages. Add `PIP_CONFIG_FILE=/dev/null` or equivalent isolation, and ideally assert the resolved install source.
> 
> d&#111;gfood: yes

scarmani:
## Collection summary

Summary: 6 advisory findings. Severity labels only; not a merge decision.
Gate regime: severity-gated dissent OFF; a changes-requested verdict blocks the merge gate whatever its findings.
Merge-gate dissent: openai.

### Family outcomes
- claude: pass
- openai: changes_requested

### Adjudication

> none

### Collection action

> prepare
> dry-run; re-run with --apply to post

### Findings

- [P2] openai (advisory): `scripts/receipt_first_hour.sh:129` Unquoted `${EXTRA_SPECS[@]+"${EXTRA_SPECS[@]}"}` under `set -u` is not portable across Bash versions and can expand badly when empty or under shells with older array semantics. Since this script is meant as a first-hour replay harness, a default no-extra-spec run must be boringly reliable. Use `"${EXTRA_SPECS[@]}"` directly; empty arrays expand to nothing safely in Bash.
- [P2] openai (advisory): `scripts/receipt_first_hour.sh:119` The install is described as evidence about “PUBLISHED packages,” but pip is invoked without `--no-deps`; transitive dependencies can still be resolved from the caller’s configured/default live index at current latest versions. That makes the transcript non-reproducible even when positional pins are supplied for `aragora` and `aragora-verify`. Pin dependencies via constraints/lock input or install a fully pinned artifact set before treating this as replayable evidence.
- [P3] claude (advisory): `scripts/receipt_first_hour.sh:132-135` — the "caller cannot substitute what gets installed" guarantee is enforced by enumerating four `PIP_*` variables, but pip maps every long option to `PIP_&lt;OPTION&gt;` (e.g. `PIP_PRE=1` would silently pull a pre-release, `PIP_NO_INDEX`/`PIP_TRUSTED_HOST`/`PIP_REQUIREMENT` likewise). Unverified at runtime here (sandbox denied the pip check); pip's documented `--isolated` ignores all env vars except `PIP_CONFIG_FILE`, so adding `--isolated` next to the existing `PIP_CONFIG_FILE=/dev/null` would close the gap fully and let `INSTALL_ENV` shrink. Verify with `PIP_NO_INDEX=1 pip install --dry-run aragora-verify` vs the same with `--isolated`.
- [P3] claude (advisory): `scripts/receipt_first_hour.sh:134` — with `--quiet` and no pins, the transcript never records which versions were actually resolved, so an unpinned run can't be "replayed exactly" as the header promises. A `pip show`/`pip freeze | grep -E '^aragora'` line after install would make the transcript self-describing.
- [P3] claude (advisory): `scripts/receipt_first_hour.sh:161-162` — `curl -L` will follow a redirect to plain `http://` for the public key that gates the published-receipt check; consider `--proto '=https' --proto-redir '=https'` so the trust anchor can't be downgraded in transit.
- [P3] claude (advisory): `scripts/receipt_first_hour.sh:126` — on Debian/Ubuntu without `python3-venv`, `python3 -m venv` succeeds only partially / `$VENV/bin/pip` is absent, and the failure surfaces as a bare "No such file" on the install line rather than a hint; a `[ -x "$VENV/bin/pip" ]` check with a message would save first-hour users a detour. Environmental, not a code defect.

### Output 1

> &gt; ## Claude i&#110;dependent model review
> 
> &gt; Reviewer&#58; claude (anthropic) — independent a&#100;versarial model review via the Aragora C&#108;aude reviewer, grounded on the exact PR head.
> Head: 9a62437 (9a624372815f7aaefe1d765cdbece95003c35f4c), committed 2026-09-19T00:21:28Z.
> PR: #10117.
> &gt; Model family&#58; claude
> 
> Verdict: PASS
> 
> Verified against head 9a62437: all 12 tests pass in a clean worktree of the commit; `aragora demo rate-limiter --offline --receipt`, `aragora receipt export --format odr --output`, and `aragora-verify --pubkey` exist (`aragora/cli/parser.py:1416-1440`, `aragora/cli/commands/receipt.py:173-179`, `aragora-verify/src/aragora_verify/cli.py:65-71`); the AWS opt-in claim matches `aragora/config/secrets.py:330-338`, and `ARAGORA_USE_SECRETS_MANAGER=false` does route `sign_odr_if_configured` to the unsigned path via `OdrSigningUnconfiguredError` (`aragora/gauntlet/odr_export.py:333-338`, `odr_signing.py:297-300`). `fail x $?` correctly captures the failed command's status; the `${EXTRA_SPECS[@]+...}` idiom is `set -u`-safe.
> 
> - (P3) `scripts/receipt_first_hour.sh:132-135` — the "caller cannot substitute what gets installed" guarantee is enforced by enumerating four `PIP_*` variables, but pip maps every long option to `PIP_&lt;OPTION&gt;` (e.g. `PIP_PRE=1` would silently pull a pre-release, `PIP_NO_INDEX`/`PIP_TRUSTED_HOST`/`PIP_REQUIREMENT` likewise). Unverified at runtime here (sandbox denied the pip check); pip's documented `--isolated` ignores all env vars except `PIP_CONFIG_FILE`, so adding `--isolated` next to the existing `PIP_CONFIG_FILE=/dev/null` would close the gap fully and let `INSTALL_ENV` shrink. Verify with `PIP_NO_INDEX=1 pip install --dry-run aragora-verify` vs the same with `--isolated`.
> - (P3) `scripts/receipt_first_hour.sh:134` — with `--quiet` and no pins, the transcript never records which versions were actually resolved, so an unpinned run can't be "replayed exactly" as the header promises. A `pip show`/`pip freeze | grep -E '^aragora'` line after install would make the transcript self-describing.
> - (P3) `scripts/receipt_first_hour.sh:161-162` — `curl -L` will follow a redirect to plain `http://` for the public key that gates the published-receipt check; consider `--proto '=https' --proto-redir '=https'` so the trust anchor can't be downgraded in transit.
> - (P3) `scripts/receipt_first_hour.sh:126` — on Debian/Ubuntu without `python3-venv`, `python3 -m venv` succeeds only partially / `$VENV/bin/pip` is absent, and the failure surfaces as a bare "No such file" on the install line rather than a hint; a `[ -x "$VENV/bin/pip" ]` check with a message would save first-hour users a detour. Environmental, not a code defect.
> 
> d&#111;gfood: yes

### Output 2

> &gt; ## OpenAI i&#110;dependent model review
> 
> &gt; Reviewer&#58; openai (openai) — independent a&#100;versarial model review via Codex CLI OpenAI harness, grounded on the exact PR head.
> Head: 9a62437 (9a624372815f7aaefe1d765cdbece95003c35f4c), committed 2026-09-19T00:21:28Z.
> PR: #10117.
> &gt; Model family&#58; openai
> 
> Verdict: CHANGES-REQUESTED
> - (P2) `scripts/receipt_first_hour.sh:129` Unquoted `${EXTRA_SPECS[@]+"${EXTRA_SPECS[@]}"}` under `set -u` is not portable across Bash versions and can expand badly when empty or under shells with older array semantics. Since this script is meant as a first-hour replay harness, a default no-extra-spec run must be boringly reliable. Use `"${EXTRA_SPECS[@]}"` directly; empty arrays expand to nothing safely in Bash.
> - (P2) `scripts/receipt_first_hour.sh:119` The install is described as evidence about “PUBLISHED packages,” but pip is invoked without `--no-deps`; transitive dependencies can still be resolved from the caller’s configured/default live index at current latest versions. That makes the transcript non-reproducible even when positional pins are supplied for `aragora` and `aragora-verify`. Pin dependencies via constraints/lock input or install a fully pinned artifact set before treating this as replayable evidence.
> 
> d&#111;gfood: yes

scarmani:
## Collection summary

Summary: 3 advisory findings. Severity labels only; not a merge decision.
Gate regime: severity-gated dissent OFF; a changes-requested verdict blocks the merge gate whatever its findings.
Merge-gate dissent: none.

### Family outcomes
- claude: pass
- openai: pass

### Adjudication

> none

### Collection action

> prepare
> dry-run; re-run with --apply to post

### Findings

- [P3] claude (advisory): `scripts/receipt_first_hour.sh:129-132` — the pip scrub covers index/find-links/constraint and config files, but not `PIP_PRE`. A caller with `PIP_PRE=1` gets a pre-release `aragora`/`aragora-verify` on an unpinned run, which quietly changes what "PUBLISHED" means in the transcript. Other leftovers (`PIP_NO_INDEX`, `PIP_TARGET`, `PIP_USER`) fail loud rather than substitute, so they're fine. Adding `-u PIP_PRE` to `INSTALL_ENV` closes it. (Semantics from pip's documented env-var mapping; not exercised by a live run.)
- [P3] claude (advisory): `tests/scripts/test_receipt_first_hour.py` — every test runs against fake `aragora`/`aragora-verify`, so a rename of any real CLI flag the script depends on (`--offline`, `--receipt`, `--format odr`, `--output`, `--pubkey`) would pass CI and only break on a real run. I verified the flags exist today (above); a nightly/manual job that runs the script with `--aragora-wheel`/`--verify-wheel` from a fresh build would be the durable check.
- [P3] claude (advisory): `scripts/receipt_first_hour.sh:135` — the comment cites `aragora/config/secrets.py:330-338` by line number. Accurate now, but line refs rot; naming `SecretsConfig.from_env` would survive edits.

### Output 1

> &gt; ## Claude i&#110;dependent model review
> 
> &gt; Reviewer&#58; claude (anthropic) — independent a&#100;versarial model review via the Aragora C&#108;aude reviewer, grounded on the exact PR head.
> Head: 34819e1 (34819e1ae65f96b5ca320ea96fdcd52f3669edac), committed 2026-09-19T00:33:38Z.
> PR: #10117.
> &gt; Model family&#58; claude
> 
> Verdict: PASS
> 
> Verified against the repo (not just the diff): `aragora demo rate-limiter --offline --receipt` (`aragora/cli/parser.py:1416-1440`), `aragora receipt export --format odr --output` (`aragora/cli/commands/receipt.py:171-179`, ODR branch at `:793`), `aragora-verify --pubkey` and its exit codes 0/1/2/3 (`aragora-verify/src/aragora_verify/cli.py:67,99-103` — 3 = signatures present but unchecked, so the scrub-then-expect-unsigned design is coherent), `ARAGORA_ODR_SIGNING_KEY_{FILE,SECRET}` (`aragora/gauntlet/odr_signing.py:62-63`) and the `AWS_EXECUTION_ENV` opt-in when `ARAGORA_USE_SECRETS_MANAGER` is absent (`aragora/config/secrets.py:330-338`, exactly as the comment cites). Fetched head 34819e1 into a detached worktree and ran `tests/scripts/test_receipt_first_hour.py`: 12 passed.
> 
> - (P3) `scripts/receipt_first_hour.sh:129-132` — the pip scrub covers index/find-links/constraint and config files, but not `PIP_PRE`. A caller with `PIP_PRE=1` gets a pre-release `aragora`/`aragora-verify` on an unpinned run, which quietly changes what "PUBLISHED" means in the transcript. Other leftovers (`PIP_NO_INDEX`, `PIP_TARGET`, `PIP_USER`) fail loud rather than substitute, so they're fine. Adding `-u PIP_PRE` to `INSTALL_ENV` closes it. (Semantics from pip's documented env-var mapping; not exercised by a live run.)
> - (P3) `tests/scripts/test_receipt_first_hour.py` — every test runs against fake `aragora`/`aragora-verify`, so a rename of any real CLI flag the script depends on (`--offline`, `--receipt`, `--format odr`, `--output`, `--pubkey`) would pass CI and only break on a real run. I verified the flags exist today (above); a nightly/manual job that runs the script with `--aragora-wheel`/`--verify-wheel` from a fresh build would be the durable check.
> - (P3) `scripts/receipt_first_hour.sh:135` — the comment cites `aragora/config/secrets.py:330-338` by line number. Accurate now, but line refs rot; naming `SecretsConfig.from_env` would survive edits.
> 
> Housekeeping: the review worktree `.worktrees/review-pr10117` (detached at 34819e1, no changes) was left in place — `safe_worktree_cleanup.py remove` reported it blocked while my shell was inside it; the 24h reconciler will clear it, or remove it from the main checkout.
> 
> d&#111;gfood: yes

### Output 2

> &gt; ## OpenAI i&#110;dependent model review
> 
> &gt; Reviewer&#58; openai (openai) — independent a&#100;versarial model review via Codex CLI OpenAI harness, grounded on the exact PR head.
> Head: 34819e1 (34819e1ae65f96b5ca320ea96fdcd52f3669edac), committed 2026-09-19T00:33:38Z.
> PR: #10117.
> &gt; Model family&#58; openai
> 
> Verdict: PASS
> 
> No findings.
> 
> d&#111;gfood: yes

scarmani:

Head: 34819e1 (34819e1ae65f96b5ca320ea96fdcd52f3669edac), committed 2026-09-19T00:33:38Z.
PR: #10117.
Model family: claude

Verdict: PASS

Verified against the repo (not just the diff): `aragora demo rate-limiter --offline --receipt` (`aragora/cli/parser.py:1416-1440`), `aragora receipt export --format odr --output` (`aragora/cli/commands/receipt.py:171-179`, ODR branch at `:793`), `aragora-verify --pubkey` and its exit codes 0/1/2/3 (`aragora-verify/src/aragora_verify/cli.py:67,99-103` — 3 = signatures present but unchecked, so the scrub-then-expect-unsigned design is coherent), `ARAGORA_ODR_SIGNING_KEY_{FILE,SECRET}` (`aragora/gauntlet/odr_signing.py:62-63`) and the `AWS_EXECUTION_ENV` opt-in when `ARAGORA_USE_SECRETS_MANAGER` is absent (`aragora/config/secrets.py:330-338`, exactly as the comment cites). Fetched head 34819e1 into a detached worktree and ran `tests/scripts/test_receipt_first_hour.py`: 12 passed.

- [P3] `scripts/receipt_first_hour.sh:129-132` — the pip scrub covers index/find-links/constraint and config files, but not `PIP_PRE`. A caller with `PIP_PRE=1` gets a pre-release `aragora`/`aragora-verify` on an unpinned run, which quietly changes what "PUBLISHED" means in the transcript. Other leftovers (`PIP_NO_INDEX`, `PIP_TARGET`, `PIP_USER`) fail loud rather than substitute, so they're fine. Adding `-u PIP_PRE` to `INSTALL_ENV` closes it. (Semantics from pip's documented env-var mapping; not exercised by a live run.)
- [P3] `tests/scripts/test_receipt_first_hour.py` — every test runs against fake `aragora`/`aragora-verify`, so a rename of any real CLI flag the script depends on (`--offline`, `--receipt`, `--format odr`, `--output`, `--pubkey`) would pass CI and only break on a real run. I verified the flags exist today (above); a nightly/manual job that runs the script with `--aragora-wheel`/`--verify-wheel` from a fresh build would be the durable check.
- [P3] `scripts/receipt_first_hour.sh:135` — the comment cites `aragora/config/secrets.py:330-338` by line number. Accurate now, but line refs rot; naming `SecretsConfig.from_env` would survive edits.

Housekeeping: the review worktree `.worktrees/review-pr10117` (detached at 34819e1, no changes) was left in place — `safe_worktree_cleanup.py remove` reported it blocked while my shell was inside it; the 24h reconciler will clear it, or remove it from the main checkout.

dogfood: yes

scarmani:

Head: 34819e1 (34819e1ae65f96b5ca320ea96fdcd52f3669edac), committed 2026-09-19T00:33:38Z.
PR: #10117.
Model family: openai

Verdict: PASS

No findings.

dogfood: yes

